PR #22, opened 2020-05-22
remove unwrap in parser

chenyan-dfinity:


chenyan-dfinity:
Yes, it seems to appear in a recent version of lalrpop :)

PR #31, opened 2020-05-25
Feature: adding a Candid-Rust binding crate

hansl:
Allows code generation for Rust, and implement a simple framework to generate other languages.


candid_rust/tests/assets/examples/linkedup.rs
```diff
@@ -0,0 +1,34 @@
+type UserId_2 = UserId;
+type UserId = principal;
```
chenyan-dfinity, line 2: Do we type check the generated code? We don't support principal yet. The parser is likely to treat it as a variable. If we type check the code, it will probably fail.

candid_rust/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+            PrimType::Int8 => quote! { i8 },
+            PrimType::Int16 => quote! { i16 },
+            PrimType::Int32 => quote! { i32 },
+            PrimType::Int64 => quote! { i64 },
+            PrimType::Float32 => quote! { f32 },
+            PrimType::Float64 => quote! { f64 },
+            PrimType::Bool => quote! { bool },
+            PrimType::Text => quote! { String },
+            PrimType::Null => quote! { () },
+            PrimType::Empty => quote! { () },
```
chenyan-dfinity, line 157: `Empty` needs to be an inhabited type, either `enum {}` or `!`.

candid_rust/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+                let label = syn::Ident::new(&label_str, span.span());
+                let ty = idl_type_usage_to_rust(context, &f.typ, &span.span(), true)?;
+
+                all_fields.push(quote! {
+                    pub #label : #ty
+                });
+            }
+
+            quote! {
+                struct #id {
```
chenyan-dfinity, line 121: should this be `pub`?

candid_rust/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+            PrimType::Nat64 => quote! { u64 },
+            PrimType::Int => quote! { i128 },
+            PrimType::Int8 => quote! { i8 },
+            PrimType::Int16 => quote! { i16 },
+            PrimType::Int32 => quote! { i32 },
+            PrimType::Int64 => quote! { i64 },
+            PrimType::Float32 => quote! { f32 },
+            PrimType::Float64 => quote! { f64 },
+            PrimType::Bool => quote! { bool },
+            PrimType::Text => quote! { String },
```
chenyan-dfinity, line 155: Do we have a freedom to choose between `String` and `&str`? #justsaying

candid_rust/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    // but idl_type_usage_to_rust() errors on some types because they're invalid in a
+    // type usage in rust.
+    return Ok(match typ {
+        IDLType::RecordT(fields) => {
+            let mut all_fields = Vec::new();
+            for f in fields {
+                let label_str = match &f.label {
+                    Label::Id(i) => format!("id_{}", i),
+                    Label::Named(n) => n.clone(),
+                    Label::Unnamed(i) => format!("unnamed_{}", i),
```
chenyan-dfinity, line 110: `Id(i)` and `Unnamed(i)` have no external difference, so the generated label should be the same, e.g. `id_{i}`.

candid_rust/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+
+pub fn idl_prog_into_stream(prog: IDLProg) -> Result<TokenStream> {
+    let mut context = Context::new(&prog);
+
+    let mut all_declarations = Vec::<TokenStream>::new();
+    for dec in &prog.decs {
+        context.add_declaration(dec);
+    }
+
+    for dec in &prog.decs {
```
chenyan-dfinity, line 197: Starting directly from `dec` may include unused type defs in the generated code. Now that we have context, a more robust way is to start from actor, and gather the mentioned type definitions into context.

candid_rust/tests/assets/examples/linkedup.rs
hansl: We don't _type check_ the generated code (as it can be part of another module, but we do syntax checking in tests to make sure it's proper rust.

candid_rust/src/lib.rs
hansl: Done.

candid_rust/src/lib.rs
hansl: Rewrote this whole code, added `pub` everywhere.

candid_rust/src/lib.rs
hansl: Not if it's in a struct :/

candid_rust/src/lib.rs
hansl: Done.

candid_rust/src/lib.rs
hansl: I think that's fine for now. This would be a larger change right now than required. I can do it as an optimization later.

hansl:
@chenyan-dfinity PTAL.

candid/src/codegen/rust.rs
```diff
@@ -0,0 +1,225 @@
...
+//! natural/integer numbers.
+use crate::codegen::LanguageBinding;
+use crate::error::{Error, Result};
+use crate::parser::types::{Binding, Dec, FuncType, IDLType, Label, PrimType, TypeField};
+use crate::{generate_code, IDLProg};
+
+#[derive(Clone, Default)]
+pub struct Config {
+    pub actor_name: Option<String>,
+    pub use_bigint: Option<()>,
```
chenyan-dfinity, line 16: While it's nice to have this option, the decoder currently only decodes to bigint. And why is the type Option<()>, not bool?

candid/src/codegen/rust.rs
```diff
@@ -0,0 +1,225 @@
...
+    config: &'a Config,
+    prog: &'a IDLProg,
+}
+
+impl<'a> LanguageBinding for RustLanguageBinding<'a> {
+    fn usage_prim(&self, ty: &PrimType) -> Result<String> {
+        Ok(match ty {
+            PrimType::Nat => {
+                if self.config.use_bigint.is_some() {
+                    "num_bigint::BigUint"
```
chenyan-dfinity, line 29:
```suggestion
                    "candid::Nat"
```

candid/src/codegen/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// The main function which is called when starting the code generation.
+    fn prog(&self, prog: &IDLProg) -> Result<String> {
+        self.start()?;
+
+        let output = self.header()?
+            + &self.declarations(&prog.decs)?
+            + &match &prog.actor {
+                None => String::new(),
+                Some(IDLType::ServT(bindings)) => self.service(bindings)?,
+                _ => unimplemented!(),
```
chenyan-dfinity, line 165:
```suggestion
                _ => unreachable!(),
```

candid/src/codegen/rust.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        Ok(format!(
+            "pub trait {} {{ {} }}",
+            actor_str,
+            all_functions.join(" ")
+        ))
+    }
+}
+
+/// Takes an IDL string and returns a Rust string, formatted.
```
chenyan-dfinity, line 217: unformated?

candid/src/codegen/rust.rs
```diff
@@ -0,0 +1,225 @@
...
+                let arg_name = format!("arg{}", i);
+                let ty = self.usage(&ty)?;
+
+                Ok(format!("{}: {}", arg_name, ty))
+            })
+            .collect::<Result<Vec<String>>>()?
+            .join(" , ");
+
+        Ok(format!(
+            "fn {id}( {arguments} ) {return_type} ;",
```
chenyan-dfinity, line 184: Be aware that this `id` can be unicode 🐂

candid/tests/assets/codegen/basic/recursive.rs
```diff
@@ -0,0 +1,8 @@
+pub struct List {
```
chenyan-dfinity, line 1: Do you want to add a derive here?

candid/src/codegen/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /// The footer string that will be appended to the output.
+    fn footer(&self) -> Result<String> {
+        Ok(String::new())
+    }
+
+    /// This method is called when a type is used as a right hand side. By default
+    /// it forwards it to a specialized method for each type.
+    fn usage(&self, ty: &IDLType) -> Result<String> {
```
chenyan-dfinity, line 40: The `.did` file is all about type declarations. There is no notion of `usage` or `right hand side`. The codegen should have the freedom to define a new type when you feel like to. And the only way you can "use" the definition is by referring to its name. The other "use" cases, e.g. vec, opt, etc, are really inlining the definition as an optimization.

candid/src/codegen/rust.rs
hansl: That's just illegal... :/

Also, it can be `fn` for example which is a keyword. I need to fix that one too.

candid/tests/assets/codegen/basic/recursive.rs
hansl: I added that as a config option.

candid_rust/src/lib.rs
chenyan-dfinity: It's possible with a lifetime in a struct...

candid/src/codegen/rust.rs
chenyan-dfinity: When importing did from Motoko, we convert the id to `_idl_hash(id)_` when id is not a valid Rust identifier.

candid/src/codegen/mod.rs
hansl: See offline discussion.

candid_rust/src/lib.rs
hansl: Yes but that complicates the generation and I'd like to keep it simple.

candid/src/codegen/rust.rs
hansl: Done.

PR #37, opened 2020-06-02
type checking for did file; typed serialization for IDLValue

chenyan-dfinity:
- type checking
- fix for bigint LEB128
- serialize IDLValue with candid types
- remove `none` token from IDLValue parser, which was never in the spec anyway.


candid/src/parser/grammar.lalrpop
```diff
@@ -14,7 +14,6 @@ extern {
         "text" => Token::Text(<String>),
         "id" => Token::Id(<String>),
         "null" => Token::Null,
-        "none" => Token::None,
```
hansl, line 17: This is not in your PR description. Which part exactly is removed?

chenyan-dfinity: Updated the description. `none` was never in the spec. I added it here to support `null` in the option types. Now that we have type annotations from candid, we should be able to use `null` as the spec intended.

candid/src/ser.rs
```diff
@@ -174,68 +190,78 @@ impl TypeSerialize {
...
-            }
+        if self.type_map.contains_key(t) {
+            return Ok(());
+        }
+        let actual_type = if let Type::Var(id) = t {
+            self.env.rec_find_type(id)?
+        } else {
+            t
+        }
+        .clone();
```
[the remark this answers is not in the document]
chenyan-dfinity: `cargo fmt` insists this.

PR #38, opened 2020-06-04
Update parse_type.rs test

seniorjoinu:
Just a typo fix.
It is impossible to construct such type.


candid/tests/parse_type.rs
```diff
@@ -12,7 +12,7 @@ fn parse_idl_prog() {
     let prog = r#"
 import "test.did";
 type my_type = nat8;
-type List = record { head: int; tail: List };
+type List = record { head: int; tail: opt List };
```
chenyan-dfinity, line 15:
```suggestion
type List = variant { head: int; tail: List };
```

rossberg: This change doesn't make sense to me. That variant doesn't define a list, whereas the record did define a (non-empty) list.

rossberg:
@chenyan-dfinity:
> The spec actually allows type A = record { a: A }. This is equivalent to the Empty type, which has no value.

Nit: they are isomorphic (in Candid), but not equivalent. ;)

seniorjoinu:
I was thinking of your suggested change (opt -> variant).
While this type can be constructed, it is not a "list" anymore - one only can define "head" or "tail" but not both. 
I think we should revert to `opt`.

candid/tests/parse_type.rs
```diff
@@ -12,7 +12,7 @@ fn parse_idl_prog() {
     let prog = r#"
 import "test.did";
 type my_type = nat8;
-type List = record { head: int; tail: List };
+type List = variant { head: int; tail: List };
```
nomeata, line 15: I guess this should be
```suggestion
type List = opt record { head: int; tail: List };
```

seniorjoinu: It is very interesting, how many forms can one simple data structure have.
I vote for this.

candid/tests/parse_type.rs
chenyan-dfinity: Right, I wasn't thinking. As @nomeata suggested, it should be `opt record` instead.

PR #40, opened 2020-06-09
Feature: add all conversions and operators to Nat/Int

hansl:
This includes a lot of native types operators, like `+= 1`.

It also includes conversions from all unsigned types to Nat,
all signed types to Int and i32 to Nat.

Added tests that will always pass but at least check that those
operators are still working as expected. It should really only
break at compile time.

### DIFF
--- a/candid/src/lib.rs
+++ b/candid/src/lib.rs
@@ -90,7 +90,7 @@
 //!   let x = "-10000000000000000000".parse::<Int>()?;
 //!   let bytes = Encode!(&Nat::from(1024), &x)?;
 //!   let (a, b) = Decode!(&bytes, Nat, Int)?;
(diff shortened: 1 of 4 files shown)


candid/src/lib.rs
```diff
@@ -90,7 +90,7 @@
 //!   let x = "-10000000000000000000".parse::<Int>()?;
 //!   let bytes = Encode!(&Nat::from(1024), &x)?;
 //!   let (a, b) = Decode!(&bytes, Nat, Int)?;
-//!   assert_eq!(a, 1024.into());
+//!   assert_eq!(a, 1024);
```
chenyan-dfinity, line 93:
```suggestion
//!   assert_eq!(a + 1, 1025);
```

PR #52, opened 2020-07-02
JS binding for Candid

chenyan-dfinity:
- JS binding for Candid, which produces the same code as the OCaml version.
- Refactor the type checker. The type checker returns `Option<Type>` instead of `ActorEnv`. This is to ensure we preserve as much information as possible from the AST.
- Type checker is now complete.
- Fix `idl_hash` for unicode


candid/src/parser/typing.rs
```diff
@@ -124,6 +123,15 @@ pub fn check_type(env: &Env, t: &IDLType) -> Result<Type> {
             for t in func.rets.iter() {
                 t2.push(check_type(env, t)?);
             }
+            if func.modes.len() > 1 {
+                return Err(Error::msg("cannot have more thaan one mode"));
```
nomeata:
```suggestion
                return Err(Error::msg("cannot have more than one mode"));
```
NB: That’s not actually part ofthe Candid spec; it allows `oneway query` functions. Not a very useful thing, of course, so maybe fine for now, until we have more annotations that actually can be composed. (Also, this is a port of the Ocaml code, right? So not an issue for this PR per se.)

chenyan-dfinity: Right, the spec doesn't explicitly exclude `oneway query` function, but in the async world, they have conflict requirements. When we have more annotations, we will definitely revisit this.

Issue #57, opened 2020-07-13
As a user/dev, I want to compare/diff/relate Candid values


matthewhammer:
**Is your feature request related to a problem? Please describe.**

This issue proposes a new crate, related to the one here.  Let's call it `candid_relate`.  It gives semantic operations over ASTs of Candid values and types, including:

 - Structural comparison (in the sense of `mo:base/Order.Order`)
 - Structural editing, with a representation of diffs/deltas.
 - Equality checks, in a semantic sense that requires types (re-implement some semantics of Candid, but at a higher level of abstraction, with more debugging information when things "go wrong").

This crate is not essential for building candid services, but would be essential for any candid tool that tries to debug expected versus observed values of candid data.

Here's an example of some related uses, demonstrating `diff`, `delta` and `equal`, and how they inter-relate:

```rust

// suppose we have two Candid ASTs, and a type that classifies each:
let (x : Candid::Ast, y : Candid::Ast, t : Candid::Type) = ... ;

// (note: candid values x and y each have a common candid super-type, t)
// At the common type, we can edit one value into the other
let d : Candid::Delta = candid_relate::diff(x, y, t);

// difference/delta is represented by d, and can be applied as follows:
let z = candid_relate::delta(x, d);

// equal at the common supertype t, but perhaps not generally:
assert_eq!( Candid::equal(z, y, t) );
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

matthewhammer:
@chenyan-dfinity Should we have multiple crates in this repo, like other Rust language projects?  Or separate repos for each crate?  (The downside of separate is that we loose visibility outside of `dfinity` org, and the cost of making more within the `dfinity` org is high, in terms of process involved --- seems easiest/best now to do a multi-crate repo?)

matthewhammer:
> Should we have multiple crates in this repo

Nervermind that question.  I see that we do now.

New question: May I do a PR to add `candid_relate` here as well?  WDYT?

chenyan-dfinity:
Sure, feel free to make a PR here.

chenyan-dfinity:
> pretty-print a report that summarizes ("usefully") how the two values differ

FYI I recently tried the `pretty` crate and find it very nice. See this PR: https://github.com/dfinity/candid/pull/56

matthewhammer:
MVP in #58 with some tracking issues for the missing pieces:
 - [ ] #68 
 - [ ] #67

PR #58, opened 2020-07-17
Candiff tool to systematically compare Candid values

matthewhammer:
Initial sketch for addressing #57


matthewhammer:
FYI @chenyan-dfinity --- I put some initial thoughts here, but there's quite a bit left to do before it's ready for a review.

chenyan-dfinity:
Just bikeshedding, `candid_relate` feels a bit PLish, do we want a less academic name like `candid_diff`, `candiff`?

candid_relate/src/lib.rs
```diff
@@ -0,0 +1,103 @@
+extern crate candid;
+
+use std::rc::Rc;
+
+pub type Type = candid::parser::types::IDLType;
+pub type Label = candid::parser::types::Label;
```
chenyan-dfinity, line 6: You may be using an outdated code. `Label` has been moved to `candid::types::Label`. In general, we prefer to use `candid::types::Type` when referring to candid types. `candid::parser::types::*` is only an AST, and come be converted to `types::Type` via `candid::parser::typing::check_prog`. The benefit of this design is that we get a consistent view for both types derived from Rust and from candid file.

candid_relate/src/lib.rs
```diff
@@ -0,0 +1,103 @@
+extern crate candid;
```
chenyan-dfinity, line 1: With Rust 2018 edition, we don't need to declare `extern crate` anymore.

matthewhammer:
I take your point about the name.  In renaming things, I chose `candid_diff` (over `candiddiff`):
 - avoid double `d`
 - follow convention set by `candid_dervive`

matthewhammer:
Oh, and `candiff` is really cool and clever for a name, but that seems like it's more about the canister content (code or state) and not the messages going into / out of that canister.

matthewhammer:
Hmm... maybe I'm coming around to `candiff` as the name of the binary tool for the `candid_diff` crate (WDYT?)

Also, we should design the CLI with some examples somewhere.  Will add a design doc for that.

candid_diff/design/candiff-CLI-tool.md
```diff
@@ -0,0 +1,89 @@
...
+common type, but we do not know if this the case or not, a priori.
+When provided, this common type helps inform the output information,
+but it is not required.  When absent, the tool gives less helpful
+information on values without a common type.
+
+### Typing
+
+Two cases:
+
+- Type invariant: Both values have a common type.
```
chenyan-dfinity, line 42: Do we want to consider subtyping, e.g. two values having a common subtype?

candid_diff/design/candiff-CLI-tool.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Variant changes
+
+- Same constructor, change payload
+- (New constructor is a higher-level change)
+
+#### Record changes
+
+- Change particular field(s) value
+- Remove particular field(s)
```
chenyan-dfinity, line 87: Because of subtyping, some of these edits can have zero cost on the value.

candid_diff/design/candiff-CLI-tool.md
```diff
@@ -0,0 +1,89 @@
+# Tool design (`candiff`)
+
+The `candid_diff` package exposes the `candiff` tool for scripting Candid value comparisons that have human-readable error messages.
+
+The tool compares input and output canister messages to those recorded earlier, from earlier run(s).
+
+## Use cases
```
chenyan-dfinity, line 7: Another use case is https://github.com/dfinity-lab/motoko/issues/1748, so that users can integrate a common test set into their own testing workflow.

candid_diff/design/candiff-CLI-tool.md
matthewhammer: Maybe?  Or "upgrades", if those are distinct notions (?).

The simplest use case doesn't require that, but we should note that other (future) uses could implement more of the type theory of candid, including that relation, and follow it.

candid_diff/design/candiff-CLI-tool.md
matthewhammer: Good point.  Will update to reflect this.

chenyan-dfinity:
Good question. The `candid` crate is also overloaded: it includes both 1) Rust-specific serialization and 2) a compiler for candid, equivalent to `didc`. Maybe we should split the crate and put them into separate directories as well: one under rust/; the other under tool/, which can also include the `candiff` tool.

matthewhammer:
I don't mind where `candid_diff` lives, but `candiff` the binary tool is only one aspect of a larger Rust crate, right?

My thinking: As is usual in Rust, the crate `candid_diff` would provide a library (same name) and a binary tool (`candiff`), and users can build on the library with more Rust, or use the binary tool as a CLI user, or both.

matthewhammer:
@chenyan-dfinity Question about the parser logic for the Candid textual format: How do we ever get non-`Number` cases?  

Currently, all of my tests of parsed (textual) numbers stay as strings, and perhaps that's intentional?

If so, do I need to have a type to resolve them to a known number type (if they start as textual format)?

candid_diff/src/candiff.rs
```diff
@@ -0,0 +1,298 @@
...
+            .stdout(predicate::eq(b"1\n" as &[u8]))
+            .success();
+    }
+
+    #[test]
+    fn echo_number_debug() {
+        let mut cmd = Command::cargo_bin("candiff").unwrap();
+        cmd.arg("echo").arg("1").arg("-d");
+        cmd.assert()
+            .stdout(predicate::eq(b"Number(\"1\")\n" as &[u8]))
```
matthewhammer, line 196: @chenyan-dfinity Here's an example of how the parser works now: I always get back a `Number`, and never a `Nat` or `Int`, etc.  Unfortunately, I don't think this is the behavior we want for the diff tool.  Thoughts?

chenyan-dfinity: Yes, the current assumption is that the values will always have an associated candid file, and you need to call https://github.com/dfinity/candid/blob/master/candid/src/parser/value.rs#L176 to get the non-Number type. Specifically, the types listed here https://github.com/dfinity/candid/blob/master/candid/src/parser/value.rs#L23 won't be derived without a candid type.

matthewhammer: Cool, makes perfect sense.  Thanks for confirming.

chenyan-dfinity: In general, you won't know the type if the parser only sees a number `1`, and you cannot enforce the input to the diff tool to annotate all number types (the parser don't support type annotations yet). For serialization, we always treat `Number` as `Int`, if it is unresolved for whatever reason.

matthewhammer: I've added some features and tests that permit the user to give the type on the command line.  Check out the tests in `candid.rs` to see use cases.

WDYT?

chenyan-dfinity: Parsing as `IDLType` won't work for recursive types, you need a candid file for that. As a first step, it's actually not too bad to treat every number as `Number` without the need for type annotations.

matthewhammer: >  you need a candid file for that

Definitely.  I have left a comment in the code where I'd supply a non-empty type environment (from one or more `.did` inputs)

matthewhammer:
@chenyan-dfinity I started wanting to see a nicer textual representation of the tool's `diff` output.

So, I've started writing pp functions for my type definitions, and also for `IDLValue`, since I couldn't find one (just `Display`, which won't do indentation and line wrapping, etc.).

Should I move those into their own PR, I wonder?

Need to visit the dentist today and switch gears to something else tomorrow, but I feel like the basic use case is almost done (for the BigTest CI script, where I originally wanted to use this diff tool, on the Candid-based log structures.)

candid/src/parser/value.rs
```diff
@@ -319,6 +319,60 @@ impl IDLValue {
     }
 }
 
+pub mod pretty {
```
matthewhammer, line 322: @chenyan-dfinity Is it okay to leave this here (this file/module), and in this PR?  WDYT?

chenyan-dfinity: Yeah, it's fine to do in this PR. Eventually, we can replace the `Display` trait with this pretty version.

chenyan-dfinity:
> There is a fundamental difference between tools like candid_derive, which is Rust-specific, and candiff, which is a general-purpose, language-agnostic tool for Candid itself. Shouldn't this be reflected somehow both in terms of naming conventions for the binaries as well as organisation in this repo?

@rossberg The compiled binary is language-agnostic, but the code for `candiff` as a library is also Rust specific and can be used by other Rust programs. I'm trying to split the `candid` crate into a serialization library and a compiler, but find the same difficulty: They may want to refer to each other as libraries, and it's hard to draw a line where one function is language-agnostic, while the others are not. So I tend to think that all Rust code should be under a directory `rust/`. Protobuf takes the approach, in their github repo, they put both the compiler code and the serialization library under a C++ directory. WDYT?

matthewhammer:
Update: This PR is ready for a more in depth review from someone (like @chenyan-dfinity).

There is lots of room for improving the quality of the vector-based diffs.  That's intentional, and okay.  I just want a MVP here to improve over time, as we use it.

WDYT @chenyan-dfinity ?

### Notes

- The PR now has a suite of tests that demonstrate how it works for each of the basic value forms (`vec`, `variant`, `record`, etc.).

- As promised, the diffs are structural.  They are also pretty printed, though the line-breaks make those test cases less pleasant to author.  I need to find a better workflow for those.  However, the pretty-printed diffs are much easier to read, e.g., for vector diffs that consist of a list of element-wise edits.

- Types are optional, but when the user gives them, they make the numeric representations more specific.  To simplify things here, I try to type check the values if they ask me to, and I return a failure code if _any_ type checking that is requested fails.  That behavior works for the `echo` subcommand too, so this tool can be used (as a special case) to type-check candid values.

- This PR is still somewhat incomplete.  There are missing pieces and questions have formed that I'd like to answer, but they mostly involve the semantics of this tool and would be more clear to work out as inference rules not more Rust code.  For instance, I'm not certain that the behavior is correct for records where one is missing a field, and there's a common purported super type with or without that field, that's given on the command line.  There are tests that show what happens now.  Take a look if you're curious.

candid/src/parser/value.rs
```diff
@@ -319,6 +319,78 @@ impl IDLValue {
...
+    use super::*;
+    use crate::pretty::*;
+
+    use ::pretty::RcDoc;
+
+    pub fn pp_label(id: &Label) -> RcDoc {
+        match id {
+            Label::Named(id) => str("\"")
```
chenyan-dfinity, line 330: You may reuse code here to quote field names only when necessary: https://github.com/dfinity/candid/pull/62/commits/a7df6abddd52f9c23355add965c1279c1b1b3764

candid/src/parser/value.rs
```diff
@@ -319,6 +319,78 @@ impl IDLValue {
...
+            Nat8(n) => RcDoc::as_string(n),
+            Nat16(n) => RcDoc::as_string(n),
+            Nat32(n) => RcDoc::as_string(n),
+            Nat64(n) => RcDoc::as_string(n),
+            Int8(i) => RcDoc::as_string(i),
+            Int16(i) => RcDoc::as_string(i),
+            Int32(i) => RcDoc::as_string(i),
+            Int64(i) => RcDoc::as_string(i),
+            Number(t) => str(t),
+            Text(t) => RcDoc::as_string(format!("{:?}", t)), // to do -- enough quoting here?
```
chenyan-dfinity, line 371: Right, this is tricky to know if we want to escape characters or not. I think it depends on application, so best to have a flag to control this.

chenyan-dfinity:
```suggestion
            Text(t) => RcDoc::as_string(format!("\"{:?}\"", t)), // to do -- enough quoting here?
```

candid/src/parser/value.rs
```diff
@@ -319,6 +319,78 @@ impl IDLValue {
...
+        pp_label(&field.id)
+            .append(kwd(" ="))
+            .append(pp_value(&field.val))
+    }
+
+    pub fn pp_value(v: &IDLValue) -> RcDoc {
+        use super::IDLValue::*;
+        match &*v {
+            Bool(true) => str("true"),
+            Bool(false) => str("false"),
```
chenyan-dfinity, line 357:
```suggestion
            Bool(p) => RcDoc::as_string(p),
```

candid/src/parser/value.rs
```diff
@@ -319,6 +319,78 @@ impl IDLValue {
...
+            Vec(vs) => {
+                let mut body = RcDoc::nil();
+                let mut is_first = true;
+                for v in vs.iter() {
+                    if is_first {
+                        is_first = false;
+                    } else {
+                        body = body.append(RcDoc::space())
+                    }
+                    body = body.append(pp_value(v).append(RcDoc::text(";")))
```
chenyan-dfinity, line 382: I'm a bit confused here. Why is `concat` or `RcDoc::intersperse` not enough here?

candid_diff/src/candiff.rs
```diff
@@ -0,0 +1,302 @@
...
+                                println!("{:?}", v)
+                            } else {
+                                println!("{}", v)
+                            }
+                        }
+                        Err(e) => error!("{}", e),
+                    }
+                }
+                CliFormat::Raw => {
+                    // how is it encoded as a string?
```
chenyan-dfinity, line 186: If we are following the format of `dfx --output raw`, `hex::encode(raw_string)` will be fine.

candid_diff/src/lib.rs
```diff
@@ -0,0 +1,370 @@
+// Logging:
+#[macro_use]
+extern crate log;
+extern crate env_logger;
```
chenyan-dfinity, line 4:
```suggestion
```

Not needed for Rust 2018. If you want to use a macro, you can just `use log::the_macro`;

candid_diff/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+pub use ::candid::parser::value::IDLField as Field;
+pub use ::candid::parser::value::IDLValue as Value;
+
+pub use ::candid::types::{Field as TypeField, Label, Type};
+pub use ::candid::Nat;
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum VecEdit<R> {
+    /// insert into position, shifting the following (backward)
+    InsertValue(Nat, Value),
```
chenyan-dfinity, line 17: What's a position, why it is a BigNat?

candid_diff/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+    /// edit the vector, in a sequence of vector edits.
+    Vec(Vec<VecEdit<R>>),
+    /// edit the record, field-wise; ignore unchanged fields.
+    Record(Vec<RecordEdit<R>>),
+    /// edit the variant payload (ignore unchanged label).
+    Variant(R),
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct RcValueEdit(pub Rc<ValueEdit<RcValueEdit>>);
```
chenyan-dfinity, line 51: Why do we need `Rc<_>`? Is it possible to have a cycle in the editing structure?

candid_diff/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
```
chenyan-dfinity, line 1: I think this file is not needed? We have .gitignore in the top directory.

rossberg:
> @rossberg The compiled binary is language-agnostic, but the code for candiff as a library is also Rust specific and can be used by other Rust programs. I'm trying to split the candid crate into a serialization library and a compiler, but find the same difficulty: They may want to refer to each other as libraries, and it's hard to draw a line where one function is language-agnostic, while the others are not. So I tend to think that all Rust code should be under a directory rust/. Protobuf takes the same approach, in their github repo, they put both the compiler code and the serialization library under a C++ directory. WDYT?

Hm, can't you put all the candiff library code under rust, and then have the code for the standalone CLI program under tools?

candid_diff/src/lib.rs
matthewhammer: Position within the vector to insert the given value.

Are you asking to use a different number type to describe the position, or questioning the meaning of this construct altogether?

matthewhammer:
> Hm, can't you put all the candiff library code under rust, and then have the code for the standalone CLI program under tools?

That's a good question.

I can see the appeal of separating tool code from library code if you think of tools as being conceptually distinct from the libraries that they use in their implementations.

In many cases, however, I want a tool to just be a thin wrapper around a library, and for the tool layer to be minimal.  That way, the (Rust code) can be reused in more ways: By shell users and by Rust developers.  Hence, I assume that every tool that I author is also a library of the same name (or related names).  That's what I'm trying to do here, in this PR.

Given this fact, I don't see the value of having `tools` at all for Rust libraries that happen to have CLI tools for them, as is the case here.  Forcing that separating is possible, but it would require coming up with two names for each library (to distinguish the tool vs library crates, which you are separating by introducing the `tools` folder, IIUC).

Perhaps other tools can go there though?

candid_diff/src/lib.rs
matthewhammer: Rc's permit sharing, not cycles.  No, there are no cycles.  Yes, it's generally nice to permit sharing.  Otherwise, I could use `Box`.  Either way is fine.

candid_diff/src/lib.rs
chenyan-dfinity: Yeah, then certainly don't need to use big number.

matthewhammer:
Regarding the `tools` directory:  Here's another statement of my current stance, and a compromise idea:  https://github.com/dfinity/candid/pull/65#issuecomment-669322461

matthewhammer:
Related question (which `candiff` can answer for all Candid values):
  https://twitter.com/jonathoda/status/1290746222714576896

candid/src/parser/value.rs
matthewhammer: Okay, I duplicated the code because those helpers are all private.  I realize that's not the best solution.

For now, I will make the one I need public (`pp_label`) and re-export it from where I am defining the duplicate above.

Maybe consider making the rest public, in another PR?

candid/src/parser/value.rs
matthewhammer: Your suggestion will give double quotes, which I don't think you want.

Notice that I'm using the Debug formatting string `{:?}`, not `{}`.

This will always include enclosing quotes.  However, it does not escape quotes in the string itself, so the string `\"` will be pretty printed as `"""`, and not `"\""`, which is what I'd want and expect.

candid/src/parser/value.rs
matthewhammer: I wrote this code by hand because the higher order options seemed more confusing, not simpler.

BTW, without an example, `concat` is confusing, and the comment `If it is displayed in a single line, omit the last separator.` is also confusing too.  This behavior is about the length of the sequence, not the 2D layout on the screen (space or line), right?

candid/src/parser/value.rs
matthewhammer: In any case, it seems like a separate PR to add tests that stress the parsing/pretty-printing round trips, and ensure that they always agree.

For this PR, having the pretty printer work at all is enough for an MVP with some reasonable human readable output.  No need for strong mathematical properties (yet).

candid_diff/src/candiff.rs
matthewhammer: Okay, good to know.

This PR is getting gigantic, so I may leave that for another one.

candid/src/parser/value.rs
chenyan-dfinity: Ah, right. I forget debug print has a quote already. For `\n`, I think debug print won't make a new line. We can tune this later.

candid_diff/design/candiff-CLI-tool.md
matthewhammer: Coming back to this question now with some more implementation and testing experience.

Currently, I don't see that `check_type` even supports the subtyping relation.  Does it?

matthewhammer: If not, that's okay.  Perhaps we can revisit this question once the Candid crate itself supports subtyping?

candid_diff/design/candiff-CLI-tool.md
matthewhammer: Agreed --- but see the other outstanding question about checking subtyping.

matthewhammer:
@chenyan-dfinity I did a pass through your comments and made revisions for many.

I'm happy to keep revising if you see anything that I missed, or take issue with anything outstanding.  PTAL when you get a chance.

candid_diff/design/candiff-CLI-tool.md
chenyan-dfinity: No, subtyping is partially supported in serialization. You need two types/two candid files for subtyping to make sense. `check_type` only checks for a single candid file.

candid/src/parser/value.rs
chenyan-dfinity: It's about the 2D layout on screen. Let's say to render `{1;2;3;}`, if it's rendered in a single line, you get `{1;2;3}`, if not, you get
```
{
  1;
  2;
  3;
}
```

PR #60, opened 2020-07-30
New textual representation of Principals

p-shahi:
**Overview**
The textual representation for Principals has changed from `ic:xxxx` to `xxxxx-xxx`

**Considerations**
This is a breaking change with regard to Principal textual format thus `from_text` and `to_text` will no longer work with the old format i.e `ic:xxxx`

### DIFF
--- a/candid/tests/serde.rs
+++ b/candid/tests/serde.rs
@@ -116,8 +116,8 @@ fn test_reserved() {
 fn test_principal() {
     use candid::Principal;
     all_check(
-        Principal::from_text("n7urw-bkejf-ceyaa-bnaaq-hsx75-y").unwrap(),
-        "4449444c000168010c4449444c0001680103caffee",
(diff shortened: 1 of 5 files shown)


candid/tests/serde.rs
```diff
@@ -116,8 +116,8 @@ fn test_reserved() {
 fn test_principal() {
     use candid::Principal;
     all_check(
-        Principal::from_text("ic:caffee59").unwrap(),
-        "4449444c0001680103caffee",
+        Principal::from_text("n7urw-bkejf-ceyaa-bnaaq-hsx75-y").unwrap(),
+        "4449444c000168010c4449444c0001680103caffee",
```
chenyan-dfinity, line 120: The binary bytes doesn't look right. We may need to change the parser as well.

chenyan-dfinity: Okay, this is actually correct, but too confusing as a test. I update to a simpler one.

PR #61, opened 2020-07-30
fix: do not run build.rs everytime we build

hansl:
This includes building crates that depend on candid.


hansl:
This is not the right fix.

Issue #59, opened 2020-07-27
Adjust repo structure or description


rossberg:
Currently specifically says only Rust implementation, is that still accurate?

nomeata:
Ah, that reminds me of an ongoing discussion that we had while you were on vacation, and should delegate to you:

Should this repo become the “official” Candid repo, being the authorative repo to contain the spec, and implementions in different langauges that have some kind of official status? I.e. the Candid-Monorepo? Including the place for design discussions around Candid?

Or should the spec be it’s “own” thing, with its own `candid-spec` repo, and this repo is but one among many repos that contain candid implementations.

rossberg:
Making this the central Candid repo makes sense to me. The name already suggests it is. If it's not, then we should rename it to candid-rust or something.

Were there arguments against?

nomeata:
I disliked blessing one particular implementation, because it dilutes the language-agnostic promise of Candid, it might make one particular implementation (possibly with bugs or idiocracies) canonical, it's easier to tell what discussions (bugs, PRs) are about the spec or the implementation, and it allows us to evolve both independently (having both in the same repo kinda requires changes to the spec to not be merged without changes to the code).

Yan had arguments for having them together; I forgot the details, and we agreed to let you decide.

rossberg:
@nomeata, you don't like monorepos, eh? 👍 

Honestly, I don't feel strongly either way. I hear what you're saying, but maybe Candid is too small to be worth splitting up into per-lang repos. If we structure the implementations into a suitable subdir and have respective README disclaimers, then I think it would be fine if impls lag behind from time to time.

The current repo structure does not meet those requirements, though. All the Rust stuff should move into some implementation/rust subdirectory.

OTOH, if we do not go that route, then we should definitely rename this repo to something more specific.

So some change is needed either way. I have a minor preference for just having one repo, but with clear structure.

nomeata:
Sounds good to me.

More bike shedding :-): I dislike deep directory nesting. How about `/spec`, `/rust`, `/js` (later `/haskell`, maybe `/test` for common test data)?

rossberg:
I don't mind deeper nesting, but fine by me as well. I'd spell it javascript, though.

chenyan-dfinity:
Sounds good to me. I will have a PR for restructuring.

chenyan-dfinity:
[...]

PR #65, opened 2020-08-04
reorganize repo structure

chenyan-dfinity:
Fix https://github.com/dfinity/candid/issues/59

* Update README
* Put everything under three top-level directories: `spec/`, `rust/`, and `docs/`


spec/Candid-Motoko.md
```diff
@@ -1,32 +1,32 @@
-The IDL-Motoko integration
```
rossberg, line 1: Sorry, I hadn't seen this before. I actually think that this file should stay in the Motoko repo. It's not part of the Candid spec.

README.md
```diff
@@ -1,18 +1,36 @@
...
-# Contribution
+## Implementations
+
+Candid supports several different programming languages. 
+This repository contains some of the implementations developed by DFINITY.
+
+* [Rust](rust/): A serialization library based on Serde, and a compiler for generating bindings for other languages.
+* Motoko: Compiler support for importing/export Candid files.
+* JavaScript: As part of the [JavaScript User Library](https://www.npmjs.com/package/@dfinity/agent), we provide a library for serialization of native JavaScript values, and a visitor class for extending Candid for building generic tools such as UI and random testing.
+* More to come...
```
rossberg, line 21: I'd remove that, as I don't think we have any specific plans at this point.

README.md
```diff
@@ -1,18 +1,36 @@
...
+
+A list of community maintained Candid libraries:
+
+* [Haskell](https://github.com/nomeata/haskell-candid)
+* [Elm](https://github.com/chenyan2002/ic-elm/)
+* [Kotlin](https://github.com/seniorjoinu/candid-kt)
```
nomeata, line 27: Oh wow, I didn't konw about this. This makes providing candid test data even more relevant (even a simple one).

spec/Candid-Motoko.md
chenyan-dfinity: It's arguable, and can stay in either repo. I think it has value to stay here, as it provides a good example of how Candid integrates with other languages, and how to deal with type gaps. So I slightly prefer it to stay here. WDYT?

nomeata: Maybe we should have one for Rust and JS (where the implemenation is in this repo), as good examples :-)

chenyan-dfinity: The Rust one is harder to write, as it needs to fight between nominal typing and structural typing. Motoko-Candid is actually the simplistic example.

matthewhammer:
I'm happy to have a `tools` directory so long as it doesn't imply that each Rust library crate with a binary tool (e.g., `candiff`) must be artificially separated into two distinct Rust crates.  That's annoying, if only because we need two distinct names in the Rust package ecosystem.

However, I see the value in a toplevel `tools` directory, that gives a comprehensive view of all CLI tools for Candid, regardless of how they are implemented.

Perhaps I can put shell scripts there that point at the (eventual) binary targets of the Rust crates?

chenyan-dfinity:
It's unclear to me what should be put under `tools/`. We can also put the whole `candiff` code under `tools/`, but tools are not necessarily written in Rust. Should we have `tools/rust/`, `tools/javascript/`, or `rust/tools`, `javascript/tools`?

One thing I would like to be under `tools/` is some kind of test suites/framework that can test different implementations. `candiff` seems to be able to help with the test suites, but maybe we only need the binary, not the library code.

matthewhammer:
> It's unclear to me what should be put under tools/

Also not clear to me, for the reasons you said.  One possible answer is to have it just be symbolic links, or one-line shell scripts, or just a README with directions to the other directories, etc.

rossberg:
> It's unclear to me what should be put under tools/

It's very simple: any tool whose _use_ is not specific to Rust. I think you guys are heavily conflating implementation language with target language. The toplevel structure of this repo should reflect the latter, not the former.

For example, a binding generator for Rust clearly is Rust-specific, so should go into rust/ -- regardless of what language it is implemented in! It could be written in OCaml.

A tool like candiff is not specific to any language, so should go under tools. Again regardless of what language it is written in.

Libraries to be used in a specific language are actually a third kind of area. They could go under rust/lib or something.

spec/Candid-Motoko.md
chenyan-dfinity: @rossberg Do you want to remove this from the repo? I'm fine either way, but need a decision to merge this PR.

rossberg: [Sorry, day off last Friday.] Yes, I think this should be removed here, but that can be a separate PR.

matthewhammer:
>  a binding generator for Rust clearly is Rust-specific, so should go into rust/ -- regardless of what language it is implemented in! It could be written in OCaml.

Oh, thanks for clarifying.  Indeed, I was confused in the way you said: I was thinking that the language directory indicated the implementation language, rather than the "input language" (nit: I wouldn't call it the "target language", since these tools may not have a meaningful notion of that term.  They all should accept input, if even just to check it somehow.)

PR #71, opened 2020-08-07
Official host-language-agnostic Candid test data

nomeata:
the README outlines the format, so worthy of comment already.


nomeata:
This looks promising, doesn’t it?
```
~/dfinity/haskell-candid $ CANDID_TESTS=../candid/test cabal -v0 run test -- -p 'spec'
tests
  Candid spec tests
    File prim
      [6] empty:                              OK
      [7] no magic bytes:                     OK
      [8] wrong magic bytes:                  OK
      [9] wrong magic bytes:                  OK
      [12]:                                   OK
      [13] nullary: too long:                 FAIL
        test.hs:232:
        unexpected decoding success
      [14] Additional parameters are ignored: OK
      [17]:                                   OK
      [18] null: too long:                    FAIL
        test.hs:232:
        unexpected decoding success
      [19] null: missing:                     FAIL
        test.hs:232:
        unexpected decoding success
      [21] bool: true:                        OK
      [22] bool: false:                       OK
      [23] bool: missing:                     OK
      [24] bool: out of range:                OK
      [25] bool: out of range:                OK
      [27] nat: 0:                            OK
      [28] nat: 0x7f:                         OK
      [29] nat: leb (two bytes):              FAIL
        test.hs:233:
        unexpected decoding error:
        Failed reading: overlong encoding
        From:	LEB128
      [30] nat: leb (two bytes, all bits):    OK
      [31] nat: leb too short:                OK
      [32] nat: leb overlong:                 OK

4 out of 21 tests failed (0.00s)
```
(It has color here)

nomeata:
I wrote a few tests and used them in the `haskell-candid` test suite, just to show that this is possible.

I’d suggest we agree on the actual format and then merge this into `master`; integration into the various host languages as well as adding more test cases can then be done in separate PRs.

nomeata:
BTW, the idea of specifying the tests in a grammar that’s an extension of the thing we want to test is of course inspired by the [wasm `.wast` format](https://github.com/WebAssembly/spec/tree/master/interpreter#scripts)

chenyan-dfinity:
Just in time for a CLI: https://github.com/dfinity/candid/pull/72

test/README.md
```diff
@@ -0,0 +1,67 @@
...
+================
+
+This directory contains Candid test data, to be used to check Candid
+implementations for compliance.
+
+This test suite covers
+
+ * parsing Candid service descriptions
+ * given a Candid type, decoding binary Candid data, asserting success or failure
+ * given a Candid type, decoding textual Candid data, asserting success or failure
```
chenyan-dfinity, line 11: What does it mean to "decode a textual data", parsing? Ah, maybe you mean encode the text value at the given type.

test/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Because this test data necessarily has to be host-language agnostic, it is not
+possible to
+
+ * assert the decoded value, as it would be host-language specific
+ * cover encoding, as it would require specifying a host-language value,
```
chenyan-dfinity, line 17: I still would like to cover encoding somehow. Given a textual Candid value and a type, it's at least possible to say if we can encode the value at this type or not. We can also give one possible encoding and use `candiff` to check for equivalence.

test/README.md
```diff
@@ -0,0 +1,67 @@
...
+ * `<name>.good.did`:
+   A Candid service description, with valid candid syntax.
+   Implementations should be able to parse them.
+
+ * `<name>.bad.did`:
+   A Candid service description, with invalid candid syntax.
+   Implementations should reject parsing them.
+
+ * `<name>.test.did`:
+   A set of Candid tests, written in a grammar that extens the Candid type and value grammar:
```
chenyan-dfinity, line 37:
```suggestion
   A set of Candid tests, written in a grammar that extends the Candid type and value grammar:
```

test/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+   The root production is
+   ```
+   <enctest>  ::= <def>;* <test>;*
+   ```
+   to allow any number of type definitions followed by encoding tests with grammar
+   ```
+   <test> = <argseq> <input> <outcome>
```
chenyan-dfinity, line 45:
```suggestion
   <test> = <argseq> <input> <outcome> <desc>
```

test/README.md
```diff
@@ -0,0 +1,67 @@
...
+   <argseq>  ::= ( <argtype>,* )
+   <input> =  <text> | blob <text>
+   <outcome> = true | false
```
chenyan-dfinity, line 48:
```suggestion
   <outcome> = fail | ok <text> | ok possible_blob <text>
```
This is more complete, and users are free to interpret `ok <text>` as `true` without further parsing.

test/README.md
chenyan-dfinity: `<argseq>` can be optional. Unless we are testing numbers or subtyping, we don't need to specify them.

test/README.md
nomeata: I do mean parsing.
```suggestion
 * given a Candid type, parsing textual Candid data at that type, asserting success or failure
```

test/README.md
nomeata: Would be nice to cover it somehow…

So by giving a candid textual value, you are actually testing “textual value parsing → conversion to host value → encoding to candid binary”, when you really care about the second arrow. Might be ok… but it’s actually equivalent to “ binary data encoding → conversion to host value → encoding to candid binary” in that way. So essentially, this is saying that round-tripping works.

I still dislike dependenies on some tools. But shoudn't it be the case that equivalence of two binary values is the same as “both values decode to the same host value”…

Maybe we need to define a candid binary normal form (similar to what we do in https://github.com/dfinity-lab/motoko/pull/1493) :-)

But continuing that thought I still end up with “host langauges should check round-tripping”: If our _decoding_ tests provide good coverage, then host-langaue round-tripping tests (host value → candid binary → host value; check for equality) should check encoding quite well.

test/README.md
nomeata: Not sure it should be optional; some host languages will only support “parsing given an expected type” (motoko, for example). The ability to parse generic binary data, without an expected type, should not be expected from all implementations.

test/README.md
nomeata: I didn't want to add additional keywords to the grammar, this might cause problems when trying to use the same lexer/parser for `.did` and `.test.did` files. We could use `variant { foo = … }`, but that’s such a mouthful (I wish we had `#foo … ` for variant values.)

What should `<text>` be? The round-tripped candid textual representation? Is that unique enough?

test/README.md
```diff
@@ -0,0 +1,67 @@
...
+This test suite covers
+
+ * parsing Candid service descriptions
+ * given a Candid type, decoding binary Candid data, asserting success or failure
+ * given a Candid type, decoding textual Candid data, asserting success or failure
+
+Because this test data necessarily has to be host-language agnostic, it is not
+possible to
+
+ * assert the decoded value, as it would be host-language specific
```
rossberg, line 16: But we have a text representation for values. So can't we at least assert the equality of a binary value against an equivalent text value?

test/README.md
rossberg: It shouldn't be difficult to implement contextual keywords, parsed as identifiers.

test/README.md
rossberg: It might also be useful to compare different representations of the same value. I'd suggest a general assertion like
```
<argseq> <input> `==` <input>
```
Perhaps also useful:
```
<argseq> <input> `!=` <input>
```

nomeata: > But we have a text representation for values. So can't we at least assert the equality of a binary value against an equivalent text value?

Can you give a concrete example?

Do you mean an assertion like `decode  "DIDL\00\00" at type () yields ()`? But that requires the test suite to not only decode candid binary into host langauge values, but also serialize host language values into Candid textual representation. Which is not what we want to test here, and not something we would require of each implementation (do we?)?

For example, Motoko’s RTS currently does not support parsing or printing textual representation.

nomeata: Comparision is interesting! This would allow us to assert, for example, that subtyping ignores a certain field…

I assume the `==` here is the host language’s equality on decoded values. I guess that is a general enough concept to except to be available in most languages – even Motoko(!), once #1812 is done.

test/README.md
nomeata: In that case ok

nomeata:
TODO:

 [ ] use more keywords in the syntax
 [ ] allow to assert host-level equality/inequality between values parsed at the same type

test/README.md
rossberg: > Can you give a concrete example?

I meant something like that comparison.

> I assume the == here is the host language’s equality on decoded values

I think we can directly define an abstract syntax of Candid values, and structural equality on those -- if we couldn't, then Candid wouldn't be very meaningful :).

rossberg: Concretely, with that you can say
```
(Int) blob "..." = 42
```

nomeata: Oh, right.

It will still mean “decoding the binary `DIDL\00\7d\*` yields the same host value as parsing the textual Candid `(42)`”
and not “decoding the binary `DIDL\00\7d\*` yields the host value 42”, but close enough.

nomeata:
@rossberg, care to suggest a syntax that you like and goes well with the rest of the Candid grammer? Either by example, or simply by pushing to this branch?

rossberg:
Hm, I'd suggest
```
assert <input> : <tuptype>
assert <input> !: <tuptype>
assert <input> == <input> : <tuptype>
assert <input> != <input> : <tuptype>
```
where `assert` is a contextual keyword (i.e., lexed as an identifier and checked to equal "assert" in the parser) and
```
<tuptype> ::= <argtype> | ( <argtype>;* )
```

nomeata:
LGTM, Thanks!

I’d prefer to use only
```
<tuptype> ::= ( <argtype>;* )
```
as encoding a value at a single candid type is not supported, and neither are method types `foo : nat -> blob`, so let’s consistenly use `(…)` when talk about these argument sequences.

rossberg:
Makes sense.

nomeata:
Implemented. I like the quality.

spec/Candid.md
```diff
@@ -70,8 +70,9 @@ This is a summary of the grammar proposed:
 
 <actortype> ::= { <methtype>;* }
 <methtype>  ::= <name> : (<functype> | <id>)
-<functype>  ::= ( <argtype>,* ) -> ( <argtype>,* ) <funcann>*
+<functype>  ::= <tuptype> -> <tuptype> <funcann>*
 <funcann>   ::= oneway | query
+<tuptype>   ::= ( <argtype>* )
```
chenyan-dfinity, line 75:
```suggestion
<tuptype>   ::= ( <argtype>,* )
```

test/README.md
```diff
@@ -42,26 +42,42 @@ The test suite contains these kind of files:
...
-   <test> = <argseq> <input> <outcome> <desc>
-   <argseq>  ::= ( <argtype>,* )
-   <input> =  <text> | blob <text>
-   <outcome> = true | false
-   <desc> = <text>?
+   <test> ::=
+      | assert <input> : <tuptype> <desc>
+      | assert <input> !: <tuptype> <desc>
+      | assert <input> == <input> : <tuptype> <desc>
+      | assert <input> != <input> : <tuptype> <desc>
```
chenyan-dfinity, line 49: Also assert subtyping? `assert <tuptype> <: <tuptype> <desc>`

test/README.md
```diff
@@ -42,26 +42,42 @@ The test suite contains these kind of files:
...
+   <input> ::= <text> | blob <text>
```
chenyan-dfinity, line 50:
```suggestion
   <input> ::= <val> | blob <text>
```

nomeata: I thought about this, but I have doubts: Consider a tool that code-generates a test suite. This tool parses the `.test.did` file, and then generate host language code. So we have a two-stage process (this is actually the case in `haskell-candid` as of today; it uses TH to parse the `foo.test.did` file at compile-time, and generates assertions at the right Haskell type.)

It is clear that the parsing of Candid binary values needs to happen in the second stage – after all, the test suite should be generated even if some binaries values are invalid (e.g. "DADL\00\00").

But it seems we should treat the textual notation here analogously: Parsing the the assertions should _not_ require parsing the test _input_, which is data.

It is also a prerequistie to do thing like
```
assert "(true" !: (bool) "missing parenthesis"
```
or
```
assert "{actor = null}" !: (record { "actor" : null}) "unescaped keyword"
```

test/README.md
nomeata: Potentially, although many Candid implementations don't implement subtyping as a relation on types, but only implicitly as “this value can be parsed at this type”. So
```
assert "DIDL\00\02\7e\7f\01)" : (bool)
```
is actually testing subtyping the subtyping of `(bool, null) <: (bool)`.

So I don’t see an urgent need, but we can of course extend the syntax (maybe later, to iterate in smaller steps)?

test/README.md
chenyan-dfinity: Interesting, so the `!=`  means either not equal or syntax error. Maybe we need a `assert_parse` for the later? Usually, `assert` assumes the inputs are syntactically correct, e.g. it's fine to test `blob "DADL\00\00"`, but not `bob "DADL`.

matthewhammer: > Maybe we need a assert_parse for the later?

It seems like it's less about _positively asserting_ the need for a correct parse (it's _always_ needed implicitly for all tests that include text values, so not useful info to assert this need explicitly).

Rather, I think we want to indicate when we do not expect a parse to succeed (`assert_bad_parse`, or something?).  That does seem useful, and missing here.

test/README.md
matthewhammer: > Potentially, although many Candid implementations don't implement subtyping as a relation on types, but only implicitly as “this value can be parsed at this type”

I came across a similar dissonance for `candiff` when it accepts and checks the value against a given type.

Does this check permit a subtyping relation?  A reasonable user may wonder this.

Currently it does not, and @chenyan-dfinity explained why with a similar point (or so it seemed to me).

In that sense, it's overly conservative about what an _implementation of the service at this type_ would do, right?  IIUC, that's the case, and we just don't worry for now?

My concern is that our CLI tools, which are meant to inform, will be confusing on this point if we don't resolve it there, at least.  (Imagine a developer using these tools to predict what candid values can and cannot be sent at a given type, a reasonable use case.)

test/README.md
matthewhammer: > It is also a prerequistie to do thing like [...]

Those examples would dictate official error messages, as strings, for each parse error?

If that's the intention, it seems like:
 - you don't really need the type in those cases, and 
 - you may want to indicate the intention of the test in the assertion kind (e.g., `assert_parse_error`, or something)

matthewhammer: Oh, nevermind --- I think those test case descriptions are not meant to be interpreted by any tools, just humans.  (Right?)

test/README.md
chenyan-dfinity: No, it does support subtyping, e.g. `(42, bool) : (nat) == (42)`. What's not implemented yet is the explicit subtyping relation, i.e. the `<:` relation over types. It's not required for a serialization library, but we do need to support subtyping when encoding/decoding value at a given type.

test/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+ * parsing Candid service descriptions
+ * given a Candid type, decoding binary Candid data, asserting success or failure
+ * given a Candid type, parsing textual Candid data, asserting success or failure
+ * comparing such decoded/parsed values, using the Host language’s notion of equality.
+
+Because this test data necessarily has to be host-language agnostic, it is not
+possible to
+
+ * assert the decoded value, as it would be host-language specific
```
rossberg, line 17: This bullet point is no longer accurate with the new assertions.

test/README.md
rossberg: Allowing plain `val` would make sense as a convenience short-hand, but not instead of, but in addition to the string form -- the latter is still needed for negative tests, as @nomeata says.

@chenyan-dfinity, `:` and `!:` _are_ assert_parse and assert_no_parse, respectively.

test/README.md
```diff
@@ -0,0 +1,84 @@
...
+      | assert <input> == <input> : <tuptype> <desc>
+      | assert <input> != <input> : <tuptype> <desc>
+   <input> ::= <text> | blob <text>
+   <desc> ::=  <text>?
+   ```
+   where the four forms assert
+    * that the input can be decoded/parsed at that type
+    * that the input cannot be decoded/parsed at that type
+    * that the two inputs values decode/parse to the same value
+    * that the two inputs values decode/parse to different values.
```
rossberg, line 58: So this assertion fails if they don't parse/decode? That actually makes sense, but is worth calling out.

test/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+ * its service parser (if present) accepts all `*.good.did` files and rejects
+   all `*.bad.did` files in this directory.
+ * all tests from all `*.test.did` files succeed.
+
+Candid implementations that do not support parsing service files can ignore the
+`*.good.did` and `*.bad.did` files.
+
+Candid implementations that do not support parsing the textual prepresentation
+can ignore those tests where all inputs are textual. They should treat a test
```
rossberg, line 79: Not sure I see the need for this paragraph. Of course, anybody is free to ignore anything, but do you want to encourage not checking the values? If they bother implementing parsing assertions then they can just as well bother implementing parsing of values.

test/README.md
nomeata: > Interesting, so the != means either not equal or syntax error. Maybe we need a assert_parse for the later? Usually, assert assumes the inputs are syntactically correct, e.g. it's fine to test blob "DADL\00\00", but not bob "DADL.

No, `!=` means: Both parse, but to different values. `!:` means assert_no_parse (but does not and should not  distiguish betwen “is totally bogus” and “might parse at a different type”).

> I think those test case descriptions are not meant to be interpreted by any tools, just humans. (Right?)

Yes, see the example output of a test suite futher up for how the description is useful.

test/README.md
nomeata: Hmm, due to the stage difference, this implication doesn't quite hold. For example Motoko RTS does not implement parsing the textual representation (and should not have to). Likey we will write a tool that converts this test suite into a motoko test file, and _that_ tool would have a parser for annotations.

I included this sentence to make sure that we include 
```
assert blob "DIDL\00\01\7e\00" == "(false)" : (bool) "bool: false";
```
in the test suite and don't have to _also_ include
```
assert blob "DIDL\00\01\7e\00" : (bool) "bool: false";
```
for those who can’t make sense of the textual code, and who might initially think they should just ignore all assertions with textual representation anywhere.

test/README.md
nomeata: Rewritten, thanks.

test/README.md
nomeata: Good point, done

test/README.md
rossberg: > Hmm, due to the stage difference, this implication doesn't quite hold.

Well, yes, that test would be run at a different stage by such a tool, as a kind of self-validation.

Anyway, I see what you're getting at, don't think it's necessary, but don't feel strongly either.

PR #72, opened 2020-08-07
Candid CLI

chenyan-dfinity:


chenyan-dfinity:
@matthewhammer Does it make sense to have a `diff` subcommand, so that we only need to maintain one CLI. It also shared a lot of logic, e.g. annotate types, parsing text and binary data.

nomeata:
Nice! It seems already subsumes `hcandid` in terms of features:
```
~/dfinity/dfinity/tests $ hcandid --help
Candid tooling

Usage: hcandid (--decode (--hex ARG | --rust ARG | --stdin) | --encode ARG)

Available options:
  -h,--help                Show this help text
  --decode                 Decode bytes to Candid
  --hex ARG                parse hex data
  --rust ARG               parse text with \xFF escapes
  --stdin                  read data from stdin
  --encode ARG             encode Candid textual form, at inferred type)
```

matthewhammer:
@chenyan-dfinity Nice work.  (Sorry for the late response.  I was recovering from dentist stuff last week.)

> @matthewhammer Does it make sense to have a `diff` subcommand, so that we only need to maintain one CLI. It also shares the same logic, e.g. annotate types, parsing text and binary data. 

I wondered this tool.  I'm glad that you took the initiative so we can see and discuss it further.

At this point, I don't mind merging the CLI tools into a single one, _or_ having several that all share well-organized library code in various crates.

I think we should settle on some CLI design at some point, once the use cases become more clear.  In the meantime, the organization of the Rust library code seems sensible to me now, and that's the most important factor for me.  (also, I think we can even experiment with how to compose the CLI-processing code as library code, if it seems that part becomes redundant across different tools)

I do hesitate about having one exclusive tool, as a matter of design process for now.  If it's okay with you, I'd still like the freedom to have a CLI tool that is dedicated to experimenting with edits:

 - `diff`'ing values, producing edits (what `candiff` does now, and the main CLI does too)
 - given a value and edit, `edit` value and produce updated value (some may call this "patching", in other contexts?)
 - other associated operations that focus on the representation of value edits, including:
    - edit composition (compose two edits, or a stream of them)
    - calculate size metrics for edits and displaying them, etc

I guess I'd like the freedom to do more experiments with edits without some future objection that it's outside the scope of the main `didc` tool, which could be fair (especially since these all concern edits to Candid values, not Candid values themselves).

chenyan-dfinity:
Sounds good. As for code sharing, I think the `TypeAnnotation` struct is certainly worth sharing. I even shared the struct within its own CLI subcommands :)

PR #73, opened 2020-08-11
Candid tests: primitive values

nomeata:
excluding floats for now.

### DIFF
--- a/test/prim.test.did
+++ b/test/prim.test.did
@@ -24,10 +24,125 @@ assert blob "DIDL\00\01\7e"                !: (bool) "bool: missing";
 assert blob "DIDL\00\01\7e\02"             !: (bool) "bool: out of range";
 assert blob "DIDL\00\01\7e\ff"             !: (bool) "bool: out of range";
 
-assert blob "DIDL\00\01\7d\00"                  : (nat) "nat: 0";
-assert blob "DIDL\00\01\7d\7f"                  : (nat) "nat: 0x7f";
+assert blob "DIDL\00\01\7d\00" == "(0)"         : (nat) "nat: 0";
+assert blob "DIDL\00\01\7d\01" == "(1)"         : (nat) "nat: 1";
+assert blob "DIDL\00\01\7d\7f" == "(127)"       : (nat) "nat: 0x7f";
 assert blob "DIDL\00\01\7d\80\01"  == "(128)"   : (nat) "nat: leb (two bytes)";
 assert blob "DIDL\00\01\7d\ff\7f"  == "(16383)" : (nat) "nat: leb (two bytes, all bits)";
 assert blob "DIDL\00\01\7d\80"                 !: (nat) "nat: leb too short";
 assert blob "DIDL\00\01\7d\80\00"              !: (nat) "nat: leb overlong";
+
+assert blob "DIDL\00\01\7c\00" == "(0)"         : (int) "int: 0";
+assert blob "DIDL\00\01\7c\01" == "(1)"         : (int) "int: 1";
+assert blob "DIDL\00\01\7c\7f" == "(-1)"        : (int) "int: -1";
+assert blob "DIDL\00\01\7c\40" == "(-64)"       : (int) "int: -64";
+assert blob "DIDL\00\01\7c\80\01"  == "(128)"   : (int) "int: leb (two bytes)";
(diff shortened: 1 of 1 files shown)


nomeata:
This is already catching bugs in haskell-candid (mostly around the textual representation for now).

test/prim.test.did
```diff
@@ -24,10 +24,125 @@ assert blob "DIDL\00\01\7e"                !: (bool) "bool: missing";
...
 assert blob "DIDL\00\01\7d\80\00"              !: (nat) "nat: leb overlong";
 
+assert blob "DIDL\00\01\7c\00" == "(0)"         : (int) "int: 0";
+assert blob "DIDL\00\01\7c\01" == "(1)"         : (int) "int: 1";
+assert blob "DIDL\00\01\7c\7f" == "(-1)"        : (int) "int: -1";
+assert blob "DIDL\00\01\7c\40" == "(-64)"       : (int) "int: -64";
+assert blob "DIDL\00\01\7c\80\01"  == "(128)"   : (int) "int: leb (two bytes)";
+assert blob "DIDL\00\01\7c\80"                 !: (int) "int: leb too short";
+assert blob "DIDL\00\01\7c\80\00"              !: (int) "int: leb overlong (0s)";
+assert blob "DIDL\00\01\7c\ff\7f"              !: (int) "int: leb overlong (1s)";
```
rossberg, line 42: By default, the LEB definition does not exclude redundant bytes. Why should we not allow them? They are actually quite useful, because they provide a way to back-patch a value later -- Wasm encoders make heavy use of that, e.g. for section sizes.

test/prim.test.did
```diff
@@ -24,10 +24,125 @@ assert blob "DIDL\00\01\7e"                !: (bool) "bool: missing";
...
+assert blob "DIDL\00\01\74\00\00\00\00\00\00"            !: (int64) "int64: too short";
+assert blob "DIDL\00\01\74\00\00\00\00\00\00\00"         !: (int64) "int64: too short";
+
+// TODO: float
+
+
+assert blob "DIDL\00\01\71\00" == "(\"\")"             : (text) "text: empty string";
+assert blob "DIDL\00\01\71\06Motoko" == "(\"Motoko\")" : (text) "text: Motoko";
+assert blob "DIDL\00\01\71\05Motoko"                  !: (text) "text: too long";
+assert blob "DIDL\00\01\71\07Motoko"                  !: (text) "text: too short";
```
rossberg, line 131: Also add tests with overlong string length LEBs?

test/prim.test.did
nomeata: > By default, the LEB definition does not exclude redundant bytes. Why should we not allow them? They are actually quite useful, because they provide a way to back-patch a value later -- Wasm encoders make heavy use of that, e.g. for section sizes.

I have the best rationale ever: Because the spec says so :-)
 
> leb128 and sleb128 are the shortest unsigned and signed LEB128 encodings of a number, respectively

(This was added in https://github.com/dfinity-lab/motoko/commit/e05d25e43511df1986dc8428dfd1403d0e59e8da)

test/prim.test.did
nomeata: Done, and also for the typ table and arg sequence length leb

test/prim.test.did
rossberg: Touche.

Do you think we should relax it? I believe the motivation at the time was to ensure that the wire format is uniquely determined. However, we have since added type tables and stuff, so we have long lost that property.

nomeata: I'm not sure. So far it hasn't bitten us. And I already spent lots of time getting the is-shortest-leb check right in the decoders :-)
 
OTOH there is the principle of being liberal in what you accept…

rossberg: It's also slightly cheaper without the restriction, since a decoder doesn't need to track the length and check it.

nomeata: Anyways, OT for this; if we change the spec we can adjust the tests easily.

chenyan-dfinity: I think we should relax this. It's reasonable to expect people to use existing LEB128 libraries to implement Candid. All the libraries from JS and Rust, including the code form Wikipedia, allow to decode this. Seems Motoko is the only decoder that enforces this check. When communicating with Rust canisters or frontend, we may get bitten by this.

PR #74, opened 2020-08-11
Candid tests: Overshooting sequences

nomeata:
The tests in this file involve large leb-encoded values that indicate sizes (of
tables, arrays, blobs, etc.). In all cases, a decode should be able to catch
that eary by noticing that the overall input is too short to potentially
contain that number of elements.

The number is always one billion (leb "\80\94\eb\dc\03"). A test suite could run
these tests with a memory limit, of, say 100MB and should still never run into
out of memory conditions, if these eager plausability checks are performed.


test/overshot.test.did
```diff
@@ -0,0 +1,27 @@
+/*
```
rossberg, line 1: Nit: `overshoot.test.did`?

nomeata: Coudn’t make up my mind :-)

PR #78, opened 2020-08-11
parser for test suite

chenyan-dfinity:
* Support comments (no nesting)
* Blob shorthand
* Fix string parser
* Reserved value

### DIFF
--- a/rust/candid/src/parser/grammar.lalrpop
+++ b/rust/candid/src/parser/grammar.lalrpop
@@ -29,9 +29,11 @@ extern {
         "type" => Token::Type,
         "import" => Token::Import,
         "principal" => Token::Principal,
-        "+" => Token::Plus,
-        "-" => Token::Minus,
+        "sign" => Token::Sign(<char>),
         "=" => Token::Equals,
+        "==" => Token::TestEqual,
+        "!=" => Token::NotEqual,
+        "!:" => Token::NotDecode,
         "(" => Token::LParen,
         ")" => Token::RParen,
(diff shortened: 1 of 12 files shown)


rust/candid/src/parser/grammar.lalrpop
```diff
@@ -220,20 +225,20 @@ pub IDLProg: IDLProg = {
...
 };
 
-Assertion: Assertion = {
-    <input:Input> ":" <typ:TupTyp> => Assertion::Decode(input, typ),
-    <input:Input> "!" ":" <typ:TupTyp> => Assertion::NotDecode(input, typ),
-    <left:Input> "=" "=" <right:Input> ":" <typ:TupTyp> => Assertion::Equal(left, right, typ),
-    <left:Input> "!" "=" <right:Input> ":" <typ:TupTyp> => Assertion::NotEqual(left, right, typ),
+Assertion: Assert = {
+    <left:Input> ":" <typ:TupTyp> <desc:"text"?> => Assert { left, right: None, typ, pass: true, desc },
+    <left:Input> "!" ":" <typ:TupTyp> <desc:"text"?> => Assert { left, right: None, typ, pass: false, desc },
```
nomeata, line 239: Would this allow
```
assert blob "foo" ! : ()
```
with a space inbetween `!:`?

Woudl `"!:"` work differently?

chenyan-dfinity: Good catch. Better to have a dedicated token then.

PR #79, opened 2020-08-13
[Spec] Relax LEB128 decoding

chenyan-dfinity:
As discussed in https://github.com/dfinity/candid/pull/73#discussion_r468460144, LEB128 spec allows redundant bytes, and all existing libraries on LEB128 decodes redundant bytes properly.  Removing the restriction encourages code reuse and an easier implementation.


nomeata:
I'll have to update http://hackage.haskell.org/package/leb128-cereal and the Motoko implemenation, but that’s fine of course.

@chenyan-dfinity care to update the spec tests together with this PR?

nomeata:
Why did we not bump the Candid version with this PR?

chenyan-dfinity:
I was going to bump the version here https://github.com/dfinity/candid/pull/88. The change documented in the Changelog.md already.

PR #83, opened 2020-08-19
more candid test data

chenyan-dfinity:
* tests for construct types and floats
* validation for type table
* handle `type T = record { T }` properly
* `annotate_types` takes a `from_parser` flag


test/construct.test.did
```diff
@@ -0,0 +1,152 @@
...
+assert blob "DIDL\01\6c\02\00\01\01\7e\01\7c\2a"                               !: (reserved) "record: type out of range";
+
+// variant
+assert "(variant {})"                                                          !: (variant {}) "variant: no empty value";
+assert blob "DIDL\01\6b\00\01\00"                                              !: (variant {}) "variant: no empty value";
+assert blob "DIDL\01\6b\01\00\7f\01\00\00" == "(variant {0})"                   : (variant {0}) "variant: numbered field";
+assert blob "DIDL\01\6b\01\00\7f\01\00\00\2a"                                  !: (variant {0:int}) "variant: type mismatch";
+
+// TODO Should this pass?
+assert blob "DIDL\01\6b\02\00\7f\01\7c\01\00\01\2a"                             : (variant {0:int; 1:int}) "??? variant: type mismatch";
```
nomeata, line 92: Sorry, too lazy to decode. What is unclear here?

chenyan-dfinity: The blob says `variant { 1 = 42 } : variant { 0: null; 1: int }`, and we want to decode at type `variant { 0:int; 1:int}`. Do we want to care variant type mismatch when the value doesn't use that branch?

nomeata: Oh, right. That was (is?) discussed in https://github.com/dfinity-lab/motoko/issues/870. I also very much vote against requiring every decode to implement a full subtyping checker, which is a non-trivial piece of code, especially for recursive types.

nomeata: Let’s continue in https://github.com/dfinity-lab/motoko/pull/1830 maybe

PR #88, opened 2020-08-25
add service initialization parameters

chenyan-dfinity:
Discussed in https://github.com/dfinity-lab/motoko/pull/1549. This adds initialization parameters to the higher order case as well. Since this requires changing the binary format, better to change it now than later.

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -862,7 +862,7 @@ T(func (<datatype1>*) -> (<datatype2>*) <funcann>*) =
   sleb128(-22) T*(<datatype1>*) T*(<datatype2>*) T*(<funcann>*) // 0x6a
 T(service (<datatype>*) -> {<methtype>*}) =
   sleb128(-23) T*(<datatype>) T*(<methtype>*)                     // 0x69
-T(principal)= sleb128(-24)                                        // 0x68
+T(principal) = sleb128(-24)                                       // 0x68
 
 T : <methtype> -> i8*
 T(<name>:<datatype>) = leb128(|utf8(<name>)|) i8*(utf8(<name>)) I(<datatype>)
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -65,9 +65,9 @@ This is a summary of the grammar proposed:
 ```
 <prog>  ::= <def>;* <actor>;?
 <def>   ::= type <id> = <datatype> | import <text>
-<actor> ::= service <id>? : (<actortype> | <id>)
+<actor> ::= service <id>? : (<tuptype> ->)? (<actortype> | <id>)
 
-<actortype> ::= { <methtype>;* } | <tuptype> -> { <methtype>;* }
+<actortype> ::= { <methtype>;* }
 <methtype>  ::= <name> : (<functype> | <id>)
 <functype>  ::= <tuptype> -> <tuptype> <funcann>*
 <funcann>   ::= oneway | query
@@ -139,6 +139,9 @@ Block comments nest properly (unlike in C).
 
 A *service* is a standalone actor on the platform that can communicate with other services via sending and receiving *messages*. Messages are sent to a service by invoking one of its *methods*, i.e., functions that the service provides.
 
+For the main service, there are two kinds of services: service constructor (or uninitialized service) and running service (or initialized service).
+A service constructor takes *initialization parameters* for installing the service on the platform. Once it is initialized, the service becomes a running service. We can initialize a service constructor to multiple running services.
+
 **Note:** Candid is in fact agnostic to the exact nature of services. In particular, it could be applied to a setting where services are synchronous (objects with RPCs) instead of asynchronous (actors with bidirectional message sends).
 
 
@@ -146,11 +149,8 @@ A *service* is a standalone actor on the platform that can communicate with othe
 
 A service's signature is described by an *actor type*, which defines the list of *methods* that the service provides. Each method is described by its *name* and a *function type* describing its signature. The function type can also be given by referring to a type definition naming a function reference type.
 
-There are two stages of a service: uninitialize and installed. An uninitialized service takes *initialization parameters*
-for installing service on the platform. Once it is installed, the parameters are removed from the service signature. 
-
 ```
-<actortype> ::= { <methtype>;* } | <tuptype> -> { <methtype>;* }
+<actortype> ::= { <methtype>;* }
 <methtype>  ::= <name> : (<functype> | <id>)
 ```
 We identify `<methtype>` lists in an actor type up to reordering.
@@ -806,8 +806,6 @@ Likewise, there are two forms of Candid values for function references:
 * `ref(r)` indicates an opaque reference, understood only by the underlying system.
 * `pub(s,n)`, indicates the public method name `n` of the service referenced by `s`.
 
-Note that service references for uninitialized services are not supported at the moment.
-
 #### Notation
 
 `T` and `M` create a byte sequence described below in terms of natural storage types (`i<N>` for `N = 8, 16, 32, 64`, `f<N>` for `N = 32, 64`). The bytes are sequenced according to increasing significance (least significant byte first, a.k.a. little-endian).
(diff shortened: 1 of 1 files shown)


spec/Candid.md
```diff
@@ -144,10 +144,12 @@ A *service* is a standalone actor on the platform that can communicate with othe
 
 #### Structure
 
-A service's signature is described by an *actor type*, which defines the list of *methods* that the service provides. Each method is described by its *name* and a *function type* describing its signature. The function type can also be given by referring to a type definition naming a function reference type.
+A service's signature is described by an *actor type*, which defines the list of *methods* that the service provides. 
+A service can optionally provide *initialization parameters* for creating services dynamically with the same signature.
+Each method is described by its *name* and a *function type* describing its signature. The function type can also be given by referring to a type definition naming a function reference type.
```
rossberg, line 149: This sentence elaborates on the one above, so should probably stay directly after.

spec/Candid.md
```diff
@@ -144,10 +144,12 @@ A *service* is a standalone actor on the platform that can communicate with othe
 
 #### Structure
 
-A service's signature is described by an *actor type*, which defines the list of *methods* that the service provides. Each method is described by its *name* and a *function type* describing its signature. The function type can also be given by referring to a type definition naming a function reference type.
+A service's signature is described by an *actor type*, which defines the list of *methods* that the service provides. 
+A service can optionally provide *initialization parameters* for creating services dynamically with the same signature.
```
rossberg, line 148: Hm, it may be good to make the difference more explicit. The parameters aren't just "optional". If they are present, the type describes a different kind of entity. That probably deserves a separate paragraph.

spec/Candid.md
```diff
@@ -856,8 +858,8 @@ T(<nat>:<datatype>) = leb128(<nat>) I(<datatype>)
 T : <reftype> -> i8*
 T(func (<datatype1>*) -> (<datatype2>*) <funcann>*) =
   sleb128(-22) T*(<datatype1>*) T*(<datatype2>*) T*(<funcann>*) // 0x6a
-T(service {<methtype>*}) =
-  sleb128(-23) T*(<methtype>*)                                    // 0x69
+T(service (<datatype>*) -> {<methtype>*}) =
+  sleb128(-23) T*(<datatype>) T*(<methtype>*)                     // 0x69
 T(principal)= sleb128(-24)                                        // 0x68
```
rossberg, line 863:
```suggestion
T(principal) = sleb128(-24)                                       // 0x68
```

spec/Candid.md
```diff
@@ -144,10 +144,12 @@ A *service* is a standalone actor on the platform that can communicate with othe
...
 #### Structure
 
-A service's signature is described by an *actor type*, which defines the list of *methods* that the service provides. Each method is described by its *name* and a *function type* describing its signature. The function type can also be given by referring to a type definition naming a function reference type.
+A service's signature is described by an *actor type*, which defines the list of *methods* that the service provides. 
+A service can optionally provide *initialization parameters* for creating services dynamically with the same signature.
+Each method is described by its *name* and a *function type* describing its signature. The function type can also be given by referring to a type definition naming a function reference type.
 
 ```
-<actortype> ::= { <methtype>;* }
+<actortype> ::= { <methtype>;* } | <tuptype> -> { <methtype>;* }
```
rossberg, line 152: We should also allow forming
```
... | <tuptype> -> <id>
```
where `<id>` refers to an initialised actor type.

rossberg (commented):
Don't you need to say something in the serialisation and value sections as well? Even if it is just that references uninitialised services are not supported at the moment.

spec/Candid.md
chenyan-dfinity: Should we? It's symmetric with functions, which don't allow `<tuptype> -> <id>`. What does `(nat) -> service (int) -> { f: () -> () }` mean? Initializing this service gets another uninitialized service?

spec/Candid.md
chenyan-dfinity: I wonder if `{ <methtype>;* }` should just be a sugar for `() -> { <methtype>;* }`? So the presence of parameters represents different stages of the service: uninitialized vs installed

spec/Candid.md
rossberg: Oh, I didn't suggest that you can nest them, thus _initialised_ actor type. Only that you can write this:
```
type A = service {...};
type B = service (Nat) -> A;
```
The `id` can only be a non-arrow type, i.e.,
```
type C = service (Nat) -> B;
```
would be an error.

spec/Candid.md
rossberg: In what context? In general, they are totally different things, the former typing a canister reference, the latter a Wasm module. They are no more related than `Nat` and `() -> Nat`.

crusso:
Hmm, I don't see why we are going higher-order now, when it is unnecessary and won't be supported for a while. What was wrong with the original proposal from the discussion:

```
<prog>  ::= <def>;* <actor>;?
<def>   ::= type <id> = <datatype> | import <text>
<actor> ::= service <id>? : (<tuptype> ->?)  (<actortype> | <id>)
```
without extending `<actortype>` itself.

(https://github.com/dfinity-lab/motoko/pull/1549#issuecomment-677493520)

This doesn't require introducing (first-class) types for classes (nor any wire format) and just adds the possibility of parameterising the main service.

spec/Candid.md
chenyan-dfinity: In the sense that the did file generated by Motoko is only a Wasm module. It becomes a canister only after installation. So the actor from Motoko is always `() -> service`.

chenyan-dfinity:
> I don't see why we are going higher-order now, when it is unnecessary and won't be supported for a while.

Because the higher order case is a breaking change for wire format. Better to change the format sooner? Otherwise, we can just extend the main actor.

crusso:
> I don't see why we are going higher-order now, when it is unnecessary and won't be supported for a while.
> 
> Because the higher order case is a breaking change for wire format. Better to change the format sooner? Otherwise, we can just extend the main actor.

Fair enough, maybe that should be @rossberg's call.

rossberg:
I'm fine "reserving" the syntax and wire format already. I'm also fine restricting for now. I'm not yet concerned about adding extensions to the format that are technically breaking changes. Ones we're live that becomes more of an issue.

What would it actually take to support the higher-order case? The wire format of a service constructor would just be the blob containing the Wasm module, wouldn't it? If that's all then it seems pretty straightforward.

crusso:
> I'm fine "reserving" the syntax and wire format already. I'm also fine restricting for now. I'm not yet concerned about adding extensions to the format that are technically breaking changes. Ones we're live that becomes more of an issue.

> What would it actually take to support the higher-order case? The wire format of a service constructor would just be the blob containing the Wasm module, wouldn't it? If that's all then it seems pretty straightforward.

Yeah, I'm not sure about that. Ideally you'd probably just want it to be the id/address of some installed code, not the code itself, but who knows.  The current model of literally passing the code around and duplicating at each instantiation seems unfortunate to me, though it's the only method of instantiation we have at the moment. To my mind, installing a service module should install the code once and return a reference to a constructor that peels off fresh instances sharing the code, but  providing new instance data/state.

rossberg:
Yeah, I agree that it would be nicer if the platform had a way to reference modules, though I doubt we'll see much excitement for that from the execution team. The easiest way to get there might be by providing an inefficient version of the feature and wait for performance pressure from actual use cases. :) Then we can still extend the wire format to support both ways of passing it.

chenyan-dfinity:
If breaking changes are not a concern for now, I would prefer to restrict the change to main actor for now. I don't fully understand the higher order case, as it is tied to the question of where we store the candid file. PTAL.

chenyan-dfinity:
Do we want to merge this?

nomeata:
I guess. I wasn’t part of the discussion the last two weeks, so maybe @crusso can make that call. But the PR as it stands seems innocent enough.

PR #89, opened 2020-08-26
Refactor Lexer

chenyan-dfinity:
* Use `logos` lexer to replace the hard-coded state machine.
* Support float e notation
* Support nested comments
* Experiment with `codespan-reporting` in CLI


tools/didc/src/main.rs
```diff
@@ -115,11 +117,34 @@ fn check_file(env: &mut TypeEnv, file: &Path) -> candid::Result<Option<Type>> {
...
+    }
+}
+
+fn main() -> Result<()> {
+    let writer = StandardStream::stderr(term::termcolor::ColorChoice::Auto);
+    let config = term::Config::default();
+
     match Command::from_args() {
         Command::Check { input } => {
+            let file = SimpleFile::new(input.to_str().unwrap(), std::fs::read_to_string(&input)?);
```
kritzcreek, line 139: Without an actual filemap this doesn't do much. Did you want to merge this experiment, or should this be its own PR?

rust/candid/src/parser/token.rs
```diff
@@ -0,0 +1,277 @@
...
+    StartString,
+    // This token is not derived. Stores the unescaped string
+    Text(String),
+    #[regex("[+-]", |lex| lex.slice().chars().next())]
+    Sign(char),
+    #[regex("[0-9][_0-9]*", parse_number)]
+    Decimal(String),
+    #[regex("0[xX][0-9a-fA-F][_0-9a-fA-F]*", parse_number)]
+    Hex(String),
+    #[regex("[0-9]*\\.[0-9]*", parse_number)]
```
kritzcreek, line 75: This'll accept leading zeroes, do we want that?

kritzcreek (approved):
Nice! Any idea how the compile times look?

rust/candid/src/parser/token.rs
chenyan-dfinity: This is the same as the Motoko lexer, and the spec says it's should be the same as Motoko :)

tools/didc/src/main.rs
chenyan-dfinity: I think it's fine to merge it, as it doesn't touch the main candid crate. 

I actually don't quite like the filemap idea, it seems an overkill to store all files we ever touched in memory, just for reporting a nice error message. But I don't have better ideas neither.

chenyan-dfinity:
> Any idea how the compile times look?

I didn't notice any compile time increase. Admittedly, `lalrpop` takes a lot of time to compile, so other increases all seem negligible :)

PR #90, opened 2020-08-26
Fix: candiff accepts unary tuples as default

matthewhammer:
The candid values saved by`dfx` via `didc` were not accepted by `candiff` to diff, because of some mismatch of input being a tuple versus a (bare) un-tupled value.

This PR fixes this issue by changing the default behavior:
 - introduces an unimplemented flag for accepted bare values (the old default).
 - the new default matches what `didc` produces.

We can implement the old behavior when it blocks something.  The missing code path traps as `unimplemented`.

### DIFF
--- a/tools/candiff/src/candiff.rs
+++ b/tools/candiff/src/candiff.rs
@@ -15,6 +15,7 @@ use candid::parser::typing::{check_type, Env, TypeEnv};
 
 use candiff::{
     //Type,
+    ArgTuple,
     pretty,
     Value,
 };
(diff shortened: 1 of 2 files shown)


chenyan-dfinity (approved):
btw, the CLI currently has a lot of code for handling parsing errors, which is not necessary. See https://github.com/dfinity/candid/blob/master/tools/didc/src/main.rs#L50. `StructOpt` will the handle the error automatically if you say the `input` expects `IDLArgs` instead of `String`

matthewhammer:
Thanks.  Will fix the failing tests (tomorrow) before merging, obviously.  I forgot to run `cargo test` locally first.

PR #93, opened 2020-09-01
Parse annvals in textual format

osa1:
The annval production is as described by the spec [1]:

    <annval> ::=
      | <val>
      | <val> : <datatype>

This production gives us concrete syntax for numeric types like nat32,
int64 etc. Currently all numbers are parsed to uninterpreted numbers.

[1]: https://github.com/dfinity/candid/blob/master/spec/Candid.md#text-format


rust/candid/src/parser/typing.rs
```diff
@@ -261,6 +264,88 @@ fn check_actor(env: &Env, actor: &Option<IDLType>) -> Result<Option<Type>> {
...
+            match prim_typ {
+                PrimType::Nat => {
+                    let biguint = BigUint::from_str(number.as_str()).map_err(error)?;
+                    Ok(IDLValue::Nat(Nat(biguint)))
+                }
+                PrimType::Int => {
+                    let bigint = BigInt::from_str(number.as_str()).map_err(error)?;
+                    Ok(IDLValue::Int(Int(bigint)))
+                }
+                // NOTE: Concrete syntax for numbers is not specified in candid spec, but the
```
nomeata, line 288: Spec says it’s same as Motoko, which I believe is the same as Wasm’s `.wat`, I believe, i.e. https://webassembly.github.io/spec/core/text/values.html.

nomeata (commented):
I’ll defer to @chenyan-dfinity for approval

rust/candid/src/parser/typing.rs
osa1: Where do we say that Motoko's nat/int syntax is the same as .wat syntax? I checked https://sdk.dfinity.org/docs/language-guide/motoko-grammar.html but couldn't see it.

nomeata: Might have been “personal communication” (or just me misremembering)

rust/candid/src/parser/grammar.lalrpop
```diff
@@ -81,6 +81,7 @@ pub Arg: IDLValue = {
     },
     "variant" "{" <VariantField> "}" => IDLValue::Variant(Box::new(<>), 0),
     "principal" <Text> =>? Ok(IDLValue::Principal(Principal::from_text(<>).map_err(error)?)),
+    AnnVal => <>,
```
chenyan-dfinity, line 84: From the spec, you also need to update line 51 to use `AnnVal` for `IDLArgs`, so that you can avoid the top level parenthesis.

rust/candid/src/parser/grammar.lalrpop
```diff
@@ -102,6 +103,11 @@ Number: String = {
     "hex" => num_bigint::BigInt::parse_bytes(<>.as_bytes(), 16).unwrap().to_str_radix(10),
 }
 
+AnnVal: IDLValue = {
+    "(" <Arg> ")" => <>,
```
chenyan-dfinity, line 107: Move the parenthesis up to `IDLValue`

rust/candid/src/parser/typing.rs
```diff
@@ -261,6 +264,90 @@ fn check_actor(env: &Env, actor: &Option<IDLType>) -> Result<Option<Type>> {
     }
 }
 
+pub(crate) fn interpret_number(number: IDLValue, typ: IDLType) -> Result<IDLValue> {
```
chenyan-dfinity, line 267: There is a `annotate_type` function in `value.rs`. It's better to reuse that function. We can convert the `IDLType` to `Type` by calling `typing::check_type`. This way, we can give arbitrary type annotations, not just restricting to numbers.

rust/candid/src/parser/grammar.lalrpop
osa1: > so that you can avoid the top level parenthesis

You mean this production in the spec, right?
```
<arg> ::= ( <annval>,* )
```
These parens don't look like they're optional. Are they supposed to be optional?

rust/candid/src/parser/grammar.lalrpop
osa1: Done.

rust/candid/src/parser/typing.rs
osa1: Oh wow.. I wasn't aware of that function. Thanks! Indeed it makes sense to use that one. Will update.

osa1: One problem is these functions need an environment (`typing::Env` for `chekc_type`, `TyEnv` for `annotate_type`) which we don't have during parsing. Should I create an empty environment? Doesn't make sense to me and the error messages will probably suffer in some edge cases, but perhaps that's fine for now.

nomeata: An empty environment is probably the right thing (for now). These annotations in textual values, by construction, cannot refer to named types (there is no place to name them – yet).

osa1: OK, just did that. There's a new problem though which is `annotate_type` doesn't allow annotating e.g. "123" with float32, which interpret_number allowed. So we again don't have a syntax for floats.

Should `annotate_type` allow annotating a number with a float type?

rust/candid/src/parser/grammar.lalrpop
chenyan-dfinity: Yes, the parens inside `annval` is optional. You can say `(42 : int, 42 : nat)`, `((42 : int), (42 : nat))` is not required.

rust/candid/src/parser/typing.rs
chenyan-dfinity: `123.` represents a float. I think this is consistent with Motoko.

rust/candid/tests/parse_value.rs
```diff
@@ -222,3 +227,29 @@ fn parse_shorthand() {
...
+    let args = parse_args("((1))");
+    assert_eq!(
+        args,
+        IDLArgs {
+            args: vec![IDLValue::Number("1".to_string())]
+        }
+    );
+
+    let args =
+        parse_args("((1 : nat), (123 : int), (456 : int32), (789 : float32), (1011 : float64))");
```
chenyan-dfinity, line 242: This is what we should be able to parse

```suggestion
        parse_args("(1 : nat, (123 : int), 456 : int32, 789. : float32, 1011.0 : float64)");
```

chenyan-dfinity:
Let me merge this, so that I can make a release

osa1:
Thanks @chenyan-dfinity for finishing this, and sorry for not being more responsive. I was busy with other work and only had time to look at this PR now..

PR #99, opened 2020-09-10
Add generic functions to encode/decode around a tuple

hansl:
This is follow-up and update to PR #38. It is backward-compatible, but
we may want to deprecate the macros as they are harder to use and non
type safe in some edge cases.

Tests were modified to ensure that the macros and new decode/encode
functions are fully compatible.

Please note that we don't verify that the decoder decoded all values, as decoding should be covariant on arguments.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -1,5 +1,3 @@
-#![allow(unused_attributes)]
-#![rustfmt::skip::macros(decode_impl)]
 //! Deserialize Candid binary format to Rust data structures
 
 use super::error::{Error, Result};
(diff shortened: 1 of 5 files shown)


rust/candid/src/de.rs
```diff
@@ -843,3 +843,184 @@ impl<'de, 'a> de::VariantAccess<'de> for Compound<'a, 'de> {
...
+/// assert_eq!(golden1, value1);
+/// assert_eq!(golden2, value2);
+/// ```
+pub fn decode_args<'a, Tuple>(bytes: &'a [u8]) -> Result<Tuple>
+where
+    Tuple: ArgumentDecoder<'a>,
+{
+    let mut de = IDLDeserialize::new(bytes)?;
+    let res = ArgumentDecoder::decode(&mut de)?;
+    Ok(res)
```
chenyan-dfinity, line 1004:
```suggestion
    de.done()?;
    Ok(res)
```
`done()` ensures the remaining bytes are proper Candid encoding, a good safe guard.

rust/candid/src/de.rs
```diff
@@ -843,3 +843,184 @@ impl<'de, 'a> de::VariantAccess<'de> for Compound<'a, 'de> {
...
+/// ```
+/// # use candid::Encode;
+/// # use candid::de::decode_one;
+/// let golden1 = 123u64;
+/// let bytes = Encode!(&golden1).unwrap();
+/// let value1: u64 = decode_one(&bytes).unwrap();
+///
+/// assert_eq!(golden1, value1);
+/// ```
+pub fn decode_one<'a, T>(bytes: &'a [u8]) -> Result<T>
```
chenyan-dfinity, line 1020: Non-blocking, but I wonder if we can implement `ArgumentDecoder` for any `Deserialize`, so that we can avoid special case for one argument.

rust/candid/tests/serde.rs
```diff
@@ -492,7 +492,11 @@ where
 }
 
 fn encode<T: CandidType>(value: &T) -> Vec<u8> {
```
chenyan-dfinity, line 494: Update the `decode` test as well.

chenyan-dfinity (commented):
LGTM. The macro is never intended to be for generic use. It's like `vec!`, convenient to use for fixed interfaces like the management canister. The generic case should always use the builder pattern. Fair enough to provide a convenience wrapper for sequence of arguments.

rust/candid/src/de.rs
hansl: Once specialization makes it past unstable, we could use it for exactly that. See https://github.com/rust-lang/rust/issues/31844

Unfortunately the type system is currently limited and trying to do a generic `impl<T> ArgumentDecoder for T` will conflict with the existing implementations that include tuples.

hansl:
@chenyan-dfinity Addressed all your comments. Thanks for the review!

chenyan-dfinity (approved):
Thanks!

PR #102, opened 2020-09-12
Add a type to the ArgumentDecoder to know its first arg

hansl:
This is the return value if one would call decode_one() on the same
bytes as the ones for decode_args. This is useful when you have a
type of DecodeArguments<> but only need the first argument.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -846,12 +846,19 @@ impl<'de, 'a> de::VariantAccess<'de> for Compound<'a, 'de> {
 
 /// Allow decoding of any sized argument.
 pub trait ArgumentDecoder<'a>: Sized {
+    /// The types returned when decoding [Self] using [decode_one].
+    type FirstArg;
+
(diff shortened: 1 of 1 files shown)


rust/candid/src/de.rs
```diff
@@ -847,30 +847,41 @@ impl<'de, 'a> de::VariantAccess<'de> for Compound<'a, 'de> {
...
+
     /// Decodes a value of type [Self], modifying the deserializer (values are consumed).
     fn decode(de: &mut IDLDeserialize<'a>) -> Result<Self>;
 }
 
 /// Decode an empty tuple.
 impl<'a> ArgumentDecoder<'a> for () {
+    /// There is loss of type information when using `(())`, but that's expected and an
+    /// edge case.
+    type FirstArg = ();
```
chenyan-dfinity, line 861: `()` means zero argument. So `FirstArg` is either `None`, `!`, or `candid::Empty`.

chenyan-dfinity (commented):
Can you give an example where this is useful? The code change looks fine, but I don't understand why we need a `FirstArg`.

PR #110, opened 2020-10-01
[spec] Reverse subtyping

rossberg:
This (finally) adds the ability to do reverse subtyping on records and variants. I tried to include various motivation, explanation, and some simple examples to make those rather exotic rules clearer.

Also filled in the missing pieces of lexical syntax for values.

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -98,12 +98,8 @@ This is a summary of the grammar proposed:
   | principal
 
 <name> ::= <id> | <text>
-<id>   ::= (A..Z|a..z|_)(A..Z|a..z|_|0..9)*
-<text> ::= "<char>*"
-<nat>  ::= (0..9)(_? 0..9)* | 0x(0..9|a..f|A..F)(_? 0..9|a..f|A..F)*
 ```
-
-A `<char>` is a *Unicode scalar value* (i.e., a codepoint that is not a surrogate part).
+See [below](#values) for the definitions of `<id>`, `<nat>`, and `<text>`.
 
 
 #### Syntactic Shorthands
@@ -538,6 +534,96 @@ An *interface description* consists of a sequence of imports and type definition
 The optional name given to the service in an interface description is immaterial; it only serves as documentation.
 
 
+## Values
+
+To enable convenient debugging, the following grammar specifies a text format for values.
+The types of these values are assumed to be known from context, so the syntax does not attempt to be self-describing.
+
+```
+<val> ::=
+  | <primval> | <consval> | <refval>
+  | ( <annval> )
+
+<annval> ::=
+  | <val>
+  | <val> : <datatype>
+
+<primval> ::=
+  | <nat> | <int> | <float>
+  | <text>
+  | true | false
+  | null
+
+<consval> ::=
+  | opt <val>
+  | vec { <annval>;* }
+  | record { <fieldval>;* }
+  | variant { <fieldval> }
+
+<fieldval> ::= <nat> = <annval>
+
+<refval> ::=
+  | service <text>             (canister URI)
+  | func <text> . <id>         (canister URI and message name)
+  | principal <text>           (principal URI)
+
+<arg> ::= ( <annval>,* )
+
+<letter> ::= A..Z | a..z
+<digit>  ::= 0..9
+<id>     ::= (<letter> | _)(<letter> | <digit> | _)*
+
+<sign>   ::= + | -
+<hex>    ::= <digit> | A..F | a..f
+<num>    ::= <digit>(_? <digit>)*
+<hexnum> ::= <hex>(_? <hex>)*
+<nat>    ::= <num> | 0x<hexnum>
+<int>    ::= <sign>? <num>
+<float>  ::=
+  | <sign>? <num> . <num>?
+  | <sign>? <num> (. <frac>?)? (e | E) <sign>? <num>
+  | <sign>? 0x<hexnum> . <hexnum>?
+  | <sign>? 0x<hexnum> (. <hexnum>?)? (p | P) <sign>? <num>
+
+<text>   ::= " <char>* "
+<char>   ::=
+  | <utf8>
+  | \ <hex> <hex>
+  | \ <escape>
+  | \u{ <hexnum> }
+<escape>  ::= n | r | t | \ | " | '
+<utf8>    ::= <ascii> | <utf8enc>
+<ascii>   ::= '\20'..'\7e' except " or \
+<utf8enc> ::=
+  | '\c2'..'\df' <utf8cont>
+  | '\e0' '\a0'..'\bf' <utf8cont>
+  | '\ed' '\80'..'\9f' <utf8cont>
+  | '\e1'..'\ec' <utf8cont> <utf8cont>
+  | '\ee'..'\xef' <utf8cont> <utf8cont>
+  | '\f0' '\90'..'\bf' <utf8cont> <utf8cont>
+  | '\f4' '\80'..'\8f' <utf8cont> <utf8cont>
+  | '\f1'..'\f3' <utf8cont> <utf8cont> <utf8cont>
+<utf8cont> ::= '\80'..'\bf'
+```
+A `<char>` is a *Unicode scalar value* (i.e., a codepoint that is not a surrogate part).
+
+
+#### Syntactic Shorthands
+
+Analoguous to types, a few syntactic shorthands are supported that can be reduced to the basic value forms:
+
+```
+<consval> ::= ...
+  | blob <text>            := vec { N;* }  where N* are of bytes in the string, interpreted [as in the WebAssembly textual format](https://webassembly.github.io/spec/core/text/values.html#strings)
+
+<fieldval> ::= ...
+  | <name> = <annval>      :=  <hash(name)> = <annval>
+  | <annval>               :=  N = <annval>  where N is either 0 or previous + 1  (only in records)
+  | <nat>                  :=  <nat> = null   (only in variants)
+  | <name>                 :=  <name> = null  (only in variants)
+```
+
+
 ## Upgrading and Subtyping
 
 Interfaces are allowed to evolve over time in a manner that is *robust*, i.e., cannot break existing client code. To capture this notion precisely, a service of type `T` is *upgradable* to a version with another type `T'` if and only if `T'` is *structural subtype* of `T`, written `T' <: T`. This defines that `T'` is more *specialised* than `T`. (Note: A more specialised type is less general, i.e., denotes a smaller set of possible values, thus the direction of the subtype ordering, even though a subtype record can have *more* fields.)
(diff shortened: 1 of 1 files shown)


spec/Candid.md
```diff
@@ -552,6 +638,77 @@ That is, outbound message results can only be replaced with a subtype (more fiel
...
+Under normal subtyping rules, this wouldn't be allowed, because such record extension isn't usually compatible (sound) when `t` occurs in inbound position, such as with the `consume` function.
+There might be existing clients that are not aware of the new fields, and would fail to provide them.
+However, by restricting such fields to option types, and interpreting them as `null` when missing, such a mismatch is bridged.
+
+Such extensibility also extends to *higher-order* examples, where functions themselves become parameters:
+```
+type t = {x : nat};
+service : {
+  h1 : (f1 : () -> t) -> ();      // might call f() and expects a t
+  h2 : (f2 : t -> ()) -> ();      // might call f({x = 5})
```
chenyan-dfinity, line 677:
```suggestion
  h1 : (f1 : () -> t) -> ();      // might call f1() and expects a t
  h2 : (f2 : t -> ()) -> ();      // might call f2({x = 5})
```

spec/Candid.md
```diff
@@ -608,6 +765,23 @@ The premise means that the rule does not apply when the constituent type is itse
...
+---------------------------------
+opt <datatype> <: opt <datatype'>
+```
+*Note:* This rule is necessary in the presence of the unusual record and variant rules shown below. Without it, certain upgrades may generally be valid one step at a time, but not taken together, which could cause problems for clients catching up with multiple upgrades.
+For example, given a record type `record {666 : opt nat}` it is valid to remove the field `666` by the rule below and evolve the type to `record {}`.
+A later step might legally re-add a field of the same name but with a different type, producing, e.g.,`record {666 : opt text}`.
+A client having missed the intermediate step will have to upgrade directly from the original to the newest version of the type.
+If two option type do not match up, its value will be treated as `null`.
+
+In practice, users are strongly discouraged to ever remove a record field or ariant tag and later re-add it with a different meaning.
```
chenyan-dfinity, line 780:
```suggestion
In practice, users are strongly discouraged to ever remove a record field or a variant tag and later re-add it with a different meaning.
```

spec/Candid.md
```diff
@@ -623,7 +797,16 @@ record { <fieldtype>;* } <: record { <fieldtype'>;* }
...
 ```
 
-**NOTE**: There is a need for a mechanism to also remove fields (which means adding a field when a record appears as an argument). The precise mechanism is still work in progress.
+In order to be able to evolve and extend record types that also occur in inbound position (i.e., are used both as function results and function parameters), the subtype relation also supports *removing* fields from records, provided they are optional.
+```
+record { <fieldtype>;* } <: record { <fieldtype'>;* }
+----------------------------------------------------------------------------------------------
+record { <fieldtype>;* } <: record { <nat> : opt <datatype'>; <fieldtype'>;* }
+```
+*Note:* This rule is unusual form a regular subtyping perspective, but necessary in practice.
```
chenyan-dfinity, line 806:
```suggestion
*Note:* This rule is unusual from a regular subtyping perspective, but necessary in practice.
```

spec/Candid.md
```diff
@@ -639,6 +822,17 @@ variant { <fieldtype>;* } <: variant { <fieldtype'>;* }
 variant { <nat> : <datatype>; <fieldtype>;* } <: variant { <nat> : <datatype'>; <fieldtype'>;* }
 ```
 
+In order to be able to evolve and extend variant types that also occur in outbound position (i.e., are used both as function results and function parameters), the subtype relation also supports *adding* tags to variants, provided the variant itself is optional.
+```
+opt variant { <fieldtype>;* } <: opt variant { <fieldtype'>;* }
+----------------------------------------------------------------------------------------------
+opt variant { <nat> : opt <datatype>; <fieldtype>;* } <: opt variant { <fieldtype'>;* }
+```
+*Note:* This rule is unusual form a regular subtyping perspective, but it is the dual to the one for records.
```
chenyan-dfinity, line 831:
```suggestion
*Note:* This rule is unusual from a regular subtyping perspective, but it is the dual to the one for records.
```

spec/Candid.md
```diff
@@ -711,7 +905,33 @@ not (null <: <datatype>)
...
 ------------------------------------------
 <datatype> <: opt <datatype'> ~> \x.?(f x)
+
+not (<datatype> <: <datatype'>)
+---------------------------------
+opt <datatype> <: opt <datatype'>
+  ~> \x.join_opt (\y.
+       if (exists <datatype''>.
+          y : <datatype''> /\
+          <datatype''> <: <datatype> /\
```
chenyan-dfinity, line 915:
```suggestion
          <datatype''> <: <datatype'> /\
```

spec/Candid.md
```diff
@@ -711,7 +905,33 @@ not (null <: <datatype>)
...
 ```
+v : <datatype>  iff  M(v : <datatype>) is defined
+```
+The last rule *optimistically* tries to decode an option value when the types are not statically known to match.
+It succceeds if there would have been a valid type for the value (written `v : t`) that is a subtype of both types.
+The effect of this rule is that decoders do not need to perform a subtype check during deserialisation.
+Instead, they can simply try to deserialise, and if they encounter a mismatch abort, to the innermost option type, returning `null`.
+
+*Note:* The working assumption is that the type describing its incoming value is typically *principal*, i.e., the most precise type assignable to the value (in a sense that could be made precise).
+In that case, `<datatype''>` will equal `<datatype>` and deserialisation succeeds exactly if the types match.
```
chenyan-dfinity, line 930:
```suggestion
In that case, `<datatype''>` equals `<datatype>` and deserialization succeeds exactly if the types match.
```

chenyan-dfinity (commented):
LGTM. Just some typo.

spec/Candid.md
```diff
@@ -711,7 +905,33 @@ not (null <: <datatype>)
 <datatype> <: <datatype'> ~> f
 ------------------------------------------
 <datatype> <: opt <datatype'> ~> \x.?(f x)
+
+not (<datatype> <: <datatype'>)
+---------------------------------
+opt <datatype> <: opt <datatype'>
+  ~> \x.join_opt (\y.
```
nomeata, line 912: `join_opt` is a bit obscure. maybe write it out with `match … with`, for greater clarity?

spec/Candid.md
```diff
@@ -711,7 +905,33 @@ not (null <: <datatype>)
...
+          <datatype''> <: <datatype> /\
+          <datatype''> <: <datatype> ~> f)
+       then ?(f y)
+       else null)
+```
+where
 ```
+v : <datatype>  iff  M(v : <datatype>) is defined
+```
+The last rule *optimistically* tries to decode an option value when the types are not statically known to match.
```
nomeata, line 924: There is no “when not statically known to match” due to “Simple Deserialisation: Deserialisation does not require (sub)type checks other than the type-directed traversal of the value blob”

spec/Candid.md
```diff
@@ -711,7 +905,33 @@ not (null <: <datatype>)
...
+It succceeds if there would have been a valid type for the value (written `v : t`) that is a subtype of both types.
+The effect of this rule is that decoders do not need to perform a subtype check during deserialisation.
+Instead, they can simply try to deserialise, and if they encounter a mismatch abort, to the innermost option type, returning `null`.
+
+*Note:* The working assumption is that the type describing its incoming value is typically *principal*, i.e., the most precise type assignable to the value (in a sense that could be made precise).
```
nomeata, line 929: I don’t see right now where we make that assumption, or how it comes into play, but it’s early in the morning… Isn’t the purpose of the whole “if exists” to deal with non-most-precise types?

Anyways, the assumption isn’t even possible (not just costly to enforce): Consider:
```
(vec { opt variant { a } ; opt  variant { a }  }) : (vec ?(variant { a; b }))
```
Clearly the whole message has principal type, but when we reach the optional value `v = opt variant { a }` and want to apply this rule, the type of that value is `?(variant { a; b })` which is not principal, and there is no way to have it principal.

So maybe this Note can be removed? Or maybe I don’t see the point yet.

spec/Candid.md
```diff
@@ -740,8 +966,14 @@ variant { <fieldtype>;* } <: variant { <fieldtype'>;* } ~> f2
 ------------------------------------------------------------------------------------------------
 variant { <nat> : <datatype>; <fieldtype>;* } <: variant { <nat> : <datatype'>; <fieldtype'>;* }
   ~> \x.case x of <nat> y => <nat> (f1 y) | _ => f2 x
+
+opt variant { <fieldtype>;* } <: opt variant { <fieldtype'>;* } ~> f
+----------------------------------------------------------------------------------------------
+opt variant { <nat> : opt <datatype>; <fieldtype>;* } <: opt variant { <fieldtype'>;* }
+  ~> \x.join_opt (map_opt (\y.case y of <nat> z => null | _ => ?(f x)))
```
nomeata, line 973: This overlaps with the more general `opt` rule. It’s clear that it morally should have precendece, but still, is that a problem?

spec/Candid.md
rossberg: Done.

spec/Candid.md
rossberg: Wording was probably misleading, tweaked. The "principality" property only needs to hold for the whole type, not inductively for its parts. In your example (I suppose the second value ought to be `b`), `?(variant { a; b })` is the principal type -- and it's enough, since it ensures that both `a` and `b` will ultimately be checked against the respective supertype.

spec/Candid.md
```diff
@@ -639,6 +822,17 @@ variant { <fieldtype>;* } <: variant { <fieldtype'>;* }
 variant { <nat> : <datatype>; <fieldtype>;* } <: variant { <nat> : <datatype'>; <fieldtype'>;* }
 ```
 
+In order to be able to evolve and extend variant types that also occur in outbound position (i.e., are used both as function results and function parameters), the subtype relation also supports *adding* tags to variants, provided the variant itself is optional.
+```
+opt variant { <fieldtype>;* } <: opt variant { <fieldtype'>;* }
+----------------------------------------------------------------------------------------------
+opt variant { <nat> : opt <datatype>; <fieldtype>;* } <: opt variant { <fieldtype'>;* }
+```
```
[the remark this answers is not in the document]
rossberg: > Ah, but they have different elaboration of course… this makes your rules overlapping (I think you avoided that before…) A problem?

Yes, it would be better if this wasn't there. Same problem as above. For now, I added similar handwaving to pepper over the non-determinism.

> But I thought we’d agree we need different relations for “sensible evolution” and “decoding”?

I realised that doesn't actually work. You need to allow all of decoding in the evolution check, because cases like that might not be under your own control: some type you use may originate from another canister, that has performed multiple evolutionary steps that are transitively non-coherent, but you have missed the intermediate step that would avoid the "ugly" transitive case when you upgrade yourself.

Hence the informal language in L785+ instead, about discouraging users to create that case and implementations to warn about it.

IOW, we need the completeness property after all, even if some cases hopefully should never occur in practice if everybody follows good style.

rossberg (commented):
@nomeata, PTAL.

spec/Candid.md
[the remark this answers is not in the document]
nomeata: I don't follow the example. Even if you miss a step, and even if there there is incoherence, the “interface evolution” relation would still be transitive, woudn’t it?  But maybe there is some complicate higher-order case where some canister is in the unfortunate situation where it is force to upgrade it’s interface in an undesirable way.

spec/Candid.md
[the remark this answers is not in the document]
nomeata: Yes, we can clean up separtely.

I don’t think it’d be a four-place relation: Didn’t we determine that the input data should be considered untyped (and the type description just be an encoding help)? So I’d expect `v <: v' : t` or maybe more suggestively `v ~> v':t`.

spec/Candid.md
nomeata: Hmm, I think I am beginning to understand why you want this principality here.

This might all be clearer if the rules for decoding would just not look at the input type at all, and only at the values.

spec/Candid.md
```diff
@@ -552,6 +638,82 @@ That is, outbound message results can only be replaced with a subtype (more fiel
...
+
+* Simple Deserialisation: Deserialisation does not require (sub)type checks other than the type-directed traversal of the value blob.
+
+* Type Erasure: Deserialised values do not require carrying dynamic type information on the language side. In particular, service and function references can be represented without type information.
+
+* No covert channels: Serialisation never includes any fields in the value that the sender is not aware of. Specifically, when passing on a value to a third party that the sender previously received itself, then that will only contain fields that the sender intends to send out per its type.
+
+However, something has to give, so one seemingly desirable property that is not maintained is *transitive coherence*, i.e., when deserialising and reserialising a value and then deserialising that again at a different supertype, it might yield another result than it would if the original value was deserialised at the later supertype.
```
nomeata, line 713: (minor tweak)
```suggestion
However, something has to give, so one seemingly desirable property that is not maintained is *transitive coherence*, i.e., given a value serialized at some type, deserialized and serialized at a supertype, and then again deserialized at a supertype of the supertype may yield a diferent result than deserialised directly at the later supertype.
```

spec/Candid.md
```diff
@@ -608,6 +770,23 @@ The premise means that the rule does not apply when the constituent type is itse
...
+---------------------------------
+opt <datatype> <: opt <datatype'>
+```
+*Note:* This rule is necessary in the presence of the unusual record and variant rules shown below. Without it, certain upgrades may generally be valid one step at a time, but not taken together, which could cause problems for clients catching up with multiple upgrades.
+For example, given a record type `record {666 : opt nat}` it is valid to remove the field `666` by the rule below and evolve the type to `record {}`.
+A later step might legally re-add a field of the same name but with a different type, producing, e.g.,`record {666 : opt text}`.
+A client having missed the intermediate step will have to upgrade directly from the original to the newest version of the type.
+If two option type do not match up, its value will be treated as `null`.
+
+In practice, users are strongly discouraged to ever remove a record field or a variant tag and later re-add it with a different meaning.
```
nomeata, line 785: Wouldn’t this work:
```suggestion
In practice, users are strongly discouraged to ever remove an optional record field or a variant tag and later re-add it with a different meaning. Instead of removing an optional record field, it should be replaced with `opt empty`, to prevent re-usd of that field.
```
Maybe we can introduce an alias for `opt empty` to make this less obscure?

This is also how I envisioned the restricted “sensible evolution” relation to look like: It would not allow removing optional fields, only “disabling” them in this way.

This might require a form of polarities in the rules for “can evolve to”. So why do I think this might work (when it didn’t work for normal subtyping)? Because in the higher order case, where two services evolve independently, the composition needs to work with regard to the unrestricted subtyping relation, which is larger.

spec/Candid.md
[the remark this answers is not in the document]
rossberg: Our previous conclusion was (unless I misremember something) that the transitive rule (opt t <: opt t' with incompatible t, t') should not be allowed in the upgrade check, because that check always deals with a single evolutionary step where you simply shouldn't be doing that. And that we hence should define two relations, one with and the other without this rule (or something like that).

But the assumption that the check always deals with a single evolutionary step is bogus if your interface has to adapt to (the latest version of) somebody else's interface types, whose evolution might have a different pace, and who makes bad transitive evolutionary choices outside your control. In that case you may need the "bad" rule on options even during upgrade checks.

spec/Candid.md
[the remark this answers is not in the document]
rossberg: I suppose we could phrase it like that, but that would be a bit of a disconnect with the actual serialisation format, which _does_ include a type -- it just may not be "principal".

nomeata (approved):
Maybe not in this PR, but reading this I get a strong urge to re-write the spec in terms of

 * A relation `T <:: T'` of “good” updates, that is more restrictive that what we allow (in particular, if your service defines an optional field and later removes it, this is marked using `opt null`).
 * A set `V` of untyped values (essentially the AST of the encoded values)
 * A relation `V : T` to describe the values of the types (likely without built-in subtyping, i.e. `{foo = true} /: record {}`, as it is not needed here)
 * A type-indexed decoding function `T ~> (V -> T ∪ {fail})`, which by construction ignores the type annotation on the input value (`V` is the set of Candid values), and models decoding failure explicitly (which is needed to describe the backtracking behavior).

And then we can describe (and prove!) properties that we want to hold for these things, e.g.

 * `∀ t ∀ v ∈ V. t ~> f ⟹ (f v) = fail ∨ (f v) : t`
 * `∀ t ∀ v : t. (f v) = v` (round-tripping is possible)
 * `<::` is transitive
 * `∀ t1 t2 ∀ v ∈ V. t1 <:: t2 ⟹  t1 ~> f1 ⟹ t2 ~> f2 ⟹ (f2 v) ≠ fail ⟹(f1 v) ≠ fail`
 * (Something similar for the polarity switched version of `<::`, if it needs polarities)

But maybe not in this PR.

spec/Candid.md
rossberg: Hm, I'm not convinced that wouldn't run afoul all the same problems in the higher-order or third-party cases. But happy to explore, later. :)

spec/Candid.md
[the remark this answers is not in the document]
nomeata: > Our previous conclusion was (unless I misremember something) that the transitive rule (opt t <: opt t' with incompatible t, t') should not be allowed in the upgrade check, 

yes… but

> because that check always deals with a single evolutionary step where you simply shouldn't be doing that

wasn’t the reason, I think. I thought the reason was that we only had problems when you need the bad rules due to transitivity because you composed two serivces, and _each_ was doing one (or more) sensible steps. But the transtive relation was only relevant for decoding, not an actual service evolution.

That’s why I hope we can have a  “can sensibly evolve” relation (probably with polarities), that’s transitive, and a separate relation “can be decoded at” relation that has the unwanted relations.

But maybe I am too optimistic, and there will always be corner cases where a service needs to change its interface is a “bad” way.

spec/Candid.md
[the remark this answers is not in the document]
nomeata: But that disconnect would be intentional: Didn't we just last week realize that it was misguided to take that type as more than just a way to help understand the _binary_ structure of the value?

PR #112, opened 2020-10-07
Bump spec version

nomeata:
the changes in https://github.com/dfinity/candid/pull/110 are definitely
more than editorial, so even if we don’t bump the Magic Bytes, I think
we should definitely bump the version of the spec.

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -1,6 +1,6 @@
 # Candid Specification
 
-Version: 0.1.1
+Version: 0.2
 
 Date: Aug 25, 2020
 
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -1,6 +1,6 @@
 # Candid Specification
 
-Version: 0.2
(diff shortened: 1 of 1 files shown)


spec/Candid.md
```diff
@@ -1,6 +1,6 @@
 # Candid Specification
 
-Version: 0.1.1
+Version: 0.2
```
chenyan-dfinity, line 3: I think the change is an extension of the existing rules. If we follow semantic versioning, it should be `0.1.2`?

nomeata: Not really: It’s an extension if a service starts relying on these rules, but not if a client starts to rely on them (and starts omitting optional fields). But I don't care as long as we have a new version, so that we can say things like “has motoko already updated the decoder to Candid 0.1.2”. We can use animal names instead, FWIW.

```suggestion
Version: 0.1.2
```

chenyan-dfinity (approved):

PR #113, opened 2020-10-07
The “reverse variant extension rule” is redundant

nomeata:
and can be replaced by a note


rossberg (commented):
Thanks. Also remove the corresponding elaboration rule?

rossberg (approved):

PR #114, opened 2020-10-07
Spec: Opt decoding also from non-opt values

nomeata:
The problem that we had before this PR was the following:

* A sends values at type `record { foo : opt bool }`
  B recvs values at type `record { }`

  All izz well, as `record { foo : opt bool } <: record { }`

* Now `A` uses the rule
  ```
  not (null <: <datatype>)
  <datatype> <: <datatype'>
  -----------------------------
  <datatype> <: opt <datatype'>
  ```
  A sends values at type `record { foo : bool }` now

* Independently, `B` uses new rule about optional fields
  ```
  <datatype> <: <datatype'>
  -----------------------------
  <datatype> <: opt <datatype'>
  <nat> not in <fieldtype>;*
  record { <fieldtype>;* } <: record { <fieldtype'>;* }
  ------------------------------------------------------------------------------
  record { <fieldtype>;* } <: record { <nat> : opt <datatype'>; <fieldtype'>;* }
  ```
  B recvs values at type `record { foo : opt nat }`

Either of these steps on their own don’t break anything, because we
allow
 * `record { foo : bool } <: record { }`
 * `record { foo : opt bool } <: record { foo : opt nat }`

But both steps together are not allowed:

 * `record { foo : bool } <: record { foo : opt nat }`

The remidy seems to be to make the rule “when decoding an optional
field, be liberal in what you accept” even more liberal, and also allow
non-optional values.

This PR implements that.

The final rule
```
<datatype> <: opt <datatype'>
```
looks very non-informative, but it seems to be the honest rule.

A direct generalization of the previous rule might have been
```
<datatype> <: opt <datatype'>
```
which is just odd.

An alternative might be to enumerate the cases, and have two rules
(which also matches the elaborated rules more closely):
```

not (null <: <datatype>)
```

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -781,7 +781,7 @@ A later step might legally re-add a field of the same name but with a different 
 A client having missed some of the  intermediate steps will have to upgrade directly to the newest version of the type.
 If the type cannot be decoded, its value will be treated as `null`.
 
-In practice, users are strongly discouraged to ever remove an record field or a variant tag and later re-add it with a different meaning. Instead of removing an optional record field, it should be replaced with `opt empty`, to prevent re-use of that field.
+In practice, users are strongly discouraged to ever remove a record field or a variant tag and later re-add it with a different meaning. Instead of removing an optional record field, it should be replaced with `opt empty`, to prevent re-use of that field.
 However, there is no general way for the type system to prevent this, since it cannot know the history of a type definition.
 Consequently, the rule above is needed for technical more than for practical reasons.
 Implementations of static upgrade checking are encouraged to warn if this rule is used.
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -781,7 +781,7 @@ not (<datatype> <: <datatype'>)
 ---------------------------------
 opt <datatype> <: opt <datatype'>
 ```
-*Note:* This rule is necessary in the presence of the unusual record and variant rules shown below. Without it, certain upgrades may generally be valid one step at a time, but not taken together, which could cause problems for clients catching up with multiple upgrades.
+*Note:* These rules are necessary in the presence of the unusual record and variant rules shown below. Without them, certain upgrades may generally be valid one step at a time, but not taken together, which could cause problems for clients catching up with multiple upgrades.
 For example, given a record type `record {666 : opt nat}` it is valid to remove the field `666` by the rule below and evolve the type to `record { 666 : nat }` and then to `record {}`.
 A later step might legally re-add a field of the same name but with a different type, producing, e.g.,`record {666 : opt text}`.
 A client having missed some of the  intermediate steps will have to upgrade directly to the newest version of the type.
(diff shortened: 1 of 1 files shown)


spec/Candid.md
```diff
@@ -770,19 +770,18 @@ The premise means that the rule does not apply when the constituent type is itse
...
 A later step might legally re-add a field of the same name but with a different type, producing, e.g.,`record {666 : opt text}`.
-A client having missed the intermediate step will have to upgrade directly from the original to the newest version of the type.
-If two option type do not match up, its value will be treated as `null`.
+A client having missed some of the  intermediate steps will have to upgrade directly to the newest version of the type.
+If the type cannot be decoded, its value will be treated as `null`.
 
-In practice, users are strongly discouraged to ever remove an optional record field or a variant tag and later re-add it with a different meaning. Instead of removing an optional record field, it should be replaced with `opt empty`, to prevent re-use of that field.
+In practice, users are strongly discouraged to ever remove an record field or a variant tag and later re-add it with a different meaning. Instead of removing an optional record field, it should be replaced with `opt empty`, to prevent re-use of that field.
```
rossberg, line 784:
```suggestion
In practice, users are strongly discouraged to ever remove a record field or a variant tag and later re-add it with a different meaning. Instead of removing an optional record field, it should be replaced with `opt empty`, to prevent re-use of that field.
```

rossberg (commented):
I'd vote for splitting the rule as you suggest in the description, both for clarity and to match up with the elaboration.

spec/Candid.md
```diff
@@ -772,16 +772,22 @@ Q: The negated nature of this premise isn't really compatible with parametric po
...
 ---------------------------------
-<datatype> <: opt <datatype'>
+opt <datatype> <: opt <datatype'>
+
+not (null <: <datatype>)
+not (<datatype> <: <datatype'>)
+---------------------------------
+opt <datatype> <: opt <datatype'>
 ```
 *Note:* This rule is necessary in the presence of the unusual record and variant rules shown below. Without it, certain upgrades may generally be valid one step at a time, but not taken together, which could cause problems for clients catching up with multiple upgrades.
```
rossberg, line 784:
```suggestion
*Note:* These rules are necessary in the presence of the unusual record and variant rules shown below. Without them, certain upgrades may generally be valid one step at a time, but not taken together, which could cause problems for clients catching up with multiple upgrades.
```

spec/Candid.md
```diff
@@ -772,16 +772,22 @@ Q: The negated nature of this premise isn't really compatible with parametric po
 
 Finally, in order to maintain *transitivity* of subtyping, an unusual rule allows, in fact, *any* type to be regarded as a subtype of an option.
```
rossberg, line 773:
```suggestion
Finally, in order to maintain *transitivity* of subtyping, two unusual rules allow, in fact, *any* type to be regarded as a subtype of an option.
```

rossberg (approved):

PR #116, opened 2020-10-07
improve pretty printing for values

chenyan-dfinity:
* Underscore for numerals
* Blob shorthand for `vec nat8`
* Show first 20 elements in `vec t`

### DIFF
--- a/rust/candid/src/parser/value.rs
+++ b/rust/candid/src/parser/value.rs
@@ -360,11 +360,11 @@ pub mod pretty {
             Principal(ref id) => RcDoc::as_string(format!("principal \"{}\"", id)),
             Opt(v) => kwd("opt").append(pp_value(v)),
             Vec(vs) => {
-                if let Some(IDLValue::Nat8(_)) = vs.first() {
+                if let Some(Nat8(_)) = vs.first() {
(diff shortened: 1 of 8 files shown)


rust/candid/src/parser/value.rs
```diff
@@ -334,6 +334,8 @@ pub mod pretty {
         enclose_space("{", fs, "}")
     }
 
+    const MAX_VEC_ELEMENTS: usize = 20;
```
nomeata, line 337: This means that you cannot really use the pretty-printed format in scripts any more, as you lose data. Is that a good idea? Does the library give applications an option whether to shorten values this way?

Also, when shortening vec, why ont `blob` similarly (we have messages with pretty large blobs…)

chenyan-dfinity: It is shortening both `vec` and `blob`. Pretty printing is mostly for human consumption. Large vector also makes the layout algorithm slow for pretty printer. I should have a non-pretty printed version for scripting.

chenyan-dfinity: Done in https://github.com/dfinity/candid/pull/118.

PR #124, opened 2020-10-27
[spec] facilitate message authentication with candid serialization

chenyan-dfinity:
As discussed with @timohanke, we would like to explore using the value part (`M`) of Candid message for computing/verifying user-defined signature scheme.

This requires `M` to be a one-to-one mapping between value and bytes given a fixed type. Besides the non-determinism of `leb128`, this is already the case. We would like to add a statement in the spec to make sure it doesn't change in the future.

For the authentication to work, we assume the following:
* Both sides agree on the method type (not sure how realistic this assumption is)
* Both sides use the same/deterministic `leb128` encoder
* Canister code has access to the raw serialized bytes
* Use domain separator to prevent replaying the message in other context

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -1127,6 +1127,11 @@ Note:
 `M` maps an Candid value to a byte sequence representing that value. The definition is indexed by type.
 We assume that the fields in a record value are sorted by increasing id.
 
+Given a fixed type, we guarantee that there is a one-to-one mapping between the Candid value and the byte sequence.
+This property ensures that we can use `M` for message authentication when both the sender and receiver agree on the same type. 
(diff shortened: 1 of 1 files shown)


rossberg:
Note that we initially required a deterministic representation (and shortest LEB) but removed that requirement (can't find the PR right now). The reason we removed it is that it seemed impractical once you enrich the format. It already is the case that the type part of a serialised value can be represented and ordered arbitrarily, and there is no affordable way to avoid that.

I don't understand the use case well enough from the description, but it seems to require piercing the serialisation abstraction itself, by looking at the M-part in isolation. Tbh, that seems rather fishy and brittle, and I would not recommend it. Moreover, IIUC, it requires locking in a fixed method type and prevent any form of evolution on it, which is in conflict with our design goals.

PR #125, opened 2020-10-27
remove multi-line string in test suites

chenyan-dfinity:


nomeata:
Well, is that legal candid or is it not? :)

PR #128, opened 2020-11-02
Spec: Refactor description of deserialization

nomeata:
This is a significant rewirte of the section in the spec describing
deserialization. Some points worth noting:

 * This replaces the elaboration relation with one that takes a value
   and a type as input, and returns a value. This makes it clearer that
   there is no “input _type_” of deeper relevance.

 * One great benefit: No more long prose about how the rules assume
   these types to be principal, even if they aren’t.

   This clarifies questions like [this
   one](https://github.com/dfinity/candid/pull/126#discussion_r515248927)

 * Also, the existing rules about “elaborating function values” were
   kinda bogus, as function values are just accepted as they are. This
   is much easier now, also good.

 * Unfortuantely for this application, our textual representation has
   overloading. Instead of defining yet another abstract value algebra,
   I did a bit of hand-waving, and defined an “overloading-free
   fragment” of the textual format for this section.

 * A fair number of rules becomes simpler. Promising!

 * I was able to phrase some interesting properties more formally. Also
   promising! Didn’t actually prove them, though, although we should.

 * Still unclear to me what we mean with “subtyping is complete”, see
   the section there.

 * Decoding is still an inductive relation, so not fully algorithmic.
   This may be an issue with a stright-forward formalization; not all
   theorem provers like negative occurrences of inductive relations in
   their rules.

 * Syntax of the relations up for discussion.

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -77,14 +77,17 @@ This is a summary of the grammar proposed:
 <datatype>  ::= <id> | <primtype> | <constype> | <reftype>
 
 <primtype>  ::=
-  | nat | nat8 | nat16 | nat32 | nat64
-  | int | int8 | int16 | int32 | int64
-  | float32 | float64
+  | <numtype>
   | bool
   | text
   | null
   | reserved
   | empty
+
+<numtype>  ::=
+  | nat | nat8 | nat16 | nat32 | nat64
+  | int | int8 | int16 | int32 | int64
+  | float32 | float64
 
 <constype>  ::=
   | opt <datatype>
@@ -230,13 +233,17 @@ The content of message arguments and results is *data*. Three basic forms of *da
 
 *Primitive types* describe the possible forms of primitive data.
 
+The primitive types that describe numbers are separted out in the grammar:
+```
+<primtype> ::= <numtype> | ...
+```
 
 #### Natural Numbers
 
 The type `nat` describes a natural number (unsigned integer) of unlimited range. There are also variants limited to 8, 16, 32, or 64 bit value range with fixed-size representations.
 
 ```
-<primtype> ::= nat | nat8 | nat16 | nat32 | nat64 | ...
+<numtype> ::= nat | nat8 | nat16 | nat32 | nat64 | ...
 ```
 **Note:** Values of type `nat` have variable length representations in the  binary serialisation format, and hence take up space proportional to (the logarithm of) their value. As long as typical values are small, they may hence be more space-efficient than the fixed size types.
 
@@ -245,7 +252,7 @@ The type `nat` describes a natural number (unsigned integer) of unlimited range.
 The type `int` describes an integer number (signed) of unlimited range. There are also variants limited to 8, 16, 32, or 64 bit value range with fixed-size representations.
 
 ```
-<primtype> ::= ... | int | int8 | int16 | int32 | int64 | ...
+<numtype> ::= ... | int | int8 | int16 | int32 | int64 | ...
 ```
 **Note:** Values of type `nat` have variable length representations in the binary serialisation format, and hence take up space proportional to (the logarithm of) their value. As long as typical values are small, they may hence be more space-efficient than the fixed size types.
 
@@ -254,7 +261,7 @@ The type `int` describes an integer number (signed) of unlimited range. There ar
 Floating-point values are represented in IEEE 754 binary format and are supported in single precision (32 bit) and double precision (64 bit).
 
 ```
-<primtype> ::= ... | float32 | float64 | ...
+<numtype> ::= ... | float32 | float64
 ```
 
 #### Boolean
@@ -865,150 +872,147 @@ service { <methtype>;* } <: service { <methtype'>;* }
 service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>; <methtype'>;* }
 ```
 
-### Decoding
+### Coercing
 
-This subtyping is implemented when decoding a Candid value. We provide a ternary relation `V :? T ~> V'` to describe when a value `V` can be decoded at type `T` to a value `V'` of type `T`.
+This subtyping is implemented during the deserialization of Candid at an expected type. As as described in [Section Deserialisation](#deserialization), the binary value is conceptually first _decoded_ into an abstract value, and then _coerced_ to the expected type.
 
-The binary relation `V !: T` describes that a value cannot be decoded at type `T`, and is defined as
-```
-not (∃ v'. <v> :? <t> ~> v')
-------
-<v> !: <t >
-```
+This section describes the covercion, as ternary relation `V ~> V' : T` to describe when a value `V` can be coerced to a value `V'` of type `T`. The fields `V` and `T` can be understood as inputs and `V'` as the output of this relation.
 
-Here `V` models untyped values, which form an abstract data model of both the message in transit (`V`), as well as the the message. In the following, we re-use the syntax of the textual representation, with the following changes to make it free of overloading:
+Here `V` models untyped values, which form an abstract data model of both the message in transit (`V`), as well as the result of the coercion (`V'`). In the following, we re-use the syntax of the textual representation, with the following changes to make it free of overloading:
 
  * Number literals (`<primval>`) must be immediately enclosed with an `<annval>` that clarifies the precise number type.
- * The value of type `reserved` is expressed as `(null : reserved)`.
+ * The canonical value of type `reserved` is expressed as `(null : reserved)`.
  * No other use of `<annval>`.
 
 #### Primitive Types
 
-Values of primitive types decode successfully at their own type:
+Values of primitive types coerce successfully at their own type:
 ```
-<t> ∈ {nat, int, nat8, nat16, nat32, nat64, int8, int16, int32, int64, float32, float64 }
--------------------------------------
-(<x> : <t>) :? <t> ~> (<x> : <t>)
+--------------------------------------------------
+(<x> : <numtype>) ~> (<x> : <numtype>) : <numtype>
 
---------------------
-true :? bool ~> true
+-------------------
+true ~> true : bool
 
 ----------------------
-false :? bool ~> false
+false ~> false : bool
 
-------------------------
-<text> :? text ~> <text>
+-----------------------
+<text> ~> <text> : text
 
 ------------------------
 null :? null ~> null
 ```
 
-Values of type `nat` decode at type `int`:
+Values of type `nat` coerce successfully at type `int`:
 ```
----------------------------------
-(<x> : nat) :? int ~> (<x> : int)
-```
-
-Any value decodes at type `reserved`, producing the canonical value of type `reserved`:
-
-```
-------------------------------
-<v> :? reserved ~> null : reserved
+--------------------------------
+(<x> : nat) ~> (<x> : int) : int
 ```
 
-NB: No rule needed for type `empty`, because there are no values of that type. By construction, `<v> !: empty`.
+Any value coerces at type `reserved`, producing the canonical value of type `reserved`:
+
+```
+-----------------------------------
+<v> ~> (null : reserved) : reserved
+```
+
+NB: No rule is needed for type `empty`, because there are no values of that type. By construction, `<v> ~> _ : empty` will not hold.
 
 #### Vectors
 
-Only vectors decode at vector types, and only if all elements decode successfully.
+Only vectors coerce at vector types, and only if all elements coerce successfully.
 
 ```
-<v> :? <t> ~> <v'>
-------------------------------------------
-vec { <v>;* } :? vec <t> ~> vec { <v'>;* }
+<v>  ~> <v'> : <t>
+-----------------------------------------
+vec { <v>;* } ~> vec { <v'>;* } : vec <t>
 ```
 
 #### Options
 
-The null value decodes at any option type:
+The null value coerce at any option type:
 ```
------------------------
-null :? opt <t> ~> null
-```
-
-An optional value decodes at an option type, if the value decodes at the constituent type:
-```
-<v> :? <t> ~> <v'>
------------------------
-opt <v> :? opt <t> ~> opt <v'>
+----------------------
+null ~> null : opt <t>
 ```
 
-If an optional value _fails_ to decode at an optional type, the result is `null`, not failure:
+An optional value coerces at an option type, if the constituent value coerces at the constituent type:
 ```
-<v> !: <t>
------------------------
-opt <v> :? opt <t> ~> null
+<v> ~> <v'> : <t>
+-----------------------------
+opt <v> ~> opt <v'> : opt <t>
 ```
 
-Decoding a non-null, non-optional and non-reserved value at an option type treats it as an optional value:
+If an optional value _fails_ to coerce at an optional type, the result is `null`, not failure:
+```
+not (<v> ~> _ : <t>)
+-------------------------
+opt <v> ~> null : opt <t>
+```
+
+Coercing a non-null, non-optional and non-reserved value at an option type treats it as an optional value:
 ```
 <v> ≠ null
 <v> ≠ (null : reserved)
 <v> ≠ opt _
-opt <v> :? opt <t> ~> <v'>
------------------------
-<v> :? opt <t> ~> <v'>
+opt <v> ~> <v'> : opt <t>
+-------------------------
+<v> ~> <v'> : opt <t>
 ```
 
 
 #### Records
 
-Only records decode at record type. Missing fields of option type turn into `null`.
+Only records coerce at record type. Missing fields of option or reserved type turn into `null`.
 
 In the following rule, the `<nat1>*` field names are those present in both the value and the type, the `<nat2>*` field names only those in the value, and `<nat3>*` are those only in the type.
 ```
-<v1> :? <t1> ~> <v1'>
---------------------------------------------------------------------------------------------------------------------------------------------
-record { <nat1> = <v1>;* <nat2> = <v2>;* } :? record {  <nat1> = <t1>;* <nat3> = opt <t2>;* } ~> record { <nat1> = <v1'>;* <nat3> = null;* }
+<v1> ~> <v1'> : <t1>
+<t3> = opt _ ∨ <t3> = reserved
+---------------------------------------------------------------------------------------------------------------------------------------
+record { <nat1> = <v1>;* <nat2> = <v2>;* } ~> record { <nat1> = <v1'>;* <nat3> = null;* } : record {  <nat1> = <t1>;* <nat3> = <t3>;* }
 ```
 
 
 #### Variants
 
+Only a variant value with an expected tag coerces at variant type.
+
 ```
-<v> :? <t> ~> <v'>
+<v> ~> <v'> : <t>
 ----------------------------------------------------------------------------------
-variant { <nat> = <v> } :? variant { <nat> = <t>; _;* } ~> variant { <nat> = <v'> }
+variant { <nat> = <v> } ~> variant { <nat> = <v'> } : variant { <nat> = <t>; _;* }
 ```
 
 
 #### References
 
-Function and services references decode unconditionally
+Function and services references coerce unconditionally
 
 ```
--------------------------------------------------------
-func <text>.<id> :? func <functype> ~> func <text>.<id>
-```
-
-```
--------------------------------------------------------
-service <text> :? service <actortype> ~> service <text>
+------------------------------------------------------
+func <text>.<id> ~> func <text>.<id> : func <functype>
 ```
 
 ```
--------------------------------------------------
-principal <text> :? principal ~> principal <text>
+------------------------------------------------------
+service <text> ~> service <text> : service <actortype>
+```
+
+```
+------------------------------------------------
+principal <text> ~> principal <text> : principal
 ```
 
 #### Tuple types
 
-Whole argument and result sequences are decoded with the same rules as tuple-like records. In particular, extra arguments are ignored, and optional parameters read as as `null` if the argument is missing or fails to decode:
+Whole argument and result sequences are coerced with the same rules as tuple-like records. In particular, extra arguments are ignored, and optional parameters read as as `null` if the argument is missing or fails to coerce:
 
 ```
-record {<v>;*} :? record {<t>;*} ~> record {<v'>,*}
----------------------------------------------------
-(<v>,*) :? (<t>,*) ~> (<v'>,*)
+record {<v>;*} ~> record {<v'>,*} : record {<t>;*}
+--------------------------------------------------
+(<v>,*) ~> (<v'>,*) : (<t>,*)
 ```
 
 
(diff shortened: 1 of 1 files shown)


nomeata:
@rossberg @chenyan-dfinity , what do you think of this?

As an implementor, I think I’d find this presentation more helpful. And if I were to prove soundess etc. formally, then too.

nomeata:
If the diff is too confusing to read, then head to https://github.com/dfinity/candid/blob/joachim/rewrite/spec/Candid.md#decoding

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+This subtyping is implemented when decoding a Candid value. We provide a ternary relation `V :? T ~> V'` to describe when a value `V` can be decoded at type `T` to a value `V'` of type `T`.
 
+The binary relation `V !: T` describes that a value cannot be decoded at type `T`, and is defined as
+```
+not (∃ v'. <v> :? <t> ~> v')
+------
+<v> !: <t >
+```
+
+Here `V` models an untyped values, and form an abstract data model of both the message in transit (`V`), as well as the the message . In the following, we re-use the syntax of the textual representation, with the following changes to make it free of overloading:
```
rossberg, line 878:
```suggestion
Here `V` models untyped values, which form an abstract data model of both the message in transit (`V`), as well as the the message. In the following, we re-use the syntax of the textual representation, with the following changes to make it free of overloading:
```

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+
+ * Number literals (`<primval>`) must be immediately enclosed with an `<annval>` that clarifies the precise number type.
+ * The value of type `reserved` is expressed as `(null : reserved)`.
+ * No other use of `<annval>`.
 
 #### Primitive Types
 
+Values of primitive types decode successfully at their own type:
 ```
+<t> ∈ {nat, int, nat8, nat16, nat32, nat64, int8, int16, int32, int64, float32, float64 }
```
rossberg, line 888: Maybe introduce a production `<numtype>` in the grammar?

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+<v> :? <t> ~> <v'>
+----------------------------------------------------------------------------------
+variant { <nat> = <v> } :? variant { <nat> = <t>; _;* } ~> variant { <nat> = <v'> }
 ```
 
 
-#### Variants
+#### References
+
+Function and services references decode unconditoinally
```
rossberg, line 986:
```suggestion
Function and services references decode unconditionally
```

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
 
-------------------------------------------------
-record { <fieldtype'>;* } <: record { } ~> \x.{}
 
-<datatype> <: <datatype'> ~> f1
-record { <fieldtype>;* } <: record { <fieldtype'>;* } ~> f2
-----------------------------------------------------------------------------------------------
-record { <nat> : <datatype>; <fieldtype>;* } <: record { <nat> : <datatype'>; <fieldtype'>;* }
-  ~> \x.{f2 x with <nat> = f1 x.<nat>}
+#### Variants
```
rossberg, line 975: No prose description here?

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
-  ~> \x.case x of <nat> y => <nat> (f1 y) | _ => f2 x
+```
+-------------------------------------------------
+principal <text> :? principal ~> principal <text>
 ```
 
+#### Tuple types
 
-#### Functions
+Whole argument and result sequences are decoded with the same rules are tuple-like records. In particular, extra arguments are ignored, and optional parameters read as as `null` if the argument is missing or fails to decode:
```
rossberg, line 1005:
```suggestion
Whole argument and result sequences are decoded with the same rules as tuple-like records. In particular, extra arguments are ignored, and optional parameters read as as `null` if the argument is missing or fails to decode:
```

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
 service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>; <methtype'>;* }
 ```
 
-### Elaboration
+### Decoding
 
-To define the actual coercion function used during deserialisation of a value, we extend the subtyping relation to a ternary *elaboration* relation `T <: T' ~> f`, where `f` is a suitable coercion function of type `T -> T'`.
+This subtyping is implemented when decoding a Candid value. We provide a ternary relation `V :? T ~> V'` to describe when a value `V` can be decoded at type `T` to a value `V'` of type `T`.
 
+The binary relation `V !: T` describes that a value cannot be decoded at type `T`, and is defined as
```
rossberg, line 871: Why bother with introducing this notation? It's only used in a single place.

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
 service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>; <methtype'>;* }
 ```
 
-### Elaboration
+### Decoding
 
-To define the actual coercion function used during deserialisation of a value, we extend the subtyping relation to a ternary *elaboration* relation `T <: T' ~> f`, where `f` is a suitable coercion function of type `T -> T'`.
+This subtyping is implemented when decoding a Candid value. We provide a ternary relation `V :? T ~> V'` to describe when a value `V` can be decoded at type `T` to a value `V'` of type `T`.
```
rossberg, line 869: Nit: The `:?` reads as if this was partial somehow. Why not plain `:`? You could even write it as `v ~> v' : T`, then the colon is on the right side.

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
-  ~> \x.{f1 x; <name> = f2 x.<name>}
-```
+* Correctness and completeness of decoding:
+  ```
+  (v : T) ⟺ (∃ v'. v' :? T ~> v)
+  ```
 
+* Roundtripping:
+  ```
+  (v : T) ⟺ v :? T ~> v
```
rossberg, line 1025: Hm, is that roundtripping? Is this assuming that serialisation is a no-op in this abstract model? Maybe this needs to be stated somewhere.

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
 service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>; <methtype'>;* }
 ```
 
-### Elaboration
+### Decoding
```
rossberg, line 867: The doc speaks of deserialisation in other places. Make consistent?

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+  ```
+  s1 ~> s2 ≔ s2 <: s1
+  t1 <: t2 ≔ (∀ v. t1: v :? t2 ~> _)
+  s1 in t1 <: s2 in t2 ≔ (to be done)
+  s1 <:h s2 ≔ (host-language dependent)
+  ```
+
+* Transitivity of subtyping:
+  ```
+  T <: T2, T2 <: T3 ⇒ T <: T3
```
rossberg, line 1049:
```suggestion
  T1 <: T2, T2 <: T3 ⇒ T1 <: T3
```

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+  ```
+
+* Transitive coherence does not hold:
+  ```
+  T <: T2, T2 <: T3
+  v1 : T
```
rossberg, line 1055:
```suggestion
  T1 <: T2, T2 <: T3
  v1 : T1
```

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+
+Other design goals are not satisfied (or hard to express formally?):
+
+* Completeness of subtyping:
+
+  A rigorous formulation of completeness could be that `<:` is the largest transitive and sound relation (in the above sense), i.e. that
+  ```
+  (∀ v. v :? T ~> _ ⇒ v :? T' ~> _) ⇒ T <: T'
+  ```
+  This does not hold as state, because of counter examples involving the empty type. For example we do not have `opt empty <: null`, or `Empty <: t` where `type Empty = rec { Empty }`
```
rossberg, line 1071: I would expect the following formulation:
```
(∀ v, v'. v : T /\ v :? T' ~> v') ⇒ T <: T'
```
Is there anything wrong with that? And should it not be a requirement?

rossberg (commented):
Good stuff, thanks!

spec/Candid.md
```diff
@@ -864,142 +864,213 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
 
 #### Records
 
+Only records decode at record type. Missing fields of option type turn into `null`.
+
+In the following rule, the `<nat1>*` field names are those present in both the value and the type, the `<nat2>*` field names only those in the value, and `<nat3>*` are those only in the type.
+```
+<v1> :? <t1> ~> <v1'>
+--------------------------------------------------------------------------------------------------------------------------------------------
+record { <nat1> = <v1>;* <nat2> = <v2>;* } :? record {  <nat1> = <t1>;* <nat3> = opt <t2>;* } ~> record { <nat1> = <v1'>;* <nat3> = null;* }
```
nomeata, line 971: This needs a tweak in light of https://github.com/dfinity/candid/issues/131 to allow for `reserved`

spec/Candid.md
nomeata: Probably a good time to revisit terminology.
 * Turning the binary form into the abstract value model (probably “decoding”)
 * Turning the textual form into the abstract value model (probably “parsing”)
 * Interpreting an abstract value at an expected type, i.e. this section. But what’s a good term?

spec/Candid.md
nomeata: Oh, `v ~> v' : T` is nice! Will do.

spec/Candid.md
nomeata: I expected it to be used more often. Remove.

spec/Candid.md
nomeata: Yes, this means that in the abstract model, roundtripping is a no-op. Note that `v : T` is the “precise” has-type relation (no extra fields, for example), so this is possible.

Where should I state this?

spec/Candid.md
nomeata: That formulation doesn't make sense to me, the antecent is never true.

Should `/\` be `→`?  Or should `v'` be behind some ∃?

nomeata (commented):
Mostly answers (and questions), will perform the changes after lunch.

spec/Candid.md
rossberg: Sounds good to me. The latter code be "coercing"?

spec/Candid.md
rossberg: Maybe the intro to the whole Decoding should somehow explain how this abstract model relates to actual serialisation/deserialisation.

spec/Candid.md
rossberg: Ah, oops, wrong quantifier:
```
(∃ v, v'. v : T /\ v ~> v' : T') ⇒ T <: T'
```

spec/Candid.md
rossberg: Will have a look.

rossberg:
Btw, nit: various inference bars are too short.

spec/Candid.md
nomeata: Now the antecent is far too weak, it says “`T` is a subtype of `T'` if _some_ values of type `T` make sense at type `T'`). This would, for example, relate all list types (use `v = []` and `v' = []`).

nomeata: How about I remove this section (so this can land if all else is fine), and we discuss this in a separate issue?

rossberg: Eh, you're right, I should turn my brain on. Okay, last try, if that's wrong as well, feel free to remove it. :)
```
(∃ v. v : T /\ ∀ v. v : T ⇒ ∃ v'. v ~> v' : T') ⇒ T <: T'
```
This looks a bit like monster-barring the uninhabited case, but I think it is a natural definition, saying: if serialised values exist for type T, and all such values can be deserialised at type T', then T ought to be considered a subtype of T'.

Uninhabited types tend to make everything harder...

spec/Candid.md
```diff
@@ -865,150 +872,147 @@ service { <methtype>;* } <: service { <methtype'>;* }
 service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>; <methtype'>;* }
 ```
 
-### Decoding
+### Coercing
```
rossberg, line 875:
```suggestion
### Coercion
```

spec/Candid.md
```diff
@@ -864,142 +872,210 @@ service { <methtype>;* } <: service { <methtype'>;* }
 service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>; <methtype'>;* }
 ```
 
-### Elaboration
+### Coercing
+
+This subtyping is implemented during the deserialization of Candid at an expected type. As as described in [Section Deserialisation](#deserialization), the binary value is conceptually first _decoded_ into an abstract value, and then _coerced_ to the expected type.
```
rossberg, line 877:
```suggestion
This subtyping is implemented during the deserialization of Candid at an expected type. As described in [Section Deserialisation](#deserialization), the binary value is conceptually first _decoded_ into an abstract value, and then _coerced_ to the expected type.
```

spec/Candid.md
```diff
@@ -864,142 +872,210 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+
+This section describes the covercion, as ternary relation `V ~> V' : T` to describe when a value `V` can be coerced to a value `V'` of type `T`. The fields `V` and `T` can be understood as inputs and `V'` as the output of this relation.
```
rossberg, line 879:
```suggestion
This section describes the covercion, as a ternary relation `V ~> V' : T` to describe when a value `V` can be coerced to a value `V'` of type `T`. The fields `V` and `T` can be understood as inputs and `V'` as the output of this relation.
```

spec/Candid.md
```diff
@@ -864,142 +872,210 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+---------------------
+false ~> false : bool
 
-------------------
-Nat <: Int ~> \x.x
+-----------------------
+<text> ~> <text> : text
 
+--------------------
+null :? null ~> null
```
rossberg, line 904: Missed one.

spec/Candid.md
```diff
@@ -864,142 +872,210 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
 
 
 #### Records
 
+Only records coerce at record type. Missing fields of option or reserved type turn into `null`.
+
+In the following rule, the `<nat1>*` field names are those present in both the value and the type, the `<nat2>*` field names only those in the value, and `<nat3>*` are those only in the type.
+```
+<v1> ~> <v1'> : <t1>
+<t3> = opt _ ∨ <t3> = reserved
```
rossberg, line 972:
```suggestion
opt empty <: <t3>
```

spec/Candid.md
```diff
@@ -864,142 +872,210 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
 
+* Roundtripping:
+  ```
+  (v : T) ⟺ v ~> v : T
+  ```
 
-## Open Questions
+* Uniqueness of decoding:
+  ```
+  v ~> v1 : T, v ~> v2 : T ⇒ v1 = v2
```
rossberg, line 1035:
```suggestion
  v ~> v1 : T /\ v ~> v2 : T ⇒ v1 = v2
```

spec/Candid.md
```diff
@@ -864,142 +872,210 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
+* Soundness of subtyping:
+  ```
+  T <: T' ⇒ ∀ v : T. ∃ v'. v ~> v' : T
+  ```
+
+* Higher-order soundness of subtyping
+  See <./IDL-Soundness.md>, with the following instantiations:
+  ```
+  s1 ~> s2 ≔ s2 <: s1
+  t1 <: t2 ≔ (∀ v. v : t1 ⇒ v ~> _:? t2 )
```
rossberg, line 1047:
```suggestion
  t1 <: t2 ≔ (∀ v. v : t1 ⇒ v ~> _ : t2 )
```

spec/Candid.md
nomeata:
```suggestion
null ~> null : null 
```

nomeata, line 904:
```suggestion
-------------------
null ~> null : null 
```

spec/Candid.md
nomeata: Had that initially and then changed it, as kinda wanted to avoid entangling the definition of the two predicates like this. No implementation or mechanizatoin would actually use to the full `<:` relation here…

But since you are quite fond of like subtyping relation, fine with me :-)

spec/Candid.md
nomeata:
```suggestion
  v ~> v1 : T ∧ v ~> v2 : T ⇒ v1 = v2
```

spec/Candid.md
nomeata: Nope, sorry still not good. 

In the antecent instead of `∀ v. v : T ⇒ ` I am sure you  `∀ v. v ~> _ : T ⇒ ` (i.e. not just the “canonical” values, but really “all values that can be deserialized at T). But I think that’s what you mean anyways.

But empty types still break it: This definition would still imply `vec Empty <: vec t` for all `t`.

And even then: Remember that coercion never fails for reference values (because we don't do a subtyping check when decoding and coercion), so any definition for completeness that is based on purely the `_ ~> _ : _` relation will probably relate all function types, which is obviously not sound (in the non-local; IDL-Soundess.md sense).

Let’s take this into a separate issue.

nomeata:
Ok, all comments handled one way or another

rossberg (approved):

spec/Candid.md
nomeata: Moved to https://github.com/dfinity/candid/issues/132

nomeata:
I’ll give @chenyan-dfinity a chance to comment before merging

spec/Candid.md
```diff
@@ -864,143 +872,200 @@ service { <methtype>;* } <: service { <methtype'>;* }
 service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>; <methtype'>;* }
 ```
 
-### Elaboration
+### Coercion
+
+This subtyping is implemented during the deserialization of Candid at an expected type. As described in [Section Deserialisation](#deserialization), the binary value is conceptually first _decoded_ into an abstract value, and then _coerced_ to the expected type.
+
+This section describes the covercion, as a ternary relation `V ~> V' : T` to describe when a value `V` can be coerced to a value `V'` of type `T`. The fields `V` and `T` can be understood as inputs and `V'` as the output of this relation.
```
chenyan-dfinity, line 879:
```suggestion
This section describes the coercion, as a ternary relation `V ~> V' : T` to describe when a value `V` can be coerced to a value `V'` of type `T`. The fields `V` and `T` can be understood as inputs and `V'` as the output of this relation.
```

spec/Candid.md
```diff
@@ -864,143 +872,200 @@ service { <methtype>;* } <: service { <methtype'>;* }
...
 
-To define the actual coercion function used during deserialisation of a value, we extend the subtyping relation to a ternary *elaboration* relation `T <: T' ~> f`, where `f` is a suitable coercion function of type `T -> T'`.
+Here `V` models untyped values, which form an abstract data model of both the message in transit (`V`), as well as the result of the coercion (`V'`). In the following, we re-use the syntax of the textual representation, with the following changes to make it free of overloading:
 
+ * Number literals (`<primval>`) must be immediately enclosed with an `<annval>` that clarifies the precise number type.
+ * The canonical value of type `reserved` is expressed as `(null : reserved)`.
+ * No other use of `<annval>`.
```
chenyan-dfinity, line 885: We also need annval for `null : opt _`

spec/Candid.md
```diff
@@ -1227,3 +1281,29 @@ The same representation is used for function results.
...
 
 * It is unspecified how the pair (B,R) representing a serialised value is bundled together in an external environment.
+
+### Deserialisation
+
+Deserialisation at an expected type sequence `(<t'>,*)` proceeds by
+
+ * checking for the magic number `DIDL`
+ * using the inverse of the `T` function to decode the type definitions `(<t>*)`
+ * using the inverse of the `M` function, indexed by `ts`, to decode the values `(<v>*)`
```
chenyan-dfinity, line 1291:
```suggestion
 * using the inverse of the `M` function, indexed by `(<t>*)`, to decode the values `(<v>*)`
```

chenyan-dfinity (approved):
LGTM. It's indeed much cleaner!

spec/Candid.md
nomeata: No, I am pretty sure we don't! That’s because 
 * On the left of `~>`, it doesn't matter which `null` it is (any `null` behaves the same) 
 * On the right of `~>`, the type is given, so again there is only no `null` value.

So it is fine to have a single `null` value in the `V` set.

This is not true for overloaded numbers; `(5 : nat) ~>`  and `(5 : int) ~>` behave differently.

rossberg: In other words, `null` has a principal type, numbers don't. (Which makes me wonder, why do we need a special case for `reserved`?)

nomeata: Because the algebra of _binary_ values has a dedicated `reserved` value (`DIDL\x00\x01\x70`), but our textual representation doesn’t. The `(null : reserved)` could just as well be `("reserved" : reserved)` or `(record {} : reserved)`; it’s arbitrary and, in a way, always lie.

For me, the conceptual “abstract value data model” _has_ a dedicated `reserved` value, and I think it would be good to have it in the textual representation as well, to avoid the above wart in the formalism, but also to not force debug output to come up with such arbitrary values when decoding a message with a `reserved` value:
```
~/dfinity/haskell-candid $ cabal -v0 run hcandid -- --decode --rust "DIDL\x00\x01\x70"
((null : reserved))
```

nomeata: More evidence that it might be useful to have it: All implementations that have a “value” type seemed to feel the need to add it to their data type:
* https://github.com/dfinity/candid/blob/03773cc2c3d86b898d78e3161be3c6bae34dbd97/rust/candid/src/parser/value.rs#L37
* https://github.com/nomeata/haskell-candid/blob/36202a9aef4387925c0d0de23a0e3645ed7d2b3e/src/Codec/Candid/Types.hs#L142

PR #130, opened 2020-11-03
Implement CandidType for std::time::SystemTime

### LINKED ISSUE #129: Implement CandidType for std::time::SystemTime
serde serializes `SystemTime` as `Duration` from `std::time::UNIX_EPOCH`. Having a `CandidType` implementation for it that does the same would be useful when sending `SystemTime` in a message. 

I actually tried to implement this myself but failed because I don't understand how to generate named fields in candid. The serde implementation for `SystemTime` wants a record with `secs_since_epoch` and `nanos_since_epoch` fields, for the second and nanosecond parts, respectively. I don't know if that's possible to do in candid.

osa1:
Fixes #129

### DIFF
--- a/rust/candid/src/types/impls.rs
+++ b/rust/candid/src/types/impls.rs
@@ -260,3 +260,52 @@ tuple_impls! {
     15 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9 10 T10 11 T11 12 T12 13 T13 14 T14)
     16 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9 10 T10 11 T11 12 T12 13 T13 14 T14 15 T15)
 }
+
+impl CandidType for std::time::SystemTime {
+    fn id() -> TypeId {
+        TypeId::of::<std::time::SystemTime>()
+    }
+
+    fn _ty() -> Type {
+        Type::Record(vec![
+            Field {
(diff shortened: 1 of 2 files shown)


rust/candid/src/types/impls.rs
```diff
@@ -260,3 +260,52 @@ tuple_impls! {
...
+impl CandidType for std::time::SystemTime {
+    fn id() -> TypeId {
+        TypeId::of::<std::time::SystemTime>()
+    }
+
+    fn _ty() -> Type {
+        Type::Record(vec![
+            Field {
+                id: Label::Named("nanos_since_epoch".to_owned()),
+                ty: u32::_ty(),
```
chenyan-dfinity, line 273:
```suggestion
                ty: u32::ty(),
```

rust/candid/src/types/impls.rs
```diff
@@ -260,3 +260,52 @@ tuple_impls! {
...
+            },
+            Field {
+                id: Label::Named("secs_since_epoch".to_owned()),
+                ty: u64::_ty(),
```
chenyan-dfinity, line 277:
```suggestion
                ty: u64::ty(),
```

rust/candid/src/types/impls.rs
chenyan-dfinity: `ty()` is the memoized version of `_ty()` to handle recursion. It won't make a difference here, but generally good coding style.

chenyan-dfinity (approved):

PR #134, opened 2020-11-04
More coercions: reserved ~> null : opt _

nomeata:
I won’t believe in this system until we have a formal proof… anyways,
still finding holes.

Imagine the sender evolves as follows:
```
record {} <: record { foo : reserved }
```
and the receiver evolves (using fancy new opt field rules) as follows:
```
record {} :> record { foo : opt bool }
```

Both are allowed, so coercion must not fail. One way to fix that is to
add
```
(null : reserved) ~> null : opt <t>
```
to the rules.

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -944,11 +944,14 @@ An optional value coerces at an option type, if the constituent value coerces at
 opt <v> ~> opt <v'> : opt <t>
 ```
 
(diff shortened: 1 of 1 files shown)


nomeata:
@rossberg , did you have a chance to double check this?

rossberg:
I think you're right.

chenyan-dfinity (approved):

Issue #146, opened 2020-12-03
Subtyping: Transitivity of opt-to-constituent rule


nomeata:
I started to do some baby steps in formalizing a “mini candid” in Coq, starting with the rules that seem safe.

But adding the “specialize to constituent type” rule is causing problems:

If the rule is 
```
not (null <: t2)     t1 <: t2
--------------------------------
        t1 <: opt t2
```
as it is now, then this is not transitive. We have `int <: opt int` (by this rule), `opt int <: opt opt int` (by the normal rule), but not `int <: opt opt int` (because of the `not (null <: t2)` restriction).

This restriction was added in #135, upon my suggestion. Without that restriction, we have the problems described in https://github.com/dfinity/candid/issues/135, i.e. bad interaction with the “catch decoding rule”, and the problem that
```
    t1 <: t2
------------------
  t1 <: opt t2
```
doesn't make progress if `t2 = opt t2`.

Or, to respond to @rossberg says in https://github.com/dfinity/candid/issues/135#issuecomment-728895485
> Interesting suggestion. I suppose options of options are not something you would see often in an interface, so limiting their evolution might be okay. How confident can we be that this has no problems?
> 
> I wish we had spare cycles to mechanise this, and prevent further holes.

I am no longer confident that this has no problems.

One way to maybe! taper over the problem is to restrict the other rule as well:
```
not (null <: t2)     t1 <: t2
--------------------------------
     opt  t1 <: opt t2
```
so that you can’t go from `opt int` to `opt opt int`… but that essentially disallows evolution under `opt opt` and might be better served by disallowing `opt opt`… probably not nice. Or maybe this rule:
```
(null <: t1) iff (null <: 2)      t1 <: t2
---------------------------------------
             opt  t1 <: opt t2
```
which rules out `opt int <: opt opt int` without ruling out `opt nat <: opt int` nor `opt opt nat <: opt opt int` ?

nomeata:
Maybe this is all fine _together_ with the other special behavior for `opt` (the opportunistic decoding). But that one is dubious for other reasons, see #141. And even if it is fine, I wonder if we get weird transitive behavior.

nomeata:
With the extra `(null <: t1) iff (null <: 2)` restriction, I can prove transitivity, see https://github.com/dfinity/candid/pull/147. Hmm, not sure I like that.

crusso:
[...]

Issue #141, opened 2020-11-22
Still not sound: Gradual typing vs. opportunistic decoding


nomeata:
Looks like we still haven’t found the grail yet…

TL;DR: Our relaxed coercion rules for opt (#110 and following) may lead to “hard” deserialization errors later on.

I seems that the following can’t all hold:

 * We are sound in the higher-order case (as per [IDL-Soundness](https://github.com/dfinity/candid/blob/master/spec/IDL-Soundness.md))
 * We have no subtyping checks while deserialization (one of our [Design goals](https://github.com/dfinity/candid/blob/master/spec/Candid.md#design-goals)).
 * Type Erasure (another of the design goal)
 * Record Extensibility, at least in the form implemented right now.

## Counter-example

Here is the counter example:

```
service A1 : { bar : () -> () }
service B1 : { baz : () -> () } 

service C : { 
  init : (service A1, service B1) -> (); // stores (a1, b1)
  run : () -> () ;  // runs a1.bar(b1.baz())
}
```
If we pass `A1` and `B1` to `C.init`, all is well. `C.run` works fine.

Now we upgrade `A1` to `A2` and `B1` to `B2`, as follows:
```
type WantsBool = service : { foo : (bool) -> () }
type WantsInt = service : { foo : (int) -> () }
service wantsInt : WantsInt
service A2 : {  bar : (opt WantsBool) -> () } // calls `arg.foo(true)` if `arg` is not `null`
service B2 : { baz : () -> (opt WantsInt) }  // returns (opt wantsInt)
```
These upgrades are permitted by our `:<`, and actually desirable (i.e. restricting `<:` to avoid “obviously bogus” upgrades doesn’t help)

But now `C.run` will cause B2 to pass `opt wantsInt` to `A2`. This will succeed at deserialization (because “Function and services references coerce unconditionally”). Now A2 invokes `wantsInt.foo(true)`, and the deserialization of `(true)` at expected type `(bool)` will fail in `wantsInt`. 

## Generic Soundness proof

In https://github.com/dfinity/candid/blob/master/spec/IDL-Soundness.md#proof-for-canonical-subtyping we have a generic soundness proof for “Canonical subtyping”, that says “A solution with canonical subtyping (transitive, contravariant) is sound.”. And our `<:` is that!

But it’s not “decompositional” any more. In particular, 
> If `t1 <: t2` and `s1 in t1 <: s2 in t2` then `s1 <: s2`.

(as required in the generic proof) doesn’t hold .

[...]

PR #147, opened 2020-12-03
Coq: MiniCandid

nomeata:
This contains an initial formalization of Candid in Coq. It covers these types:
`nat`, `int`, `null`, `opt t`, `empty`, `reserved`

It considers two different formalisms:

* `NoOpportunisticDecoding`
  This is without `t <: opt t'`, but with `t <: opt t`.
  Things go through, although the `opt t <: opt t'` rule has to be tweaked to keep subtyping transitive (see https://github.com/dfinity/candid/issues/146).

* `OpportunisticDecoding`
  This is with `t <: opt t'`.
  Because of the negative hypotheses in the coercion relation, this can’t be defined as a simple inductive relation (which is a seroius smell!).
  So instead I define it as a function (which is mostly straight forward), and prove the properties there.
  We _could_ take this as indication that maybe our spec should also just define it via equations. We don’t gain anything from the relational presentation, I think, and implementors all anyways implement functions.
  Nevertheless, I did prove that the relation defined by admits the intro and induction rules that _would_ come out of the inductive relation (ignoring the negative hypotheses). This revealed some issues with the way the rules are presented, will create a PR for that soon.

In both cases, I prove
 * Correctness of decoding
 * Roundtripping
 * Uniqueness of decoding
 * Soundness of subtyping
 * Transitivity of subtyping

I do _not_ prove Higher-order soundness, because we know it does not hold (see https://github.com/dfinity/candid/issues/141).

This also uses some proof-of-concept “named Coq cases“ feature, see https://www.joachim-breitner.de/blog/777-Named_goals_in_Coq.


coq/MiniCandid.v
```diff
@@ -0,0 +1,166 @@
...
+Proof. intros x. apply ReflST. Qed.
+
+Lemma is_not_opt_like_type_contravariant:
+  forall t1 t2,
+     is_not_opt_like_type t1 -> t2 <: t1 -> is_not_opt_like_type t2.
+Proof. intros. destruct t1, t2; easy. Qed.
+
+Theorem subtyping_trans: transitive _ Subtype.
+Proof.
+  cofix Hyp.
```
nomeata, line 151: Just to show off: This is a proof by coinduction.

crusso: Bravo!

nomeata:
This is ready for `master`; I updated the PR summary with a overview of what we are doing.

nomeata:
I woudn’t mind parking this in `master`.

chenyan-dfinity (approved):

PR #151, opened 2020-12-11
Spec: Change subtype-to-constituent rule

nomeata:
the coq formalization in #147 shows that the old rule didn’t quite work
(it was confused about `5 ~> ? : opt null` I believe), but this variant
does. I think all implemnetations do the right thing anways (so maybe
rewriting this as a partial coercion _function_ would be helpful)


nomeata:
Closing in favor of #168

PR #153, opened 2020-12-18
support reference types

### LINKED ISSUE #7: Support reference types

chenyan-dfinity:
* Fix #7
* output blob format for `didc encode`

### DIFF
--- a/test/reference.test.did
+++ b/test/reference.test.did
@@ -22,9 +22,8 @@ assert blob "DIDL\02\6a\01\71\01\7d\00\69\01\03\66\6f\6f\00\01\01\01\03\ca\ff\ee
 assert blob "DIDL\02\6a\01\71\01\7d\00\69\02\03\66\6f\6f\00\04\66\6f\6f\32\00\01\01\01\03\ca\ff\ee"
   == "(service \"w7x7r-cok77-xa\")" : (service { foo : (text) -> (nat); foo2 : (text) -> (nat) }) "service";
 
-// TODO: should this fail?
-//assert blob "DIDL\02\6a\01\71\01\7d\00\69\01\03\66\6f\6f\68\01\01\01\03\ca\ff\ee" !: (service {}) "service { foo: principal }";
-//assert blob "DIDL\02\6e\7e\69\01\03\66\6f\6f\00\01\01\01\03\ca\ff\ee" !: (service {}) "service { foo: opt bool }";
+//assert blob "DIDL\02\6a\01\71\01\7d\00\69\01\03\66\6f\6f\68\01\01\01\03\ca\ff\ee" !: (service { foo: () -> () }) "service { foo: principal }";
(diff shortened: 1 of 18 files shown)


test/reference.test.did
```diff
@@ -0,0 +1,40 @@
...
+assert blob "DIDL\01\69\00\01\00\01\03\ca\ff\ee" == "(service \"w7x7r-cok77-xa\")" : (service {}) "service";
+assert blob "DIDL\01\69\00\01\00\01\03\ca\ff\ee" == "(service \"w7x7r-cok77-xa\")" : (service {}) "service";
+assert blob "DIDL\02\6a\01\71\01\7d\00\69\01\03\66\6f\6f\00\01\01\01\03\ca\ff\ee"
+  == "(service \"w7x7r-cok77-xa\")" : (service { foo : (text) -> (nat) }) "service";
+assert blob "DIDL\02\6a\01\71\01\7d\00\69\02\03\66\6f\6f\00\04\66\6f\6f\32\00\01\01\01\03\ca\ff\ee"
+  == "(service \"w7x7r-cok77-xa\")" : (service { foo : (text) -> (nat); foo2 : (text) -> (nat) }) "service";
+
+// TODO: should this fail?
+//assert blob "DIDL\02\6a\01\71\01\7d\00\69\01\03\66\6f\6f\68\01\01\01\03\ca\ff\ee" !: (service {}) "service { foo: principal }";
+//assert blob "DIDL\02\6e\7e\69\01\03\66\6f\6f\00\01\01\01\03\ca\ff\ee" !: (service {}) "service { foo: opt bool }";
```
chenyan-dfinity, line 27: @nomeata Should we check for `service { foo : opt bool }` in the type table? This requires a second pass on the type table.

nomeata: I am too lazy to manually decode these blob.

But note that our current design says that there are _no_ subtyping checkes when decoding reference types (see https://github.com/dfinity/candid/blob/master/spec/Candid.md#references-1). So these should probably all pass.

This means that including the type of references in the type table is a bit silly, and a historic artifact. But maybe we’ll change that due to #141.

chenyan-dfinity: It's not about subtyping. It's when the service type is ill-formed, e.g. the service type in the type table is `service { f : opt bool }`.

nomeata: Ah, now I see, thanks.

I guess we should reject that; this is not in the range of the `T` function, so one can argue that the spec disallows that.

chenyan-dfinity: Then we need a second pass on the type table, as the type index may not be available when we see the service type. Or we can defer the check when getting to the `M` part.

nomeata: > Then we need a second pass on the type table, as the type index may not be available when we see the service type.

Oh, indeed!

> Or we can defer the check when getting to the M part.

Hmm, probably not 100% correct that way, I think  we want to reject bad type tables even if there are no such values (e.g. `opt service { foo : bool }` with value `null`)

chenyan-dfinity: Okay, we don't need a second pass, we just need to check the type table lazily.

PR #155, opened 2021-01-05
fix record_nesting_depth for native Rust types

chenyan-dfinity:
All `deserialize_T` functions can be entry points for deserialization. We have to reset the counter for `record_nesting_depth` for all these functions to track the empty record type properly.


matthewhammer:
Now I get this error (with 17 and 16 printed out):

```Candid decoding error: Deserialize("There is an infinite loop in the record definition, the type is isomorphic to an empty type 17/16", "Trailing type: [I(4), U(1136381571), I(7), U(1269255460), I(10), I(4), U(1136381571), I(7),```

matthewhammer:
The fix works!

PR #158, opened 2021-01-07
A Candid users’s guide

### LINKED ISSUE #144: Developer can read documentation for Candid
We keep getting user questions about Candid, e.g., about its syntax. But all documentation we currently have is the "spec", which is not really targeted at users and is overly technical.

We should have some primary "user guide" that is accessible for programmers, not implementers.

nomeata:
this is the beginning of a user-facing, user-friendly guide.

Fixes #144.

### DIFF
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -749,6 +749,7 @@ Type syntax::
 variant {}
 variant { ok : nat; error : text }
 variant { "name with spaces" : nat; "unicode, too: ☃" : bool }
+variant { spring; summer; fall; winter }
 ....
 
 Textual syntax::
@@ -756,6 +757,7 @@ Textual syntax::
 ....
 variant { ok = 42 }
 variant { "unicode, too: ☃" = true }
+variant { fall }
 ....
 
 Subtypes::
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -891,7 +891,7 @@ Dually, the following modifications to a function type change it to a supertype:
 
 Corresponding Motoko type::
 
-Candid function types correspond to `shared` Motoko functions, with the result type wrapped in `async`. Arguments resp. results become tuples, unless there is exactly one, in which case it is used directly:
+Candid function types correspond to `shared` Motoko functions, with the result type wrapped in `async` (unless they are annotated with `oneway`, then the result type is simply `()`).  Arguments resp. results become tuples, unless there is exactly one, in which case it is used directly:
 +
 ....
 type F0 = func () -> ();
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -93,7 +93,7 @@ The method `reverse` expects a single parameter of type `text` and returns one v
 
 The method `divMod` expects and returns two values, all of type `int`. In its signature, we use the ability to name the parameters and result values.
 
-CAUTION
+[CAUTION]
 ====
 Naming the parameters or results of a method is purely for documentation purposes, but does not actually change the method’s type, or the values being passed. Parameter and results are identified by their _position_, independent of the name.
 
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -11,7 +11,7 @@ If you are about to _implement_ Candid, want to learn about its internals or its
 
 == What is Candid?
 
-Candid is an _Interface description language_, and is used to describe the public interface of a service (such as a canister running on the Internet Computer). It is language agnostic, and thus allows seamless interoperation between services and front-ends written in different programming languages (Motoko, Rust, JavaScript, etc.)
+Candid is an _Interface description language_, and is used to describe the public interface of a service (such as a canister running on the Internet Computer). It is language agnostic, and thus allows interoperation between services and front-ends written in different programming languages (Motoko, Rust, JavaScript, etc.)
 
 A typical interface description in Candid might look as follows:
 ....
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -876,8 +876,9 @@ The following modifications to a function type change it to a subtype (cf. the r
  * The result type list may be extended
  * The parameter type list may be shortened.
  * The parameter type list may be extended with optional arguments (type `opt …`).
- * Existing parameter types may be changed to to a _supertype_(!).
+ * Existing parameter types may be changed to to a _supertype_ (!). In other words: The function type is _contravariant_ in the argument type.
  * Existing result types may be changed to a subtype.
+
 
 Supertypes::
 
@@ -886,7 +887,7 @@ Dually, the following modifications to a function type change it to a supertype:
  * The result type list may be shortened
  * The result type list may be extended with optional arguments (type `opt …`)
  * The parameter type list may be extended.
- * Existing parameter types may be change to to a subtype(!).
+ * Existing parameter types may be change to to a _subtype_ (!). In other words: The function type is _contravariant_ in the argument type.
  * Existing result types may be changed to a supertype.
 
 Corresponding Motoko type::
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -1055,7 +1055,7 @@ service { foo : (first_name : text, middle_name : reserved, last_name : text) ->
 to indicate that `foo` used to take a second argument, but you no longer care about that.
 
 
-You can also avoid this problem if you use the idiom where function with possibly changing arguments take a single record, i.e.
+You can avoid this pitfall by adopting the pattern any function that is anticipated to have changing arguments, or whose arguments can only be distinguished by position, not type,  is declared to take a single record, i.e.
 ....
 service { foo : (record { first_name : text; middle_name : text; last_name : text}) -> () }
 ....
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -150,7 +150,7 @@ As a Candid user, you do not have to worry about the details of this binary form
 
 Services evolve over time: They gain new methods, existing methods return more data, or expect additional arguments. Usually, service authors want to do that without breaking existing clients.
 
-Candid supports such evolution by defining precise rules that indicate when the new service type will still be able to communicate with all other parties that are using the previous interface description. The underlying formalim is that of _subtyping_.
+Candid supports such evolution by defining precise rules that indicate when the new service type will still be able to communicate with all other parties that are using the previous interface description. The underlying formalism is that of _subtyping_.
 
 Services can safely evolve in the following ways:
 
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -718,7 +718,7 @@ Supertypes::
 
 Supertypes of a record are record types with some fields removed, some fields’ types changed to supertypes, or with optional fields added.
 +
-The latter is what allows you to extend your argument records with additional fields. Clients using the old interface will not include the field in their record, which will decode as `null`.
+The latter is what allows you to extend your argument records with additional fields. Clients using the old interface will not include the field in their record, which will decode, when expected in the upgraded service, as `null`.
 +
 For example, if you have a function expecting a record of type
 +
--- a/docs/modules/candid-guide/index.adoc
+++ b/docs/modules/candid-guide/index.adoc
@@ -64,7 +64,7 @@ Candid is a strongly typed system with a set of types that canonically cover mos
 
 All types are described in detail in the <<reference, reference section>>.
 
-The philosophy behind this set of types is that they are sufficient to describe the _structure_ of data, so that information can be encoded, passed around and decoded, but intentionally do not describe _semantics_ constraints beyond what’s needed to describe the representation (e.g. no way to express that a numer should be even, or that the elements of a vector are sorted).
+The philosophy behind this set of types is that they are sufficient to describe the _structure_ of data, so that information can be encoded, passed around and decoded, but intentionally do not describe _semantic_ constraints beyond what’s needed to describe the representation (e.g. no way to express that a number should be even, that a vector has a certain length, or that the elements of a vector are sorted).
 
 Furthermore, the set of types is chosen to allow a natural mapping to the respective host language (Motoko, Rust, JavaScript, etc.), settling on reasonable, canonical choices.
 
@@ -77,7 +77,7 @@ The simplest service description is
 ....
 service : {}
 ....
-and describes a service with no public methods. This is not very useful, so lets add a simple method:
+and describes a service with no public methods. This is not very useful, so let's add a simple method:
 ....
 service : {
   ping : () -> ();
@@ -102,11 +102,11 @@ The method `divMod` expects and returns two values, all of type `int`. In its si
 ====
 Naming the parameters or results of a method is purely for documentation purposes, but does not actually change the method’s type, or the values being passed. Parameter and results are identified by their _position_, independent of the name.
 
-In particular, Candid will does not prevent you from changing the type to
+In particular, Candid does not prevent you from changing the type to
 ....
   divMod : (dividend : nat, divisor : nat) -> (mod : nat, div : nat);
 ....
-or passing the above `divMod` to a service expeting a method that returns `mod` first.
+or passing the above `divMod` to a service expecting a method that returns `mod` first.
 
 This is thus very different from named _record_ fields, which are semantically relevant.
 ====
@@ -120,7 +120,7 @@ type address = record {
   country : text;
 };
 service address_book : {
-  set_address: (name : text, address) -> ();
+  set_address: (name : text, addr : address) -> ();
   get_address: (name : text) -> (opt address) query;
 }
 ....
@@ -131,17 +131,17 @@ In the last example, we also see the use of the `query` annotation. This Interne
 
 === Encoding and decoding
 
-The point of Candid is to allow seamless invocation of service methods, passing arguments encoded to a binary format and transferred by an underlying transporation method (such as messages into or within the Internet Computer), and decoded on the other side.
+The point of Candid is to allow seamless invocation of service methods, passing arguments encoded to a binary format and transferred by an underlying transportation method (such as messages into or within the Internet Computer), and decoded on the other side.
 
-As a Candid user, you do not have to worry about the details of this binary format. If you plan to _implemnet_ Candid yourself (e.g. for a new host language), you can consult the Candid specification for details. However, some aspects of the format are worth knowing:
+As a Candid user, you do not have to worry about the details of this binary format. If you plan to _implement_ Candid yourself (e.g. for a new host language), you can consult the Candid specification for details. However, some aspects of the format are worth knowing:
 
 * The Candid binary format starts with `DIDL…` (or, in hex, `4449444c…`). If you see this in some low-level log output, you are very likely observing a Candid-encoded value.
 
 * The Candid binary format always encodes _sequences_ of values, because methods parameters and results are sequences of types.
 
-* The binary format is quite compact. A `(vec nat64)` with 125000 entries takes 1000007 bytes.
+* The binary format is quite compact. A `(vec nat64)` with 125 000 entries takes 1 000 007 bytes.
 
-* The binary is self-describing, and includes a (condended) description of type of the values therein. This allows the receiving side to detect if a message was sent at a different, incompatible type.
+* The binary is self-describing, and includes a (condensed) description of type of the values therein. This allows the receiving side to detect if a message was sent at a different, incompatible type.
 
 * As long as the sending side serializes the arguments at the same type that the receiving side expects them, deserialization will succeed.
 
@@ -165,7 +165,7 @@ See the reference sections the supertypes and subtypes of a given type.
 
 [TIP]
 ====
-Consider a serivce with the following API:
+Consider a service with the following API:
 ....
 service counter : {
   add : (nat) -> ();
@@ -228,7 +228,7 @@ actor {
   public func subscribe(handler : func (Int) -> async ()) { … }
 }
 ....
-and it will have the interface shown above.
+and the compiler can automatically generate the interface shown above.
 
 In other languages, like Rust, you can still develop your service while only
 touching Rust types, but for now you still have to write the Candid interface description manually.
@@ -237,7 +237,7 @@ In either case it is helpful to know the mapping between host language types and
 
 === Candid tools and libraries
 
-The Motoko language has built-in support for Candid.
+The Motoko language has built-in support for Candid and uses it implicitly.
 
 To create Candid services or clients in Rust, you can use the https://crates.io/crates/candid[Candid crate].
 
@@ -253,7 +253,7 @@ The https://github.com/dfinity/candid/tree/master/tools/ui[Candid UI] creates a 
 
 Now that you have a good overview of Candid and its various moving parts, this section gives you concrete instructions to achieve certain tasks.
 
-=== Interact with canisters on the command line?
+=== How do I interact with canisters on the command line?
 
 The `dfx` tool provides the `dfx canister call` command, which understands <<textual-values,Candid textual values>>:
 ....
@@ -262,24 +262,26 @@ dfx canister call msgt2-6nlzx-xq add_entry ("John Doe", record { street = "Haupt
 
 See TODO to read more about how to use the `dfx` tool.
 
-=== Interact with canisters via Candid UI?
+=== How do I interact with canisters via Candid UI?
 
-=== Use Candid as a Motoko canister developer?
+TODO
+
+=== How do I use Candid as a Motoko canister developer?
 
 If you are writing a service from scratch in Motoko, then you need to take no special actions: The Motoko compiler will translate the signature of your Canister’s top-level actor or actor class into a Candid description, and the `dfx` build tool will make sure it ends up where it needs to be.
 
 If you want to implement a _specific_ interface, for example because you want to interact with a service that expects your canister to have that interface, you can consult the 
 reference below to figure which Motoko types to use to achieve this effect. In the future, this will be simplified.
 
-=== Use Candid as Rust canister developer?
+=== How do I use Candid as Rust canister developer?
 
 TODO: Someone who implements rust canisters should write this section.
 
-=== Use Candid as a JavaScript front-end developer?
+=== How do I use Candid as a JavaScript front-end developer?
 
 TODO: Someone who implements Frontends should write this section
 
-=== Implement Candid?
+=== How do I implement Candid?
 
 If you want to create a new Candid implementation, this document will not suffice. In this case, consult the https://github.com/dfinity/candid/blob/master/spec/Candid.md[Candid specification].
 
@@ -466,13 +468,13 @@ Corresponding JavaScript values::
 TODO
 
 [#type-floatN]
-=== Type float16 and float32
+=== Type float32 and float64
 
-The types `float16`, `float32` represent IEEE 754 floating point numbers, in single precision (32 bit) and double precision (64 bit).
+The types `float32`, `float64` represent IEEE 754 floating point numbers, in single precision (32 bit) and double precision (64 bit).
 
 Type syntax::
 
-`float16`, `float32`
+`float32`, `float64`
 
 Textual syntax::
 
@@ -494,7 +496,7 @@ Corresponding Motoko type::
 
 `float64` corresponds to `Float`.
 +
-`float32` does _not_ have a representation in Motoko. Candid interfaces using `float32` cannot be served from or used from Motoko programs.
+`float32` does _not_ currently have a representation in Motoko. Candid interfaces using `float32` cannot be served from or used from Motoko programs.
 
 Corresponding Rust type::
 
@@ -507,7 +509,7 @@ TODO
 [#type-bool]
 === Type bool
 
-The type of truths values.
+The type of truth values.
 
 Type syntax::
 
@@ -598,7 +600,7 @@ The canonical rules for subtyping with `opt` are:
 * `null` is a subtype of `opt t'`.
 * `t` is a subtype of `opt t` (unless `t` itself is `null`, `opt …` or `reserved`)
 +
-In addition, for technical reasons related to upgrading and higher-order services, _every_ type is a subtype of `opt t`. Users are advised, however, to not directly make use of that rule.
+In addition, for technical reasons related to upgrading and higher-order services, _every_ type is a subtype of `opt t`, yielding `null` if the types do not match. Users are advised, however, to not directly make use of that rule.
 
 Supertypes::
 
@@ -680,9 +682,9 @@ record { first_name : text; second_name : text; score : nat; country : text }
 
 Supertypes::
 
-Supertypes of a record are record types with any some fields removed, some fields’ types changed to supertypes, or with optional fields added.
+Supertypes of a record are record types with some fields removed, some fields’ types changed to supertypes, or with optional fields added.
 +
-The latter is what allows you to extend your argument records with additional arguments. Clients using the old interface will not include the field in their record, which will decode as `null`.
+The latter is what allows you to extend your argument records with additional fields. Clients using the old interface will not include the field in their record, which will decode as `null`.
 +
 For example, if you have a function expecting a record of type
 +
@@ -720,14 +722,14 @@ A `variant` type represents a value that is from exactly one of the given cases,
 ....
 type shape = variant {
   dot : null;
-  circle : record { radius : float64 };
+  circle : float64;
   rectangle : record { width : float64; height : float64 };
   "💬" : text;
 };
 ....
 is either a dot, or a circle (with a radius), or a rectangle (with dimensions) or a speech bubble (with a text, and showing off unicode label names).
 
-The lables in variants are, just like the labels in records, actually numbers, and string labels refer to their hash value.
+The tags in variants are, just like the labels in records, actually numbers, and string tags refer to their hash value.
 
 Often, some or all of the the tags do not carry data. It is idiomatic to then use the `null` type, as in the `dot` above. In fact, Candid encourages this by allowing you to omit the `: null` type annotation in variants, so
 ....
@@ -760,7 +762,7 @@ Subtypes::
 
 Subtypes of a variant type are variant types with some tags removed, and the type of some tags themselves changed to a subtype.
 +
-If you want to be able to _add_ new tags in variants in a method result, you can do so if the variant it itself wrapped in `opt …`. This requires planning ahead! When you design an interface, instead of writing
+If you want to be able to _add_ new tags in variants in a method result, you can do so if the variant is itself wrapped in `opt …`. This requires planning ahead! When you design an interface, instead of writing
 +
 ....
 service { get_member_status (member_id : nat) -> (variant {active; expired}); }
@@ -786,7 +788,7 @@ Variant types are represented as Motoko variant types, e.g.
 ....
 type Shape = {
   #dot : ();
-  #circle : { radius : Float };
+  #circle : Float;
   #rectangle : { width : Float; height : Float };
   #_2669435721_ : Text;
 };
@@ -806,7 +808,7 @@ TODO
 [#type-func]
 === Type func (…) -> (…)
 
-Candid is designed to support higher-order use cases, where a service may receive or provide references to other serivces and their methods. The `func` type lies at the code or this: It indicates the function’s _signature_ (argument and results types, annotations), and values of this type are references to functions with that signature.
+Candid is designed to support higher-order use cases, where a service may receive or provide references to other serivces or their methods, e.g., as callbacks. The `func` type is central to this: It indicates the function’s _signature_ (argument and results types, annotations), and values of this type are references to functions with that signature.
 
 Type syntax::
 +
@@ -822,7 +824,7 @@ See the <<candid-service-descriptions,note on function parameter names>>.
 +
 The supported annotation are:
 +
-* `query` indicates that the referenced function is a query method, meaning it can does not alter the state of its canister, and that it can be invoked using the cheaper “query call” mechanism.
+* `query` indicates that the referenced function is a query method, meaning it does not alter the state of its canister, and that it can be invoked using the cheaper “query call” mechanism.
 * `oneway` indicates that this function returns no response, intended for fire-and-forget scenarios.
 
 Textual syntax::
@@ -842,8 +844,8 @@ The following modifications to a function type change it to a subtype (cf. the r
  * The result type list may be extended
  * The parameter type list may be shortened.
  * The parameter type list may be extended with optional arguments (type `opt …`).
- * Existing parameter types may be change to to a supertype.
- * Existing result types may be _changed_ to a subtype(!).
+ * Existing parameter types may be changed to to a _supertype_(!).
+ * Existing result types may be changed to a subtype.
 
 Supertypes::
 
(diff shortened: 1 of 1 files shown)


nomeata:
(Pre-lunch stash push, still WIP of course)

nomeata:
More work done, need a break now. Primariliy missing are the reference types (func, serivce, principal).

nomeata:
Reference types added. I think this can be added now; we don't have to fix all TODOs in a PR.

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+TODO
+
+Corresponding JavaScript values::
+
+TODO
+
+[#type-floatN]
+=== Type float16 and float32
+
+The types `float16`, `float32` represent IEEE 754 floating point numbers, in single precision (32 bit) and double precision (64 bit).
```
rossberg, line 471:
```suggestion
The types `float32`, `float64` represent IEEE 754 floating point numbers, in single precision (32 bit) and double precision (64 bit).
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+0xDEAD.BEEF
+0xDEAD.BEEFP-10
+0xDEAD.BEEFp+10
+....
+
+Corresponding Motoko type::
+
+`float64` corresponds to `Float`.
++
+`float32` does _not_ have a representation in Motoko. Candid interfaces using `float32` cannot be served from or used from Motoko programs.
```
rossberg, line 497: Does it not map to Float in Motoko?
```suggestion
`float32` does _not_ currently have a representation in Motoko. Candid interfaces using `float32` cannot be served from or used from Motoko programs.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+TODO
+
+Corresponding JavaScript values::
+
+TODO
+
+[#type-bool]
+=== Type bool
+
+The type of truths values.
```
rossberg, line 510:
```suggestion
The type of truth values.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+func (dividend : nat, divisor : nat) -> (div : nat, mod : nat);
+func () -> (int) query
+func (func (int) -> ()) -> ()
+....
++
+See the <<candid-service-descriptions,note on function parameter names>>.
++
+The supported annotation are:
++
+* `query` indicates that the referenced function is a query method, meaning it can does not alter the state of its canister, and that it can be invoked using the cheaper “query call” mechanism.
```
rossberg, line 825:
```suggestion
* `query` indicates that the referenced function is a query method, meaning it does not alter the state of its canister, and that it can be invoked using the cheaper “query call” mechanism.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+Subtypes::
+
+The following modifications to a function type change it to a subtype (cf. the rules in <<upgrades>>):
++
+ * The result type list may be extended
+ * The parameter type list may be shortened.
+ * The parameter type list may be extended with optional arguments (type `opt …`).
+ * Existing parameter types may be change to to a supertype.
+ * Existing result types may be _changed_ to a subtype(!).
```
rossberg, line 846:
```suggestion
 * Existing parameter types may be changed to to a _supertype_(!).
 * Existing result types may be changed to a subtype.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+Supertypes::
+
+Dually, the following modifications to a function type change it to a supertype:
++
+ * The result type list may be shortened
+ * The result type list may be extended with optional arguments (type `opt …`)
+ * The parameter type list may be extended.
+ * Existing parameter types may be change to to a subtype.
+ * Existing result types may be changed to a supertype(!).
```
rossberg, line 856:
```suggestion
 * Existing parameter types may be change to to a subtype(!).
 * Existing result types may be changed to a supertype.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
```
rossberg, line 857: Should th
```suggestion

TODO: add a remark on contravariance?
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+Corresponding Motoko type::
+
+Candid function types correspond to `shared` Motoko functions, with the result type wrapped in `async`. Arguments resp. results become tuples, unless there is exactly one, in which case it is used directly:
```
rossberg, line 860: Oneway functions?

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+Corresponding Rust type::
+
+TODO
+
+Corresponding JavaScript values::
+
+TODO
+
+[#type-null]
+=== Type null
```
rossberg, line 974: Maybe move this further up (e.g., next to opt?), since it's referenced in various previous sections.

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+[#candid-service-descriptions]
+=== Candid service descriptions
+
+With these candid types, we now can describe a service. Such a service description file can be written by hand, but typically it is generated from the service implementation, as described later.
+
+The simplest service description is
+....
+service : {}
+....
+and describes a service with no public methods. This is not very useful, so lets add a simple method:
```
crusso, line 80:
```suggestion
and describes a service with no public methods. This is not very useful, so let's add a simple method:
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+The method `reverse` expects a single parameter of type `text` and returns one value of type `text`.
+
+The method `divMod` expects and returns two values, all of type `int`. In its signature, we use the ability to name the parameters and result values.
+
+[CAUTION]
+====
+Naming the parameters or results of a method is purely for documentation purposes, but does not actually change the method’s type, or the values being passed. Parameter and results are identified by their _position_, independent of the name.
+
+In particular, Candid will does not prevent you from changing the type to
```
crusso, line 105:
```suggestion
In particular, Candid does not prevent you from changing the type to
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+....
+  divMod : (dividend : nat, divisor : nat) -> (mod : nat, div : nat);
+....
+or passing the above `divMod` to a service expeting a method that returns `mod` first.
```
crusso, line 109:
```suggestion
or passing the above `divMod` to a service expecting a method that returns `mod` first.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+You can also avoid this problem if you use the idiom where function with possibly changing arguments take a single record, i.e.
+....
+service { foo : (record { first_name : text; middle_name : text; last_name : text}) -> () }
+....
+Now, changing that to
+....
+service { foo : (record { first_name : text; last_name : text}) -> () }
+....
+does the right thing, and you don’t even need to keep a record of the removed argument around.
+
```
rossberg, line 1058:
```suggestion

Note: In general, it is not recommended to remove arguments from methods. Usually it is preferable to introduce a new method that omits the argument.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+`reserved`
+
+Textual syntax::
+
+`reserved`
+
+Subtypes::
+
+Any type
```
rossberg, line 1018:
```suggestion
All types
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+`empty`
+
+Textual syntax::
+
+None, as this type has no values
+
+Supertypes::
+
+Any type
```
rossberg, line 1076:
```suggestion
All types
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+Corresponding Rust type::
+
+TODO
+
+Corresponding JavaScript values::
+
+TODO
+
+Practical uses of the `empty` type are relatively rare.
```
rossberg, line 1090: One use case is for expressing an (output) record field that is no longer needed, by upgrading
```
record {foo : nat}  -->  record {foo : empty}
```

rossberg (commented):
Very nice, thanks for doing this!

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+  zip_code : nat;
+  country : text;
+};
+service address_book : {
+  set_address: (name : text, address) -> ();
+  get_address: (name : text) -> (opt address) query;
+}
+....
+
+These type definitions merely name an _existing_ type, they do not define a new type. It does not matter whether you use `address` in the function signature, or write out the records. In orther words, Candid uses _structural_ typing.
```
crusso, line 128:
```suggestion
These type definitions merely abbreviate an _existing_ type, they do not define a new type. It does not matter whether you use `address` in the function signature, or write out the records. In other words, Candid uses _structural_ typing.
```
Maybe say more here. Maybe "Two abbreviations with different names but equivalent definitions, describe the same type and are interchangeable."

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+In the last example, we also see the use of the `query` annotation. This Internet-Computer specific annotation indicates that the method `get_address` can be invoked using a query call (TODO: reference general documentation about query calls here)
+
+=== Encoding and decoding
+
+The point of Candid is to allow seamless invocation of service methods, passing arguments encoded to a binary format and transferred by an underlying transporation method (such as messages into or within the Internet Computer), and decoded on the other side.
```
crusso, line 134:
```suggestion
The point of Candid is to allow seamless invocation of service methods, passing arguments encoded to a binary format and transferred by an underlying transportation method (such as messages into or within the Internet Computer), and decoded on the other side.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+As a Candid user, you do not have to worry about the details of this binary format. If you plan to _implemnet_ Candid yourself (e.g. for a new host language), you can consult the Candid specification for details. However, some aspects of the format are worth knowing:
```
crusso, line 136:
```suggestion
As a Candid user, you do not have to worry about the details of this binary format. If you plan to _implement_ Candid yourself (e.g. for a new host language), you can consult the Candid specification for details. However, some aspects of the format are worth knowing:
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+* The Candid binary format starts with `DIDL…` (or, in hex, `4449444c…`). If you see this in some low-level log output, you are very likely observing a Candid-encoded value.
+
+* The Candid binary format always encodes _sequences_ of values, because methods parameters and results are sequences of types.
+
+* The binary format is quite compact. A `(vec nat64)` with 125000 entries takes 1000007 bytes.
+
+* The binary is self-describing, and includes a (condended) description of type of the values therein. This allows the receiving side to detect if a message was sent at a different, incompatible type.
```
crusso, line 144:
```suggestion
* The binary is self-describing, and includes a (condensed) description of type of the values therein. This allows the receiving side to detect if a message was sent at a different, incompatible type.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+* As long as the sending side serializes the arguments at the same type that the receiving side expects them, deserialization will succeed.
+
+[#upgrades]
+=== Service upgrades
+
+Services evolve over time: They gain new methods, existing methods return more data, or expect additional arguments. Usually, service authors want to do that without breaking existing clients.
```
crusso, line 151:
```suggestion
Services evolve over time: They gain new methods, existing methods return more data, or expect additional arguments. Usually, service authors want to do that without breaking existing clients.
```
or expect additional arguments - is exactly what we shouldn't be doing ;->

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+Candid supports such evolution by defining precise rules that indicate when the new service type will still be able to communicate with all other parties that are using the previous interface description. The underlying formalim is that of _subtyping_.
```
crusso, line 153:
```suggestion
Candid supports such evolution by defining precise rules that indicate when the new service type will still be able to communicate with all other parties that are using the previous interface description. The underlying formalism is that of _subtyping_.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+Services can safely evolve in the following ways:
+
+ * New methods can be added.
+ * Existing methods can return additional values, i.e. the sequence of result types can be extended. Old clients will simply ignore additional values.
+ * Existing methods can shorten their parameter list. Old clients may still send the extra arguments, but they will be ignored.
+ * Existing methods can extend their parameter list with optional arguments (type `opt …`). When reading messages from old clients, who do not pass that argument, a `null` values is assumed.
+ * Existing parameter types may be _changed_, but only to a _supertype_ of the previous type.
+ * Existing result types may be _changed_, but only to a _subtype_ of the previous type.
+
+See the reference sections the supertypes and subtypes of a given type.
```
crusso, line 164:
```suggestion
See the reference sections that define the supertypes and subtypes of a given type.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+Services evolve over time: They gain new methods, existing methods return more data, or expect additional arguments. Usually, service authors want to do that without breaking existing clients.
+
+Candid supports such evolution by defining precise rules that indicate when the new service type will still be able to communicate with all other parties that are using the previous interface description. The underlying formalim is that of _subtyping_.
+
+Services can safely evolve in the following ways:
+
+ * New methods can be added.
+ * Existing methods can return additional values, i.e. the sequence of result types can be extended. Old clients will simply ignore additional values.
+ * Existing methods can shorten their parameter list. Old clients may still send the extra arguments, but they will be ignored.
+ * Existing methods can extend their parameter list with optional arguments (type `opt …`). When reading messages from old clients, who do not pass that argument, a `null` values is assumed.
```
crusso, line 160: Is this safe yet in the higher-order setting? I've lost track of what implemented/intended.

docs/modules/candid-guide/index.adoc
nomeata:
```suggestion

NOTE: In general, it is not recommended to remove arguments from methods. Usually it is preferable to introduce a new method that omits the argument.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+  get : () -> (nat, last_change : timestamp) query;
+  subscribe : (func (nat) -> (unregister : opt bool)) -> ();
+}
+....
+====
+
+[#textual-values]
+=== Candid textual values
+
+The main purpose of Candid is to connect programs written in some host language (Motoko, Rust, JavaScript, etc.), and developers therefore usually do not have to deal with “Candid values” per se: They work with, say, JavaScript values that are then, transparently and seamelessly, transported to a canister written in Rust, where they show up as Rust values.
```
crusso, line 193:
```suggestion
The main purpose of Candid is to connect programs written in some host language (Motoko, Rust, JavaScript, etc.), and developers therefore usually do not have to deal with “Candid values” per se: They work with, say, JavaScript values that are then, transparently and seamlessly, transported to a canister written in Rust, where they show up as Rust values.
```

docs/modules/candid-guide/index.adoc
nomeata: That doesn’t work, right? Because now the whole `record` is empty. This way you can mark methods as “I will never return any more”, but then you just use `empty` instead of `record { foo : empty }`.

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+Nevertheless there are use cases where it is helpful to see the Candid values per se, e.g. for logging, debugging, or command-line interaction with services. In this cases, the _textual presentation_ for Candid values can be used.
```
crusso, line 195:
```suggestion
Nevertheless there are use cases where it is helpful to see the Candid values per se, e.g. for logging, debugging, or command-line interaction with services. In these scenarios, the _textual presentation_ for Candid values can be used.
```

docs/modules/candid-guide/index.adoc
nomeata:
```suggestion
These type definitions merely abbreviate an _existing_ type, they do not define a new type. It does not matter whether you use `address` in the function signature, or write out the records. Also, two abbreviations with different names but equivalent definitions, describe the same type and are interchangeable. In other words, Candid uses _structural_ typing.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+  public query func get() : async Int { v };
+  public func subscribe(handler : func (Int) -> async ()) { … }
+}
+....
+and it will have the interface shown above.
+
+In other languages, like Rust, you can still develop your service while only
+touching Rust types, but for now you still have to write the Candid interface description manually.
+
+In either case it is helpful to know the mapping between host language types and Candid types; this mapping is describeed in the <<reference, reference section>>.
```
crusso, line 236:
```suggestion
In either case it is helpful to know the mapping between host language types and Candid types; this mapping is described in the <<reference, reference section>>.
```

docs/modules/candid-guide/index.adoc
nomeata: No, still not fully sound (see #141). But the intention is sound, we know of one possible fix (subtyping in the decoder)

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+=== Candid tools and libraries
+
+The Motoko language has built-in support for Candid.
```
crusso:
```suggestion
The Motoko language has built-in support for Candid. A Motoko program can import a service as a Motoko actor, just by consuming the Candid of that service. A Motoko actor can implement some Candid service by exporting a Candid description of its Motoko interface.
```

docs/modules/candid-guide/index.adoc
nomeata: You could use it to declare that you stopped producing a certain _variant_ tag, but then you could just remove that variant…

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+....
+This can evolve to the following interface:
+....
+type timestamp = nat;
+service counter : {
+  set : (nat) -> ();
+  add : (int) -> (new_val : nat);
+  subtract : (nat, trap_on_underflow : opt bool) -> (new_val : nat);
+  get : () -> (nat, last_change : timestamp) query;
+  subscribe : (func (nat) -> (unregister : opt bool)) -> ();
```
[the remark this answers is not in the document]
nomeata: Hmm, I’d like to keep too much scary terms out of the intro, and hope this is covered by “Existing parameter types may be _changed_, but only to a _supertype_ of the previous type.  See the reference sections that define the supertypes and subtypes of a given type.”

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+dfx canister call msgt2-6nlzx-xq add_entry ("John Doe", record { street = "Hauptstraße", city = "Zürich" })
+....
+
+See TODO to read more about how to use the `dfx` tool.
+
+=== Interact with canisters via Candid UI?
+
+=== Use Candid as a Motoko canister developer?
+
+If you are writing a service from scratch in Motoko, then you need to take no special actions: The Motoko compiler will translate the signature of your Canister’s top-level actor or actor class into a Candid description, and the `dfx` build tool will make sure it ends up where it needs to be.
```
crusso, line 271:
```suggestion
If you are writing a service from scratch in Motoko, then you need to take no special action: The Motoko compiler will translate the signature of your Canister’s top-level actor or actor class into a Candid description, and the `dfx` build tool will make sure it ends up where it needs to be.

For example, in a multi-canister project, dfx will ensure that any clients of your Motoko service will import the generated Candid description of that service.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+:icons: font
+
+Welcome to the Candid’s user guide. This document will explain what Candid is, what it can do for you, and how you use it.
+
+NOTE: The primary target audience is Candid _users_, i.e. canister and front-end developers.
+If you are about to _implement_ Candid, want to learn about its internals or its meta-theory, you should look at the https://github.com/dfinity/candid/blob/master/spec/Candid.md[Candid specification].
+
+== What is Candid?
+
+Candid is an _Interface description language_, and is used to describe the public interface of a service (such a caniter running on the Internet Computer). It is language agnostic, and thus allows seamless interoperation between services and front-ends written different programming languages (Motoko, Rust, JavaScript, etc.)
```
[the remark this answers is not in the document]
nomeata: Well, there definitely is Candid support for JS. Only on the Frontend, but that’s why I write “services and front-ends”

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+This section lists all the types supported by Candid, including their type syntax, the syntax for the textual representation, their upgrading rules and the corresponding types in Rust, Motoko and Javascript.
+
+TIP: Subtypes are the types you can change your method _results_ to, and supertypes are the types that you can change your method _arguments_ to.
+
+We only list the particular subtypes and supertypes in each section, and do not repeat every time that that the type `empty` is a subtype of any type, and that the types `reserved` and `opt t` are supertypes of any types. See the sections on these types for more details.
+
+[#type-text]
+=== Type text
+
+The `text` type is used for human readable text. More precisely, its values are sequences of unicode code points (excluding surrogate parts).
```
crusso, line 304:
```suggestion
The `text` type is used for human readable text. More precisely, its values are sequences of unicode code points (excluding surrogate parts).
```
is that parts or pairs?

docs/modules/candid-guide/index.adoc
nomeata: I’d rather add that, with actual instructions, to “How do I use Candid as a Motoko canister developer?”. Although unclear where to draw the line between the candid manual and the Motoko manual.

docs/modules/candid-guide/index.adoc
nomeata: it’s “part” in https://github.com/dfinity/candid/blob/master/spec/Candid.md – I guess even a part is forbitten, let alone two that form a pair

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+`String`
+
+Corresponding JavaScript values::
+
+TODO:
+
+[#type-blob]
+=== Type blob
+
+The `blob` type can be used for binary data, i.e. sequences of bytes. It is interchangable with `vec nat8`, i.e. interfaces written in terms of `blob` and interfaces written in terms of `vec nat8` are compatible.
```
crusso, line 335: Is that really true? I thought we had a dedicated Blob candid type now. Does that mean we still don't round-trip blob from motoko?

nomeata: The Candid specification treats it like a short-hand, but it has a strong focus on encoding and subtyping rules. There they are indeed equivalent. The type syntax and the type mappings to the languages distinguishes them, though, so I chose to present it as a separate type (still related to `vec nat8` by subtyping requivalence).

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+`nat8`, `nat16`, `nat32`, `nat64`, `int8`, `int16`, `int32` or `nat64`
+
+Textual syntax::
+
+Same as `nat` (for `natN`) resp. `int` (for `intN`).
+
+Corresponding Motoko type::
+
+`natN` translate by default to `NatN`, but can also correspond got `WordN`
```
crusso, line 460:
```suggestion
`natN` translates by default to `NatN`, but can also correspond to `WordN` when required.
```

docs/modules/candid-guide/index.adoc
crusso: Maybe we should fix that...

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+TODO
+
+Corresponding JavaScript values::
+
+TODO
+
+[#type-vec]
+=== Type vec t
+
+The `vec` type representes vectors (sequences, lists, arrays): A value of type `vec t` contains zero or more values of type `t` in a specific order
```
crusso, line 569:
```suggestion
The `vec` type represents vectors (sequences, lists, arrays): A value of type `vec t` contains a sequence of zero or more values of type `t`.
```

docs/modules/candid-guide/index.adoc
nomeata: Isn’t it nice how writing documentation really makes you feel the pain of certain design choices? Almost as good as actually using the stuff

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+`Vec<T>`, where the Rust type `T` corresponds to `t`
+
+Corresponding JavaScript values::
+
+TODO:
+
+[#type-opt]
+=== Type opt t
+
+The `opt t` type contains of the values of type `t`, plus the special `null` value. It is used to express that some value is optional, i.e. it may be present, or there may be no value.
```
crusso, line 607:
```suggestion
The `opt t` type contains all the values of type `t`, plus the special `null` value. It is used to express that some value is optional, i.e. it may be present as some value of type `t`, or absent as the value `null`.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+The `opt` type can be nested (e.g. `opt opt text`), and the values `null` and `opt null` are distinct values.
+
+The `opt` type plays a crucial role in the evolution of Candid interfaces, and has atypical subtyping rules; see below.
```
crusso, line 611:
```suggestion
The `opt` type plays a crucial role in the evolution of Candid interfaces, and has special subtyping rules; see below.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+Subtypes::
+
+The canonical rules for subtyping with `opt` are:
++
+* Whenever `t` is a subtype of `t'`, then `opt t` is a subtype of `opt t'`.
+* `null` is a subtype of `opt t'`.
+* `t` is a subtype of `opt t` (unless `t` itself is `null`, `opt …` or `reserved`)
++
+In addition, for technical reasons related to upgrading and higher-order services, _every_ type is a subtype of `opt t`. Users are advised, however, to not directly make use of that rule.
```
crusso: weasel words

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+....
+type address = record {
+  street : text;
+  city : text;
+  zip_code : nat;
+  country : text;
+};
+....
+gives the name `address` to the type of records that have textual fields `street`, `city` and `country` and a numerical field `zip_code`. The order of fields in the record type declaration does not matter.
+
```
crusso, line 666:
```suggestion
Each field can have a different type (unlike vectors).
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+type address2 = record {
+  288167939 : text;
+  1103114667 : text;
+  220614283 : nat;
+  492419670 : text;
+};
+....
+In fact, textual labels are treated as their _field hash_, and incidentially, `address` and `address2` are – to Candid – the same types.
+
+If you omit the label, Candid numbers automatically assigns sequentially increasing labels. This leads to the following short syntax, which is typically used to represent pairs and tuples. The type `record { text; text; opt bool }` is equivalent to `record { 0 : text;  1: text;  2: opt bool }`
```
crusso, line 678:
```suggestion
If you omit the label, Candid automatically assigns sequentially increasing labels. This leads to the following short syntax, which is typically used to represent pairs and tuples. The type `record { text; text; opt bool }` is equivalent to `record { 0 : text;  1: text;  2: opt bool }`
```
Does this only kick in if you omit all labels? or does it start numbering from the previous label? Does that even make sense for record values where you'd expect to be able to provide labeled values in any order?

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+....
+record {}
+record { first_name = "John"; second_name = "Doe" }
+record { "name with spaces" = 42; "unicode, too: ☃" = true }
+record { "a"; "tuple"; null }
+....
+
+Subtypes::
+
+Subtypes of a record are record types that have additional fields (of any type), where some field’s types are changed to subtypes, or where optional fields are removed. It is, however, bad practice to remove optional fields in method results; you can change their type to `opt empty` to indicate that this field is no longer used.
```
crusso, line 700:
```suggestion
Subtypes of a record are record types that have additional fields (of any type), where some field’s types are changed to subtypes, or where optional fields are removed. It is, however, bad practice to remove optional fields in method results; you can change a field's type to `opt empty` to indicate that this field is no longer used.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
++
+For example, if you have a function returning a record of type
++
+....
+record { first_name : text; second_name : text; score : int }
```
crusso, line 705: Maybe add `nick_name: opt text` then deprecate?

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+....
+record { first_name : text; second_name : text; score : nat; country : text }
+....
+
+
+Supertypes::
+
+Supertypes of a record are record types with any some fields removed, some fields’ types changed to supertypes, or with optional fields added.
++
+The latter is what allows you to extend your argument records with additional arguments. Clients using the old interface will not include the field in their record, which will decode as `null`.
```
crusso:
```suggestion
The latter is what allows you to extend your argument records with additional arguments. Clients using the old interface will not include the field in their record, which will decode, when expected in the upgraded service, as `null`.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+A `variant` type represents a value that is from exactly one of the given cases, or _tags_. So a value of the type
+....
+type shape = variant {
+  dot : null;
+  circle : record { radius : float64 };
+  rectangle : record { width : float64; height : float64 };
+  "💬" : text;
+};
+....
+is either a dot, or a circle (with a radius), or a rectangle (with dimensions) or a speech bubble (with a text, and showing off unicode label names).
```
crusso, line 762:
```suggestion
is either a dot, or a circle (with a radius), or a rectangle (with dimensions) or a speech bubble (with some text). The speech bubble illustrates use of a unicode label name (💬).
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+
+The lables in variants are, just like the labels in records, actually numbers, and string labels refer to their hash value.
```
crusso:
```suggestion
The labels in variants are, just like the labels in records, actually numbers, and string labels refer to their hash value.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+....
+type Shape = {
+  #dot : ();
+  #circle : { radius : Float };
+  #rectangle : { width : Float; height : Float };
+  #_2669435721_ : Text;
+};
+....
++
+Note that if the type of a tag is `null`, this corresponds to `()` in Motoko, to preserve the mapping between the respecitve idiomatic ways to model enumerations as variants.
```
crusso, line 831:
```suggestion
Note that if the type of a tag is `null`, this corresponds to `()` in Motoko, to preserve the mapping between the respective idiomatic ways to model enumerations as variants.
```

crusso: Is it a problem that
```
variant {id ;} ==  variant { id : null } <: variant {id : ? T}
```` in Candid, 
but  
````
{# id} == {#id ()}  <|= {#id ? T}
````
in Motoko?

File: docs/modules/candid-guide/index.adoc [hunk 47]
```diff
@@ -0,0 +1,1098 @@
...
+
+Corresponding JavaScript values::
+
+TODO
+
+
+[#type-func]
+=== Type func (…) -> (…)
+
+Candid is designed to support higher-order use cases, where a service may receive or provide references to other serivces and their methods. The `func` type lies at the code or this: It indicates the function’s _signature_ (argument and results types, annotations), and values of this type are references to functions with that signature.
```
crusso:
```suggestion
Candid is designed to support higher-order use cases, where a service may receive or provide references to other services and their methods. The `func` type lies at the code or this: It indicates the function’s _signature_ (argument and results types, annotations), and values of this type are references to functions with that signature.
```

docs/modules/candid-guide/index.adoc
nomeata: It starts numbering from the prevoius label, but I made the opinionated choice to _not_ explain that in the user’s manual (I don't really see a good use case for it). The specification has all the answers…

docs/modules/candid-guide/index.adoc
crusso:
```suggestion
* `query` indicates that the referenced function is a query method, meaning it does not alter the state of its canister, and that it can be invoked using the cheaper “query call” mechanism.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+....
+
+Subtypes::
+
+The following modifications to a function type change it to a subtype (cf. the rules in <<upgrades>>):
++
+ * The result type list may be extended
+ * The parameter type list may be shortened.
+ * The parameter type list may be extended with optional arguments (type `opt …`).
+ * Existing parameter types may be change to to a supertype.
```
crusso:
```suggestion
 * Existing parameter types may be change to a supertype.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+ * Existing result types may be _changed_ to a subtype(!).
+
+Supertypes::
+
+Dually, the following modifications to a function type change it to a supertype:
++
+ * The result type list may be shortened
+ * The result type list may be extended with optional arguments (type `opt …`)
+ * The parameter type list may be extended.
+ * Existing parameter types may be change to to a subtype.
```
crusso:
```suggestion
 * Existing parameter types may be change to a subtype.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
++
+....
+service "w7x7r-cok77-xa"
+service "zwigo-aiaaa-aaaaa-qaa3a-cai"
+service "aaaaa-aa"
+....
+
+Subtypes::
+
+The subtypes of a service type are those service types that possibly have additional method, and where the type of existing methods is changed to a subtype.
```
crusso, line 950:
```suggestion
The subtypes of a service type are those service types that possibly have additional methods, and where the type of an existing method is changed to a subtype.
```

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+TODO
+
+Corresponding JavaScript values::
+
+TODO
+
+[#type-principal]
+=== Type principal
+
+The Internet computer uses _principals_ as the common scheme to idnetify canisters, users and other entities.
```
crusso, line 982:
```suggestion
The Internet computer uses _principals_ as the common scheme to identify canisters, users and other entities.
```

docs/modules/candid-guide/index.adoc
nomeata:
```suggestion
Candid is designed to support higher-order use cases, where a service may receive or provide references to other services and their methods. The `func` type lies at the heart of this: It indicates the function’s _signature_ (argument and results types, annotations), and values of this type are references to functions with that signature.
```

nomeata: I meant “core”, but Andreas already fixed that one.

docs/modules/candid-guide/index.adoc
```diff
@@ -0,0 +1,1098 @@
...
+it would be disastrous: If a client talks to you using the old interface, you will silently ignore the `last_name` and take the `middle_name` as the `last_name`. Remember that method parameter names are just convention, and method arguments are identified by their position.
+
+Instead, you can use
+....
+service { foo : (first_name : text, middle_name : reserved, last_name : text) -> () }
+....
+to indicate that `foo` used to take a second argument, but you no longer care about that.
+
+
+You can also avoid this problem if you use the idiom where function with possibly changing arguments take a single record, i.e.
```
crusso, line 1056:
```suggestion
You can avoid this pitfall by adopting the pattern any function that is anticipated to have changing arguments, or whose arguments can only be distinguished by position, not type,  is declared to take a single record, i.e.
```

crusso (approved):
Looks good to me. Would be great if the SDK folks could fill in the JavaScript examples so our users know how handle everything, bit in particular, Candid bignums, options and variants, whose JS mappings may not be obvious.

nomeata:
> Looks good to me. Would be great if the SDK folks could fill in the JavaScript examples so our users know how handle everything, bit in particular, Candid bignums, options and variants, whose JS mappings may not be obvious.

Definitely! But maybe in separate PRs againts `master`, so that it’s off my plate :-)

docs/modules/candid-guide/index.adoc
crusso: So I guess my question is if I export a motoko service producing a Blob and then import the service, do I get `[nat8]` or `Blob` in the client? That's confused our users with NatX/WordsX and will likely confuse them again...

docs/modules/candid-guide/index.adoc
crusso: Yeah, highly recommend both writing doc and dog-fooding...

docs/modules/candid-guide/index.adoc
crusso: https://en.wikipedia.org/wiki/The_road_to_hell_is_paved_with_good_intentions

docs/modules/candid-guide/index.adoc
rossberg: Ah yes, sorry, I was thinking about the case where you have an option, should have been `...opt nat --> ...opt empty`. But admittedly that is a narrow use case.

docs/modules/candid-guide/index.adoc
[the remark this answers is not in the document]
rossberg: Oh, totally fine with avoiding the term, but thought that some explanation might be in order (since the case doesn't match the earlier explanations).

docs/modules/candid-guide/index.adoc
nomeata: I think you get `Blob`.

nomeata:
Why can’t I merge this? It says “All checks have passed” but also “Required statuses must pass before merging”?

PR #160, opened 2021-01-11
Candid test suite: More tests related to references

nomeata:
I think we didn’t have any that asserted that a decoder does not do a
subtyping check, which means that any service (func) reference decodes
at any expected service (func) type.


chenyan-dfinity:
Right, I intentionally omitted these tests, as it will change in the future?

nomeata:
I am not sure it will. And even if something changes, I think it’s advisible to clearly specify (also with tests) the status quo, so that changes to this specifciation are loud and noisy, and noticed by all implementations.

chenyan-dfinity (approved):

PR #161, opened 2021-01-12
candid asts as candid data

matthewhammer:
- [x] derive equality and cloning for ASTs of candid types (type `IDLType` and related types)
- [x] derive (binary) deserializer for ASTs of candid types
- [ ] derive (binary) serializer for ASTs of candid types --- had trouble using the directive, because of a crate-path issue?

#### Back story

Note that this is asking for candid asts in Rust to be usable as candid data, not as a plaintext string.  Plaintext forms for candid types are supported in Rust, but not in every language (not in Motoko).  By converting these ASTs, as data, into candid binary form, we can send them to canisters that act on them as ASTs (as data, not as strings) and get these ASTs back without re-doing the pretty printer for candid types.

### DIFF
--- a/rust/candid/src/parser/types.rs
+++ b/rust/candid/src/parser/types.rs
@@ -1,22 +1,33 @@
 use crate::types::Label;
 use crate::Result;
 use pretty::RcDoc;
+use crate::{CandidType, Deserialize};
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Deserialize, Eq, PartialEq, Hash)]
 pub enum IDLType {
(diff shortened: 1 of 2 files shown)


rust/candid/src/parser/types.rs
```diff
@@ -1,22 +1,33 @@
 use crate::types::Label;
 use crate::Result;
 use pretty::RcDoc;
+use crate::{CandidType, Deserialize};
```
matthewhammer, line 4: @chenyan-dfinity I tried to use `CandidType` below each place where I added `Deserialize`, but I got some macro errors.  I think the issue is about the macro being used from the crate that defines it?

PR #166, opened 2021-01-19
generate random candid values

chenyan-dfinity:
Design: https://github.com/dfinity/candid/blob/random/tools/didc/config.md

Fix https://github.com/dfinity/dx-triage/issues/41


nomeata (commented):
Nice!

Generating good random values for recursive data structures is a bit tricky. E.g. if your top-level type is an `opt`, you’d get boring `null` half the time.

What you could do is to first make a pass over the types and estimate the “approximate size“ of the type: 0 for empty, 2 for bool; maybe cap at, dunno, 100 for “many or infinite”.

Then in cases like `opt` or `variant`, you can draw weighted by that approximate size.

This way, you get better exploration of `opt` and variants. And you can handle `variant { foo : empty; bar : bool }` without crashes.

rossberg:
FWIW, Hughes et al. recently wrote a whole journal article about this problem. Public version can be found [here](https://arxiv.org/abs/1808.01520). It's a good read.

rust/candid/src/parser/random.rs
```diff
@@ -216,3 +121,75 @@ impl IDLValue {
...
+            Empty => 0,
+            Opt(t) => 1 + self.size_helper(seen, t)?,
+            Vec(t) => 1 + self.size_helper(seen, t)? * 2,
+            Record(fs) => {
+                let mut sum = 0;
+                for Field { ty, .. } in fs.iter() {
+                    sum += self.size_helper(seen, ty)?;
+                }
+                1 + sum
+            }
```
nomeata, line 178: Technically, the size of a product is, well, the product of the sizes (not the sum). And then, if you want to be correct, you have to make sure that the size of `{empty; int}` comes out as  0, not ∞ (i.e. `0 * ∞ = ∞`, which is not the semantics you get with Rust’s `?` operator here).

rust/candid/src/parser/random.rs
```diff
@@ -216,3 +121,75 @@ impl IDLValue {
...
+        use Type::*;
+        Some(match t {
+            Var(id) => {
+                if seen.insert(id.to_string()) {
+                    let ty = self.rec_find_type(id).unwrap();
+                    self.size_helper(seen, &ty)?
+                } else {
+                    return None;
+                }
+            }
```
nomeata, line 168: This returns ∞ whenever a type was seen before, but I _think_ you want to return ∞ whenever a type is recursive?
I.e.  `type a = opt bool; record { a; a }` is not infinite.

It seems you are using one mutable `seen` for the whole type, rather than an immutable that you modify only locallly as you recurse.

chenyan-dfinity: Thanks for catching this! Fixed.

rust/candid/src/parser/random.rs
chenyan-dfinity: It's computing the size of the `IDLValue` of type t, so sum is more appropriate. For empty record, it's a bit tricky, because `type A = record { A }` has ∞ value size, but it's actually 0.

nomeata: I don’t follow. Is this the size as in “number of elements” or “size of nodes in the type expressions”?

Surely `{opt bool; opt bool}` has _(1+2)*(1+2) = 9_ different values, not 6.

nomeata: Ah, indeed, size of `bool` is 1, not 2. Ignore me then, _thought_ this is the cardinality of the type (which I would expect to use as a weight when trying to pick elements at random) not the size of the type expression… but maybe using the type expressoin is also sensible.

chenyan-dfinity: I think all metrics make sense for picking variants: cardinality, depth, value size. But we also use this to estimate the length of a vector, value size feels more accurate.

matthewhammer: > I think all metrics make sense for picking variants: cardinality, depth, value size. But we also use this to estimate the length of a vector, value size feels more accurate.

I wonder, when is this (very cool) avenue going to fully devolve into fractal geometry, rather than merely flirt with it?

matthewhammer: E.g., We can also talk of "area" and "dimension" of the values that we get, a la fractal geometry.

tools/didc/src/main.rs
```diff
@@ -205,6 +213,26 @@ fn main() -> Result<()> {
...
         }
+        Command::Random { annotate, format } => {
+            use candid::parser::configs::Configs;
+            use rand::Rng;
+            use serde_dhall::SimpleValue;
+            let mut rng = rand::thread_rng();
+            let seed: Vec<u8> = (0..2048).map(|_| rng.gen::<u8>()).collect();
+            let mut u = arbitrary::Unstructured::new(&seed);
+            let (env, types) = annotate.get_types(Mode::Encode)?;
+            let config = serde_dhall::from_file("random.dhall").parse::<SimpleValue>()?;
```
matthewhammer, line 224: What's the intended workflow here?  (The hardcoded name "random.dhall" is meant to be blessed/special, or merely provisional in this PR?)

chenyan-dfinity: It's provisional. We can have a flag to read the dhall config and fall back to a default setting when not available.

PR #167, opened 2021-01-20
Improve wording for type mismatches

knl:
**Overview**
> Why do we need this feature? What are we trying to accomplish?

The original error message was creating confusion on several occasions, and this change tries to make it more obvious where the problem is coming from.

**Requirements**
> What requirements are necessary to consider this problem solved?

Engineer reading the error message does not second guess what was the observed and what was the expected type.

**Considered Solutions**
> What solutions were considered?

Only this one.

**Recommended Solution**
> What solution are you recommending? Why?

Most of the serialization protocols use word 'expected'.

**Considerations**
> What impact will this change have on security, performance, users (e.g. breaking changes) or the team (e.g. maintenance costs)?

This should reduce user confusion.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -321,7 +321,7 @@ impl<'de> Deserializer<'de> {
         let wire_type = self.parse_type()?;
         if wire_type != expected {
             return Err(Error::msg(format!(
-                "Type mismatch. Type on the wire: {:?}; Provided type: {:?}",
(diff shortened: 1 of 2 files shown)


nomeata (approved):
I agree that this is better from the user’s point of view.

luc-mercier (approved):
Thanks!

chenyan-dfinity (approved):
LGTM

PR #168, opened 2021-01-20
Spec: Do a subtyping check when decoding

### LINKED ISSUE #141: Still not sound: Gradual typing vs. opportunistic decoding

nomeata:
this introduces a subtyping check

 * before even starting to decode
 * when decoding `opt` values.

An implementation can probably pre-compute something at first, and then
at each `opt` occurrence quickly look which way to go…

With these changes, the coercion relation never fails on well-typed
inputs, and it is anyways right-unique.
I am wondering if it makes sense to rewrite it as a family of functions,
indexed by the input and output types (like Andreas had it originally).

This fixes #141, I hope.


spec/Candid.md
```diff
@@ -884,20 +884,22 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
 
 This subtyping is implemented during the deserialisation of Candid at an expected type. As described in [Section Deserialisation](#deserialisation), the binary value is conceptually first _decoded_ into an abstract value, and then _coerced_ to the expected type.
 
-This section describes the coercion, as a ternary relation `V ~> V' : T` to describe when a value `V` can be coerced to a value `V'` of type `T`. The fields `V` and `T` can be understood as inputs and `V'` as the output of this relation.
+In the following, we use `V` as an abstract data model of both the message in transit (`V`), as well as the result of the coercion (`V'`).  We re-use the syntax of the textual representation.
 
-Here `V` models untyped values, which form an abstract data model of both the message in transit (`V`), as well as the result of the coercion (`V'`). In the following, we re-use the syntax of the textual representation, with the following changes to make it free of overloading:
+Since messages are self-describing, i.e. carry type description, the rules below can refer to the given type of the message using the `<annval>` syntax (i.e. `(v : t)`) if the type matters. This is used in particular
```
rossberg, line 889:
```suggestion
Since messages are self-describing, i.e. carry a type description, the rules below can refer to the given type of the message using the `<annval>` syntax (i.e. `(v : t)`) if the type matters. This is used in particular
```

rossberg (approved):
Cool, thanks!

Yeah, at this point the functional specification would probably be preferable again -- more compact and easier to grok for non-PLT-initiates. But we can leave that for later.

nomeata:
I don’t think I’d mind rewriting it right away. OTOH, the current formulation does it’s job… we’ll see.

Are you fine with requiring a subtyping check ahead of time? This means that `vec {} : vec int` no longer decodes at type `vec bool`? (Probably a good thing)

rossberg:
Seems reasonable, but would the alternative of doing a check on each empty vector and on variants be equivalent?

nomeata:
> Seems reasonable, but would the alternative of doing a check on each empty vector and on variants be equivalent?

Then you’d also have to do it on func and service referenes. Essentially, everytime when a consitutent type of the type doesn't have a manifest value it referes to.

rossberg:
Right, but specification-wise, maybe that is a more natural way to formulate it? Hm, and even practically, perhaps it might save you from running any check in many cases.

nomeata:
I think I find it much easier to assume that the coercion function/relation only works on input that has already passed the subtyping check. And yes, better a single check than many (in the case of `[[], [],[]]` for example)

rossberg:
An implementation could of course cache the pairs of types already checked (which it wants to do anyway for recursion), so I think it should be less work rather than more. But I don't mind going the other way for the spec.

nomeata:
I wasn't sure if we’d still need to fix https://github.com/dfinity/candid/pull/151, but since that issue said “so maybe
rewriting this as a partial coercion function would be helpful” I just rewrote _this_ PR to make coercion a total function indexed by the actual and expected types. Seems simpler this way?

nomeata:
Direct link to the spec: https://github.com/dfinity/candid/blob/joachim/spec-subtype-on-decode/spec/Candid.md

chenyan-dfinity (approved):
LGTM. Do we want to bump the spec version as well?

nomeata:
I think a minor bump at least.

I am also abit torn about the role of the test data – I consider them part of the spec, but if we update them now we break the rust CI (it’s the downside of monorepos enforcing lockstep development of components that are trailing each other.) But I guess we can be pragmatic and update the test data when we update the rust implementation.

chenyan-dfinity:
> if we update them now we break the rust CI 

Ah, right. Maybe we can update the test in comments, and uncomment them when we are ready?

nomeata:
The problem is that writing these tests by hand, without _any_ implementation to cross-check them, will likely just cause them to be riddled with typos in the binary parts…

Let’s just merge this with a version bump, and maybe add a version comment to all tests?

nomeata:
@rossberg would you mind to re-review? I rewrote it with a coercion function indexed by `t1 <: t2`.

spec/Candid.md
```diff
@@ -882,201 +882,150 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
-
----------------------
-false ~> false : bool
-
------------------------
-<text> ~> <text> : text
-
--------------------
-null ~> null : null
+c[<t> <: <t>](x) = x    for every <t> ∈ <numtype>, bool, text, null
```
rossberg, line 895: For hygiene, I think it's safer to expand the def for each case, so that you can limit the argument value to the right shapes as well.

spec/Candid.md
```diff
@@ -882,201 +882,150 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 #### Variants
 
 Only a variant value with an expected tag coerces at variant type.
 
 ```
-<v> ~> <v'> : <t>
-----------------------------------------------------------------------------------
-variant { <nat> = <v> } ~> variant { <nat> = <v'> } : variant { <nat> = <t>; _;* }
+c[variant { <nat> = <t>;* _;* } <: variant { <nat> = <t'>;* _;* }](variant { <natN> = <v> })
+    = variant { <natN> = c[<tN> <: <t'N>](<v>) }
```
rossberg, line 958: This formulation is a bit unclear, and perhaps more complicated than necessary. I assume the first `*` are the common tags, whereas the `_;*` are additional ones? But then, why can the latter occur on the l.h.s.? And how does `natN` relate to `nat*`?
```suggestion
c[variant { <nat> = <t>; _;* } <: variant { <nat> = <t'>; _;* }](variant { <nat> = <v> })
    = variant { <nat> = c[<t> <: <t'>](<v>) }
```

spec/Candid.md
```diff
@@ -882,201 +882,150 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
-null ~> null : null
+c[<t> <: <t>](x) = x    for every <t> ∈ <numtype>, bool, text, null
 ```
 
-Values of type `nat` coerce successfully at type `int`:
+Values of type `nat` coerce at type `int`:
 ```
-------------------------
-(<x> : nat) ~> <x> : int
+c[nat <: int](x) = x
```
rossberg, line 900:
```suggestion
c[nat <: int](<nat>) = <nat>
```
(to restrict to actual nat values)

spec/Candid.md
```diff
@@ -882,201 +882,150 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 #### Tuple types
 
 Whole argument and result sequences are coerced with the same rules as tuple-like records. In particular, extra arguments are ignored, and optional parameters read as as `null` if the argument is missing or fails to coerce:
 
 ```
-record {<v>;*} ~> record {<v'>,*} : record {<t>;*}
---------------------------------------------------
-(<v>,*) ~> (<v'>,*) : (<t>,*)
+c[(<t>,*) <: (<t'>,*)]((<v>,*)) = (<v'>,*)
+    if c[record {<t>;*} ~> record {<t'>,*}](record {<v>;*}) = record {<v'>;*}
```
rossberg, line 978: Left-over notation?

spec/Candid.md
```diff
@@ -882,201 +882,150 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
-  T <: T' ⇒ ∀ v : T. ∃ v'. v ~> v' : T
+  <t> <: <t'> ⇒ ∀ <v> : <t>. v[<t> <: <t'>](<v>) : <t'>
   ```
 
 * Higher-order soundness of subtyping
   See <./IDL-Soundness.md>, with the following instantiations:
   ```
   s1 ~> s2 ≔ s2 <: s1
-  t1 <: t2 ≔ (∀ v. v : t1 ⇒ v ~> _ : t2 )
+  t1 <: t2 ≔ t1 <: t2
```
rossberg, line 1010: I suppose you could just drop this line?

spec/Candid.md
```diff
@@ -1302,8 +1251,8 @@ Deserialisation at an expected type sequence `(<t'>,*)` proceeds by
  * checking for the magic number `DIDL`
  * using the inverse of the `T` function to decode the type definitions `(<t>,*)`
  * check that `(<t>,*) <: (<t'>,*)`, else fail
- * using the inverse of the `M` function, indexed by `(<t>*)`, to decode the values `(<v>*)`
- * use the coercion relation `(<v>,*) ~> (<v'>,*) : (<t'>,*)` to understand the decoded values at the expected type.
+ * using the inverse of the `M` function, indexed by `(<t>*,)`, to decode the values `(<v>*,)`
```
rossberg, line 1254:
```suggestion
 * using the inverse of the `M` function, indexed by `(<t>,*)`, to decode the values `(<v>,*)`
```

spec/Candid.md
```diff
@@ -1302,8 +1251,8 @@ Deserialisation at an expected type sequence `(<t'>,*)` proceeds by
...
+ * use the coercion function `c[(<t>,*) <: (<t'>,*)]((<v>*,))` to understand the decoded values at the expected type.
```
rossberg, line 1255:
```suggestion
 * use the coercion function `c[(<t>,*) <: (<t'>,*)]((<v>,*))` to understand the decoded values at the expected type.
```

rossberg (commented):
Looks good modulo a few minor things.

I would like to suggest an alternative notation, however. Following the `R(v : t)` style that the doc uses for serialisation functions, we could write `c[t <: t'](v)` as `C(v : t <: t')`.

nomeata:
>  that the doc uses for serialisation functions, we could write `c[t <: t'](v)` as `C(v : t <: t')`.

I’d really like to have a notation that shows the stage distinction, i.e. “for every t and t', with `t <: t`' you get a function `t -> t'`”. Or, in dependently-typed phrased `c : forall t1 t2, t1 <: t2 -> t1 -> t2`.

This matters, because all the type-level stuff (including subtype checks) happend during computation of the value function; once we have partially applied `c[t <: t']`, there is no more type analysis needed.

spec/Candid.md
nomeata: Oh, yes, that's a bit bogus. Should be
```suggestion
c[variant { <nat> = <t>;* } <: variant { <nat> = <t'>;* _;* }](variant { <natN> = <v> })
    = variant { <natN> = c[<tN> <: <t'N>](<v>) }
```
where only the supertype has extra tags.

The `<natN>` is one of the `<nat>*`, I wasn’t sure how to best express that in this pseudo-syntax. `<nat>[i]` maybe?

spec/Candid.md
nomeata: `c[nat <: int]` is, by definition and construction, a function `nat -> int`, so is this need?

spec/Candid.md
nomeata: No; this is a way to say “this function does on tuples what it does on records”.

The altnerative would be to write functions `(<t>,*) -> record {<t>;*} ` and `record {<t>;*}  -> (<t>,*)` to explain the bijection; that seemd to be even less clear

spec/Candid.md
nomeata: Don’t think so, the left hand side is “parameters of IDL-Soundness”, the right hand side is “stuff defined here”. The punnig is mostly accidential.

spec/Candid.md
rossberg: I was referring to the `~>`.

spec/Candid.md
rossberg: That definition isn't necessarily self-evident here. And it would match the more explicit value patterns in other cases. But it's a small thing.

spec/Candid.md
rossberg: What's wrong with my simpler suggestion?

nomeata: Hmm, I guess I tried to keep a closer connection to the the subtyping rule, modeling that the supertype has more tags than the subtype… but you are right, it’s not needed here.

spec/Candid.md
nomeata: Ah, right!
```suggestion
    if c[record {<t>;*} <: record {<t'>,*}](record {<v>;*}) = record {<v'>;*}
```

chenyan-dfinity:
What's the algorithm for subtyping check for recursive types? Do we want subtyping to check for isomorphism as well, e.g. `\mu t. record { t } <: empty`. Amber rule doesn't seem to handle this case. If we use equi-recursive, we need to compute the fix point, which is quite expensive.

Also for memoization, the table size is quadratic to the number of subterms in the type. Storing this table during decoding requires a lot of memory, and the entries are rarely used.

nomeata:
> What's the algorithm for subtyping check for recursive types?

Not quite an algorithm, but if you interpret the infinite (coinductive) unfolding of the trees, then the inference rules given in the spec (understood as a coinductive predicate) should uniquely answer all quesitons about which subtypig relations hold.

> Do we want subtyping to check for isomorphism as well, e.g. `\mu t. record { t } <: empty`.

No, I don’t think so. No subtyping rule matches `record { … } <: empty`, so you can fail there already, no matter what the `…` is.

> Also for memoization, the table size is quadratic to the number of subterms in the type. Storing this table during decoding requires a lot of memory, and the entries are rarely used.

Depends on the representation: You need to remember a set of pairs. A full matrix (as you are hinting at) is one possible representation, but (as you say) maybe not a good one. Storing just the set of pairs might be better.

chenyan-dfinity:
> if you interpret the infinite (coinductive) unfolding of the trees, then the inference rules given in the spec (understood as a coinductive predicate) should uniquely answer all quesitons about which subtypig relations hold.

Let me rephrase this to see if I understand correctly: checking `var1 <: t` equals to checking `unfold(var1) <: t` assuming `var1 <: t`. Or we need to find the recursive variable `var2` in `t`, and check `unfold(var1) <: unfold(var2)` assuming `var2 <: var2`? I've never done any coinductive proofs, so it's a bit hard for me to reason about correctness.

> equivalence vs subtyping

I think my question is does equivalence implies subtyping? Is it possible that `unfold(t1) <: t2` but not `t1 <: t2`. We didn't formalize recursive types in the spec, so this question doesn't exist in the meta theory. But when we implement subtyping in practice, we are handling recursive types.

> Storing just the set of pairs might be better.

I think the space complexity is `O(n^2)`, where `n` is the size of type table? This still seems a lot to me, given a lot of deserialization algorithms aim for sublinear space.

nomeata:
> > if you interpret the infinite (coinductive) unfolding of the trees, then the inference rules given in the spec (understood as a coinductive predicate) should uniquely answer all quesitons about which subtypig relations hold.
> 
> Let me rephrase this to see if I understand correctly: checking `var1 <: t` equals to checking `unfold(var1) <: t` assuming `var1 <: t`.

Correct! The essence of a coinductive proof is that you are trying to see if anything _contradicts_ the subtyping to hold. For `µx. opt x <: µx.opt x`, you will just keep running in circles if you apply the rules. In an inductive setting, that would mean that the relation does not hold; in a coinductive settings it means that it does hold. That’s why one says that inductive relations describe the smallest fixpoint, and coinductive relations describe the largest.

But pragmatically, what you describe with the “assumes” suffices: You remember which relations are you checking, and consider them to hold even before you are done.

Note that if you bring both actual and expected type into the type table form that we use in the messages, every type _is_ a variable (or a type table index, same thing), so you can memoize pairs of these ids.

The `rel_type` function in `mo_types/type.ml` in Motoko is doing a very similar thing, you can have a peek there (although I find it’s use of mutable maps and their identity confusing).
 
> > equivalence vs subtyping
> 
> I think my question is does equivalence imply subtyping? Is it possible that `unfold(t1) <: t2` but not `t1 <: t2`. We didn't formalize recursive types in the spec, so this question doesn't exist in the meta theory. But when we implement subtyping in practice, we are handling recursive types.

Yes, equivalence is a subrelation of subtyping, and both respect the unfolding relation, i.e. if `T = t1` is a type alias, then `T <: t2'` if and only if `t1 <: t2`.

The meta-theory _does_ formalize recursive types, but only implicit, but _only_ talking about the (sometimes infinite) unfolding. It doesn’t tell how you how to implement things on finite representations though :-)

> > Storing just the set of pairs might be better.
> 
> I think the space complexity is `O(n^2)`, where `n` is the size of type table? This still seems a lot to me, given a lot of deserialization algorithms aim for sublinear space.

Yes, the complexity class is `O(n*m)` (where `n` is the size of the incoming type table, and `m` the size of the expected type’s table). We didn’t see a way around that :-/

chenyan-dfinity:
> equivalence is a subrelation of subtyping

`µx. record { x }` is equivalent to `empty`, but `µx. record { x }` is not a subtype of `empty`. Is this the only exception?

chenyan-dfinity:
> the complexity class is O(n*m)

We don't have to store the subtyping relation. The subtyping check takes `O(n+m)` time, where `n` and `m` are the size of the type table from the incoming type and the expected type. During decode, we can check for subtyping at the same time when we do coercion. The decoding still takes linear time with a local memo table that only contains recursive variables (in practice, this should be very small).

nomeata:
> > equivalence is a subrelation of subtyping
> 
> `µx. record { x }` is equivalent to `empty`, but `µx. record { x }` is not a subtype of `empty`. Is this the only exception?

What's your definition of equivalent here?

`variant {}` also has no values, but isn't a subtype of `empty` or `mu t. record t`.

nomeata:
> > the complexity class is O(n*m)
> 
> We don't have to store the subtyping relation. The subtyping check takes `O(n+m)` time, where `n` and `m` are the size of the type table from the incoming type and the expected type. During decode, we can check for subtyping at the same time when we do coercion. The decoding still takes linear time with a local memo table that only contains recursive variables (in practice, this should be very small).

I'd be surprised if this was true (with `O(n+m)`). The current state of your PR, although you only memorize pairs where one or the other type is a `Var` or `Knot`, is, I think, still quadratic. But maybe you can prove me wrong, I wouldn't mind :-)

nomeata:
Here I think the prototypical conter example. Let `m` and `n` be co-prime (no common divisor besides 1), and let
```
t1 = vec vec vec … vec t1 = vec^n t1
t2 = vec vec vec … vec t2 = vec^m t2
```
The type tables have sizes `n` and `m`. To do the check `t1 <: t2`, you’ll end up comparing each `vec` in `t1` with each vec in `t2`, so that’s in _O(n×m)_. But only time or also space? Less sure how to make the argument here, though. But we can look at these types to check concrete algorithms for size usage.

I know that `t1` and `t2` are actually both equal to `µ t. vec t`, and if both were canonicalized, the’d had a type table of size one. But we can easily fix that:
```
Ta[t] = rec { nat; t }
Tb[t] = rec { int; t }
Tc[t] = rec { reserved; t }
t1 = Ta^{n-1}[ Tb [t1] ]
t2 = Tb^{n-1}[ Tc [t1] ]
```
The `nat`, `int`, `reserved` merely make sure we can’t have smaller type tables for the `t1`, `t2`, but otherwise don't affect the subtyping relation.

Also note that trying to store _less_ in the memotable (e.g. have less memo-points) will likely just trade heap space for stack space. So when assessing space complexity, one should write the algorithm as a loop without stack.

Speaking of stack: It is rather straight-forward to implement this subtyping check in a loop without stack, using the standard work list approach that one uses on most graph algorithms: Keep a list or set of pairs of types (or type table indices, if both types are stored with type tables) that you still need to check. Until that is empty, pop one, add it to the “done” set, check that it makes sense locally, and then add all subchecks that are not done yet to the todo list.

rossberg:
> > equivalence is a subrelation of subtyping
> 
> `µx. record { x }` is equivalent to `empty`, but `µx. record { x }` is not a subtype of `empty`. Is this the only exception?

They are not equivalent, they are merely isomorphic. Those are very different notions, with different implications. The type system doesn't magically equate type isomorphisms, just like `Bool <> {#a; #b}`. Maybe one day somebody will invent homotopy type theory with full path inference. :)

As @nomeata says, the worst case (space and time) complexity is quadratic. In practical cases, it typically is linear, though.

To be honest, this complexity has been worrying me quite a bit. That really is Candid's Achilles heel. I wonder whether, in practice, we couldn't get away with some form of iso-recursive interpretation, which would always be linear.

chenyan-dfinity:
> Also note that trying to store less in the memotable (e.g. have less memo-points) will likely just trade heap space for stack space.

This is a good trade. Memo table is global and stack is bounded by the depth of the recursion. For the `vec^m`, `vec^n` example, only memoizing `Var` node, the size of the memo table is only `n + m`, as one of the tuple is always the single `Var` node, while the other type unfolds. Similarly, the recursion depth would only increase linearly instead of quadratically. Overall, I think space is still linear.

> I wonder whether, in practice, we couldn't get away with some form of iso-recursive interpretation, which would always be linear.

I hope by unfolding types along with the value structure, it is enough to rule out invalid types.

nomeata:
> > Also note that trying to store less in the memotable (e.g. have less memo-points) will likely just trade heap space for stack space.
> 
> This is a good trade. Memo table is global and stack is bounded by the depth of the recursion. For the `vec^m`, `vec^n` example, only memoizing `Var` node, the size of the memo table is only `n + m`, as one of the tuple is always the single `Var` node, while the other type unfolds. Similarly, the recursion depth would only increase linearly instead of quadratically. Overall, I think space is still linear.

isn’t there `n` vars, if you represent the type in A-normal form (as it is in the type table, where _every_ type is its own var)?

And I am certain I can construct an example where _every_ subterm has to be its own var.

rossberg:
Well, we don't need to guess, the results are well-known. See e.g. TAPL Ch. 21. Structural subtyping is quadratic (it is equivalent to deciding DFA inclusion) when you do memoization properly. Without it, it's exponential, as was the case for Amadio & Cardelli's original algorithm.

spec/Candid.md
```diff
@@ -882,200 +882,150 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 ```
-<v> ≠ null
-<v> ≠ (null : reserved)
-<v> ≠ opt _
-not (null <: <t>)
-opt <v> ~> <v'> : opt <t>
--------------------------
-<v> ~> <v'> : opt <t>
+C[<t> <: opt <t'>](<v>) = opt C[<t> <: <t'>](v)  if not (null <: <t>) and <t> <: <t'>
+C[<t> <: opt <t'>](_)   = null                   if not (null <: <t>) and not (<t> <: <t'>)
```
nomeata, line 935:
```suggestion
C[<t> <: opt <t'>](<v>) = opt C[<t> <: <t'>](v)  if not (null <: <t'>) and <t> <: <t'>
C[<t> <: opt <t'>](_)   = null                   if not (null <: <t'>) and not (<t> <: <t'>)
```

These conditions need to be stricter, see https://dfinity.slack.com/archives/C016H8MCVRC/p1617755023039700

chenyan-dfinity, line 935:
```suggestion
C[<t> <: opt <t'>](_)   = null                   otherwise
```
We don't have a rule for `null <: t'` then. I think this applies to the `null <: t'` case as well?

chenyan-dfinity: How about the empty record case?
`C[ μx. record { x } <: μx. opt record { x } ](_) = opt C[μx. record { x } <: record { μx. opt record { x } } ]`

Do we want to allow `μx. record { x }` to be subtype of `opt t`?

nomeata: Every type is a subtype of `opt t`. The question is only whether it coerces to `null` or something else.

I think this case is ok: Note that the coercion function takes a value `v : µx.record{x}`, and no such value exists, so `C[ μx. record { x } <: μx. opt record { x } ]` is (independent of the equations) the empty function with an empty domain.

nomeata: I think you are right. But let’s be explicit here, and and not rely on fall-through-semantics.

So it seems the fomulation that we want, so that each line is correct independent of their order, might be

```
C[<t> <: opt <t'>](<v>) = opt C[<t> <: <t'>](v)  if not (null <: <t'>) and <t> <: <t'>
C[<t> <: opt <t'>](_)   = null                   if not (null <: <t'>) and not (<t> <: <t'>)
C[<t> <: opt <t'>](_)   = null                   if null <: <t'> and not (null <: <t>)
```

(Remember that `null <: t` implies that `t = null`, `t = opt …` or `t = reserved`, and these cases have their own rules)


I think this needs a table:

| null <: t | null <: t' | t <: t' | Rule?
| --- | --- | --- | --- |
| no | no | no | second rule
| no | no | yes | first rule
| no | yes | no | third rule
| no | yes | yes | third rule
| yes | no | no | second rule
| yes | no | yes | impossible (<: is transitive)
| yes | yes | no | see rules for  `t = null`, `t = opt …` or `t = reserved`
| yes | yes | yes | see rules for  `t = null`, `t = opt …` or `t = reserved`

chenyan-dfinity: I'm not sure how to detect loops in this case. The usual trick is to count the depth of the continuous nesting of records. But here, `µx.record{x}` doesn't change, we are only reducing the expected type, i.e. `C[ μx. record { x } <: μx. opt record { x } ] = opt C[ μx. record { x } <: μx. opt record { x } ]`. Operationally, the type on the left doesn't change at all, so we cannot count the depth. 

Of course, there are more expensive ways to detect loops, e.g. memoization or decoding the value with the actual type first. But to rule out other possible implementations just because of this case seems a bit wasteful.

chenyan-dfinity:
The variant rule is a bit unexpected. While it's true that we can add tags via the opt rule, so `opt variant { a } <: opt variant { a; b }`. But because `variant { a }` is not a subtype of `variant { a; b }`, if the receiver expects `opt variant {a;b}`, and the sender sends `opt variant {a}`, the receiver always gets `null`. 

I was hoping we can still decode `opt variant {a}`. Otherwise, adding tags is effectively a breaking change: Until the sender upgrades the type, it is always null.

spec/Candid.md
nomeata: Is this really a new problem? The problem with `µx.record{x}` is about decoding, not coercion, and that isn’t changed by this PR? There are no values of that type anyways, so one wouldn’t ever get past the decoding stage before erroring out (or looping.)

In the Motoko decoder, I bump a counter whenever I go from `record { t1, … }` to `t1`, and reset whenever I actually consum a byte from the input stream. If that counter exceeds the table size I have recognized a loop.

In the Haskell decoder, I pre-process the type table and replace any empty type like `µx. record { …., x , …}` with `empty`.

nomeata:
> Also, for the main service type, adding a new method is a breaking change?

No, I don't think so. Why should it? Do you have an example?

chenyan-dfinity:
> > Also, for the main service type, adding a new method is a breaking change?
> 
> No, I don't think so. Why should it? Do you have an example?

`service { <methtype'>;* } <: service { }`. This means we can only remove methods?

nomeata:
No, it means you can only _add_ methods, and never remove any: You upgrade a service towards a _sub_-type (so that the new version can stand in where the old version is expected).

chenyan-dfinity:
Everything here is implemented at https://github.com/dfinity/candid/pull/211. Should we merge this PR?

nomeata:
> Everything here is implemented at #211. Should we merge this PR?

I guess. How much breakage do you expect for users of the library? I wonder if it would make sense to hold off until after launch…

I see you expanded the test suite a bit, thanks. Are you content the the coverage? In other words: If someone who isn’t you implements this logic (say in another language), and they get the test suite to pass, would you trust that they had to think about and fix all the corner cases?

chenyan-dfinity:
Maybe we can merge this to `beta` branch for now. This enables us to implement the new spec in other languages by pulling in the test suites.

> How much breakage do you expect for users of the library?

From a user's perspective, the breakage is minimal: requires both serialize and deserialize trait for deserialization; relocate some traits to another module.

For test coverage, I think it's reasonably well. This PR doesn't change much of the existing tests. Out of 200+ tests, we only need to update around 5 tests.

nomeata:
> Maybe we can merge this to `beta` branch for now. This enables us to implement the new spec in other languages by pulling in the test suites.

Good idea
 
> For test coverage, I think it's reasonably well. This PR doesn't change much of the existing tests. Out of 200+ tests, we only need to update around 5 tests.

I’d rather conclude that the coverage is reasonably bad then, as we don’t have many tests that are affected by this change ;-)

I expect we need a fair amount of tests where we check if subtyping is really properly checked even in the absence of values (e.g. with a `vec {}` value, or in references.)

chenyan-dfinity:
> I expect we need a fair amount of tests where we check if subtyping is really properly checked even in the absence of values (e.g. with a vec {} value, or in references.)

Good catch for the `vec {}` case. I add tests for record, opt and op variant already. Let's merge this to `beta`, and we can iterate.

PR #171, opened 2021-01-27
Coq: Model subtype check on decoding, IDL-Soundness, Transitive Coherence

nomeata:
A revamp of the Coq development:

 * It models the subtype-checking on decoding (#168). Looks good
 * It connects MiniCandid to the IDL-Soundness theorem. The main work here is the subtyping-compositoinaliy lemma.
   ```
   If t1 <: t2 and s1 in t1 <: s2 in t2 then s1 <: s2.
   ```
   With this in place, instantiating the “canonical subtyping” proof there works nice.
 * It proofs transitive coherence with regard to the relaxed relation as per #173
 * Mild coqdoc’ifiacation. I’d like to eventually render these to HTML and host them somewhere.
   It’s very annoying that Github Action artifacts, even if they are HTML, are not directly accessible with the browser.
   Maybe setup Github pages?
   
It is still a Mini-Candid with a limited set of types, but I think it has all the interesting ones to cover the corner cases. Even adding vectors adds a lot of technical noise with little additional insight (see #154.)


nomeata:
(Should be merged after #168 hits master)

coq/MiniCandid.v
```diff
@@ -115,428 +123,679 @@ CoInductive Subtype : T -> T -> Prop :=
...
+
+  | SomeV v, OptT t1, OptT t2 =>
+    if t1 <:? t2
+    then SomeV (coerce t1 t2 v)
+    else NullV
+  
+  (* We’d prefer the equation from [coerce_consituent_eq] below,
+     but that will not satisfy the termination checker,
+     so let’s repeat all the above ruls for OptT again.
+  *)
```
rossberg, line 191: Hm, I'm confused why this does not need a case for e.g. `OptT ReservedT`?

coq/MiniCandid.v
```diff
@@ -115,428 +123,679 @@ CoInductive Subtype : T -> T -> Prop :=
...
+  end.
 
-Lemma ConstituentOptC:
-    forall v1 v2 t,
-    is_not_opt_like_value v1 ->
-    is_opt_like_type t = false ->
-    v1 ~> v2 :: t ->
-    v1 ~> SomeV v2 :: OptT t.
+(* We can prove the desired equation at least as an equality *)
+Lemma coerce_consituent_eq:
```
rossberg, line 206:
```suggestion
Lemma coerce_constituent_eq:
```

coq/MiniCandid.v
```diff
@@ -115,428 +123,679 @@ CoInductive Subtype : T -> T -> Prop :=
...
+  | Here => Some v
+  | The p =>
+    match v with
+    | SomeV v => val_idx p v
+    | _ => None
+    end
+  end.
+
+(**
+This is a lenitent variant, which is total (returning [NullV]
```
rossberg, line 483: Did you mean lenient?

coq/MiniCandid.v
```diff
@@ -115,428 +123,679 @@ CoInductive Subtype : T -> T -> Prop :=
...
+    match v with
+    | SomeV v => val_idx' p v
+    | NullV => NullV
+    | ReservedV => NullV
+    | _ => val_idx' p v
+    end
+  end.
+
+(**
+The corresponding function for types, also lenitent.
```
rossberg, line 503: Same here.

coq/MiniCandid.v
```diff
@@ -115,428 +123,679 @@ CoInductive Subtype : T -> T -> Prop :=
...
+Lemma no_new_values:
+  forall t1 t2 v1,
   t1 <: t2 ->
-  forall v1, v1 :: t1 -> exists v2, v1 ~> v2 :: t2.
+  v1 :: t1 ->
+  forall p iv2,
+  val_idx p (coerce t1 t2 v1) = Some iv2 ->
+    (iv2 = NullV \/ typ_idx' p t1 <: typ_idx' p t2) /\
+    val_idx' p v1 :: typ_idx' p t1 /\
+    coerce (typ_idx' p t1) (typ_idx' p t2) (val_idx' p v1) = iv2.
```
rossberg, line 574: Nice.

rossberg (commented):
Cool stuff. I had to realise far behind the curve my Coq expertise is, so I probably only have superficial comments on definitions and theorems.

coq/MiniCandid.v
nomeata: Which case exactly are you missing? (`coerce` has three parameters)

rossberg: I guess `ResV, ResT, OptT ResT`? (I was also wondering about OptT NullT, but I suppose that's covered by the default.)

nomeata: That is handled by the final `NullV` producing catch-all. Remember that the constituent-to-opt rule only applies to `t <: opt t` if `not (null <: opt t)`.

So `reserved <: opt reserved` will always result `null`.

spec/Candid.md
```diff
@@ -910,13 +910,12 @@ NB: No rule is needed for type `empty`, because there are no values of that type
 
 Vectors coerce pointwise:
 ```
-C[vec <t> <: vec <t'>]( vec { <v>;* } ) = vec { C[<t> <: <t'>](<v>);* }
+C[vec <t> <: vec <t'>]( vec { <v>;* } ) = vec { c[<t> <: <t'>](<v>);* }
```
nomeata, line 913:
```suggestion
C[vec <t> <: vec <t'>]( vec { <v>;* } ) = vec { C[<t> <: <t'>](<v>);* }
```

PR #187, opened 2021-02-14
didc call canister method

chenyan-dfinity:


nomeata (commented):
Huh `candid` to depend on the IC agent? Isn't that a blatant inversion of the natural dependencies?

Is there a reason why `candid` cannot stay a pure library for Candid, and stuff like calling canisters goes into `dfx` or `icx`?

chenyan-dfinity:
This dependency is not part of the `candid` library, but in the binary `didc` tool. This is extracted from a side project I was working on for Candid REPL and canister testing.

We certainly don't want ic-agent to be part of the candid library, but for tooling, the line is a bit blurry to me. I will convert this PR to a draft and explore it on the side.

nomeata:
Hmm, maybe for tools it's a bit different. But still feels a bit odd; you don't expect `jq` to speak REST APIs? Also we do sell Candid as something that could be used independently of the IC, so again such a dependency would be odd

PR #189, opened 2021-02-17
[Tool] Canister REPL

chenyan-dfinity:


nomeata:
Fun! JFTR, though, I still think that nothing in `candid` (the repo) should depend on the agents, and `candid` should remain a crate purely (heh)  dedicated to Candid stuff. I would expect a look like this to be either part of the agent repo (agnet depending on candid is more justifiable), part of the `sdk` repo, or its own thing.

chenyan-dfinity:
> nothing in candid (the repo) should depend on the agents

It's already not true. The Candid UI does depend on JS agent, and it's more suitable to live in the candid repo, as it is mostly about interpreting Candid types.

Same reasoning for the REPL, we are not using agent in any interesting ways, but this project opens up some language questions:
* How to derive auto-completion from grammar rules and typing information
* Better assistance in writing Candid values (test script, seeding data), as Candid values are not optimized for writing.

I would imagine this tool to be absorbed by dfx eventually, but there is still a long way for this to be merged.

nomeata:
> > nothing in candid (the repo) should depend on the agents
> 
> It's already not true. The Candid UI does depend on JS agent, and it's more suitable to live in the candid repo, as it is mostly about interpreting Candid types.

I disagree, CandidUI is mostly about interacting with services on the Internet Computer

> Same reasoning for the REPL, we are not using agent in any interesting ways, but this project opens up some language questions:
> 
>     * How to derive auto-completion from grammar rules and typing information
> 
>     * Better assistance in writing Candid values (test script, seeding data), as Candid values are not optimized for writing.

Sure, every _use_ of Candid may impose new insights into what a Candid library needs to be provide. That doesn’t mean that the candid library itself has to be tied to a specifc agent (or an agent at all).

You (well, I) don’t expect libraries and tools that decodes mime attachements  to speak SMTP.
I don’t expect a libraries and tools that deal with protobuf to speak RPC.
A HTML parsing library doesn't speak HTTP.

I believe that separation of concerns and composition (one repo _exclusively_ about Candid, one repo _exclusively_ about iterfacing with the IC, and then tools on top of that) pays off in the long run. More than avoiding creating a fresh `ic-repl` repo for the work in this PR.

Also (and kinda unfortunately from a separation of concern point of view), once https://github.com/dfinity-lab/ic-ref/issues/248 lands the agent actually have to have some Candid support. So if anything, it might be better to keep code and tools that depend on both agent and candid together with the agent. E.g. it the repl could be an extension to https://github.com/dfinity/agent-rs/tree/next/icx?

chenyan-dfinity:
Moved to https://github.com/chenyan2002/ic-repl

PR #192, opened 2021-02-26
subtyping check

chenyan-dfinity:
Part of https://github.com/dfinity/dx-triage/issues/82

- [x] Memoization for recursion
- [x] Subtyping for reserved and int

### DIFF
--- a/rust/candid/src/types/subtype.rs
+++ b/rust/candid/src/types/subtype.rs
@@ -1,13 +1,8 @@
 use super::internal::{find_type, Field, Label, Type};
 use crate::parser::typing::TypeEnv;
-use std::cell::RefCell;
 use std::collections::{HashMap, HashSet};
 
 pub type Gamma = HashSet<(Type, Type)>;
-
-thread_local! {
-    pub static COUNT: RefCell<usize> = RefCell::new(0);
-}
 
 /// Check if t1 <: t2
 pub fn subtype(gamma: &mut Gamma, env: &TypeEnv, t1: &Type, t2: &Type) -> bool {
@@ -31,9 +26,6 @@ pub fn subtype(gamma: &mut Gamma, env: &TypeEnv, t1: &Type, t2: &Type) -> bool {
         }
         return res;
     }
-    COUNT.with(|e| {
-        *e.borrow_mut() += 1;
-    });
     match (t1, t2) {
         (_, Reserved) => true,
         (Empty, _) => true,
--- a/rust/candid/src/types/subtype.rs
+++ b/rust/candid/src/types/subtype.rs
@@ -30,49 +30,26 @@ pub fn subtype(env: &TypeEnv, t1: &Type, t2: &Type) -> bool {
         }
         (Record(fs1), Record(fs2)) => {
             let fields: HashMap<_, _> = fs1.iter().map(|Field { id, ty }| (id, ty)).collect();
-            for Field { id, ty: ty2 } in fs2.iter() {
-                match fields.get(id) {
-                    Some(ty1) => {
-                        if !subtype(env, ty1, ty2) {
-                            return false;
-                        }
-                    }
-                    None => {
-                        if !subtype(env, &Opt(Box::new(Empty)), ty2) {
-                            return false;
-                        }
-                    }
-                }
-            }
-            true
+            fs2.iter()
+                .all(|Field { id, ty: ty2 }| match fields.get(id) {
+                    Some(ty1) => subtype(env, ty1, ty2),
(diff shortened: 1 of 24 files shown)


rust/candid/src/types/subtype.rs
```diff
@@ -0,0 +1,130 @@
...
+                        }
+                    }
+                    None => {
+                        if !subtype(env, &Opt(Box::new(Empty)), ty2) {
+                            return false;
+                        }
+                    }
+                }
+            }
+            true
```
nomeata, line 47: Maybe nicer when using `fs2.iter().all(|Field {id, ty : ty2 }| … )`? Or is that just my functional programming fetish at work?

chenyan-dfinity: Thanks, that's indeed better!

rust/candid/src/types/subtype.rs
```diff
@@ -0,0 +1,83 @@
...
+        (Opt(ty1), Opt(ty2)) if subtype(env, ty1, ty2) => true,
+        (t1, Opt(ty2)) if subtype(env, t1, ty2) && !subtype(env, &Null, ty2) => true,
+        (Opt(ty1), Opt(_ty2)) if !subtype(env, ty1, t2) => {
+            eprintln!("{} <: {} via special opt rule", t1, t2);
+            true
+        }
+        (Opt(ty1), Opt(ty2)) if !subtype(env, ty1, ty2) && !subtype(env, &Null, ty1) => {
+            eprintln!("{} <: {} via special opt rule", t1, t2);
+            true
+        }
```
nomeata, line 30: You can simplify these, and simply use `t < opt t'` for all types `t` and `t'`. The spec lists more complicated rules, but really just for educational purposes; that single rule would suffice (also cf. the definition in the Coq proof)

chenyan-dfinity: From the rules in the spec, `opt opt int <: opt int` doesn't hold.

nomeata: By this rule it should?
```
```

Really, the intention is that every type is a subtype of `opt t`, else using that for new optional fields in records doesn't works. I'd the rules in the spec are not exhaustive here, that's a bug in the spec.

rust/candid/src/types/subtype.rs
```diff
@@ -2,52 +2,66 @@ use super::internal::{find_type, Field, Label, Type};
...
+            return true;
+        }
+        let res = match (t1, t2) {
+            (Var(id), _) => subtype(gamma, env, env.rec_find_type(id).unwrap(), t2),
+            (_, Var(id)) => subtype(gamma, env, t1, env.rec_find_type(id).unwrap()),
+            (Knot(id), _) => subtype(gamma, env, &find_type(id).unwrap(), t2),
+            (_, Knot(id)) => subtype(gamma, env, t1, &find_type(id).unwrap()),
+            (_, _) => unreachable!(),
+        };
+        gamma.remove(&(t1.clone(), t2.clone()));
```
nomeata, line 24: I don't think you have to remove them: if you reach this point, the two types  _are_ subtypes, and you can use this information freely in further passes. Again, compare the ocaml code for subtyping in the Motoko compiler.

rust/candid/src/types/subtype.rs
chenyan-dfinity: It's a typo then. The rule in the spec says:

```
```

rust/candid/src/types/subtype.rs
chenyan-dfinity: We need to remove when `res` is `false`. The OCaml code doesn't need to do this because it's using a functional set. In Rust, `HashSet` is imperative.

rust/candid/src/types/subtype.rs
nomeata: Hmm, indeed, I must have copied from a wrong version. But note that the prose says

> Finally, in order to maintain transitivity of subtyping, two unusual rules allow, in fact, any type to be regarded as a subtype of an option.

so `opt opt int <: opt int` _must_ hold, and if it doesn't, the rules are still buggy.

I see this rule in the spec:
```
not (null <: <datatype>)
```
but I think it should be
```
not (null <: <datatype'>)
```
to complement the rule
```
not (null <: <datatype'>)
<datatype> <: <datatype'>
-----------------------------
<datatype> <: opt <datatype'>
```

rust/candid/src/types/subtype.rs
nomeata: When `res` is false, the whole function exits, and you through away `gamma`, doesn’t it?

PR #204, opened 2021-03-13
fix serde_bytes derive

chenyan-dfinity:
Restrict `#[serde(with = "serde_bytes"]` to be only used for `Vec<u8>` and `&[u8]` in `struct` field.


chenyan-dfinity:
Actually this is fine, Rust type checker is able to detect this error. We don't need to catch this from the Rust AST in proc macro.

PR #206, opened 2021-03-16
Fix build for rustc 1.45.2

osa1:


chenyan-dfinity (approved):

chenyan-dfinity:
> How can this change cause changes in benchmarks? Seems like bad benchmark setup or unreliable benchmarks.

None of the numbers are in bold, meaning that the confidence interval is too wide to be considered a reliable change. In general, measuring wall time in virtual CI is not reliable. It's best to run the benchmark locally if you want reliable numbers.

PR #215, opened 2021-04-07
JS binding for BigInt

chenyan-dfinity:


rust/candid/tests/assets/ok/class.d.ts
```diff
@@ -1,7 +1,6 @@
 import type { Principal } from '@dfinity/agent';
-import type BigNumber from 'bignumber.js';
 export type List = [] | [
-  [BigNumber, List]
+  [BigInt, List]
```
krpeacock, line 3: In spots like this, the type would be `bigint`, which means we expect an instance of a BigInt

```suggestion
  [bigint, List]
```

rust/candid/tests/assets/ok/escape.d.ts
```diff
@@ -1,10 +1,9 @@
 import type { Principal } from '@dfinity/agent';
-import type BigNumber from 'bignumber.js';
 export interface t {
-  '\"' : BigNumber,
-  '\'' : BigNumber,
-  '\"\'' : BigNumber,
-  '\\\n\'\"' : BigNumber,
+  '\"' : BigInt,
```
krpeacock, line 3: See above

```suggestion
  '\"' : bigint,
```

Issue #217, opened 2021-04-12
In Rust, there is no type to encode/decode a `Func` Candid type


hansl:
Right now we have to use an `IDLValue` which isn't ideal (as then we have to match everywhere). There should be a native encapsulation, similar to how we do `Nat` and `Int`.

PR #218, opened 2021-04-14
Rust support for Func and Service

chenyan-dfinity:
Fix https://github.com/dfinity/candid/issues/217

### DIFF
--- a/rust/candid/src/types/reference.rs
+++ b/rust/candid/src/types/reference.rs
@@ -1,0 +1,96 @@
+//! Data structure for Candid value Func and Service
+
+use super::{CandidType, Function, Serializer, Type};
+use ic_types::Principal;
+use serde::de::{self, Deserialize, Visitor};
+use std::convert::TryFrom;
+use std::{fmt, io::Read};
+
+#[derive(PartialEq, Debug, Clone)]
+pub struct Func {
+    pub principal: Principal,
+    pub method: String,
+}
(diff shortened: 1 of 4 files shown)


rust/candid/src/types/reference.rs
```diff
@@ -0,0 +1,96 @@
...
+        D: serde::Deserializer<'de>,
+    {
+        struct FuncVisitor;
+        impl<'de> Visitor<'de> for FuncVisitor {
+            type Value = Func;
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("Func value")
+            }
+            fn visit_byte_buf<E: de::Error>(self, bytes: Vec<u8>) -> Result<Func, E> {
+                if bytes[0] != 5u8 {
```
nomeata, line 58: Maybe I am lost, but what does the value `5` mean here?

chenyan-dfinity: It's an ugly hack just to make serde happy. It's not related to the spec at all.

serde generates values via this `Visitor` trait, and it cannot consult any internal state in the deserializer. So the data we pass to the visitor should be self-contained. This `5` tag is an internal representation of `Function` value. This tag is generally not needed, because we can build different visitors for different values, e.g. `FuncVisitor`, `IntVisitor`. But to decode into the untyped `IDLValue`, we need this tag to distinguish different values.

PR #219, opened 2021-04-14
fix typescript binding for tuple

chenyan-dfinity:


krpeacock (approved):
thanks for the fast turnaround!

PR #230, opened 2021-05-23
Create didc-js

FloorLamp:
`didc-js` is a JS/Wasm package for generating Candid bindings, useful in client-side UIs. It simply calls `candid::bindings::javascript::compile` and `candid::bindings::typescript::compile`.


chenyan-dfinity:
I like the direction of this. But maintaining it as a general JS library feels a bit ad-hoc, since people may want to expose different APIs. It would be a great example to put in the doc to show how to use the Rust library in JS. WDYT?

chenyan-dfinity:
Moved to https://github.com/dfinity/candid/pull/233 as a section in `rustdoc`

PR #231, opened 2021-05-25
Motoko binding

chenyan-dfinity:
Will make another PR to handle blob -> Blob mapping properly.


rust/candid/src/bindings/motoko.rs
```diff
@@ -0,0 +1,236 @@
+use super::candid::is_valid_as_id;
```
nomeata, line 1: Nice!

I assume this is implementing the `i(…)` function from https://github.com/dfinity/motoko/blob/master/design/IDL-Motoko.md?

Maybe include a link to that document in the code here.

rust/candid/src/bindings/motoko.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+pub fn compile(env: &TypeEnv, actor: &Option<Type>) -> String {
+    let header = r#"// This is a generated Motoko binding. Please use `import service "ic:canister_id"` instead to call canisters on the IC if possible.
+"#;
+    let doc = match actor {
+        None => pp_defs(env),
+        Some(actor) => {
+            let defs = pp_defs(env);
+            let actor = kwd("type _SERVICE =").append(pp_actor(actor));
```
chenyan-dfinity, line 227: @nomeata Do you have a better idea to name the main actor? Put everything in a module or flat?

nomeata: I guess everything in a module makes sense, so that you can import the resulting `.mo` file, assuming that is the goal?

No good idea for a better name. When the optional name in
```
<actor> ::= service <id>? : (<tuptype> ->)? (<actortype> | <id>)
```
is present, one can use that. Otherwise, maybe `Self`, so that you can use `M.Self`? but `Service` isn’t bad either… no great idea, sorry.

Too bad we cannot somehow pun modules and types this, so that after
```
import M "generated_file.mo";
```
we can write `M` for the service type, and still get access to `M.Foo` for the types defined in `input.did`. It would also be closer to what `moc` does when you import a `.did` right now.

@crusso, do you have more ideas here?

chenyan-dfinity: Yeah, `Self` is better, because it's sometimes a service, sometimes a service class.

PR #237, opened 2021-06-03
Change the version label from master to latest

lsgunnlsgunn:
**Overview**
To be more PC and "accurate" changing the label for the docs bucket from "master" to "latest"

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -1,2 +1,2 @@
 name: docs
-version: 'master'
+version: 'latest'


lsgunnlsgunn:
Looks like this isn't going to be possible with how the site-entry page is being circumvented.

rossberg:
I believe the new naming convention for the master branch, at least on GitHub, is `main`. Should we follow that?

PR #235, opened 2021-05-29
docs: Update interacting with the Candid UI

lsgunnlsgunn:
**Overview**
Update the How to section because working with the Candid UI is different in 0.7.0.
Fix a missing link.

PR #243, opened 2021-06-05
Fix "Interact with a service from a browser" in Candid Guide

johnkruger37:
https://sdk.dfinity.org/docs/candid-guide/candid-howto.html#candid-ui as-is seems wrong, that does not work (for me, on current/latest `dfx 0.7.1`).

This fixes the doc to reflect how this appears to be intended to be used.


lsgunnlsgunn:
You are correct, The Candid UI flow changed recently and the doc needed an update.
Fixed in https://github.com/dfinity/candid/pull/235 so closing this as a duplicate of that.

PR #244, opened 2021-06-07
chore: updates type import to new home in @dfinity/principal

krpeacock:
**Overview**

We just updated `agent-js` with a new release to have @dfinity/principal live in its own repo, and now the import of Principal from `@dfinity/agent` is outdated.

**Considerations**

It's a breaking change with older versions of agent, but one we decided was worth the costs.


tools/ui/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "webpack"
   },
   "devDependencies": {
-    "@dfinity/agent": "0.8.4",
+    "@dfinity/principal": "0.9.0",
```
chenyan-dfinity, line 10: This will probably break the UI code. Let's either fix the import, or keep the version at `0.8.9`

krpeacock: Ah you're right. I'll fix the other references so UI is up to date with the new 0.9.0 imports

chenyan-dfinity (approved):

PR #247, opened 2021-06-17
test suite: A few more subtyping tests

nomeata:


nomeata:
This fails with errors in `lint`… unlikely a cause of this PR. Looks like CI picked up a newer version of cargo or clippy or so?

chenyan-dfinity (approved):

Issue #248, opened 2021-06-18
As a library developer, I want to have candid_path rethought so that consumers of my library do not need to depend on ic_cdk directly


lastmjs:
**Is your feature request related to a problem? Please describe.**

I am developing Sudograph, a procedural macro that is meant to be used within a Rust canister. Sudograph generates a GraphQL database, and reexports ic_cdk and other crates from cdk-rs. Right now, I have it designed so that the consumer of Sudograph only needs to depend on `sudograph` in their Cargo.toml for their canister. But as I've gotten into enabling cross-canister GraphQL resolvers, the `CandidType` attribute macro has caused me problems. Because of the way `candid_path` is designed, see here: https://github.com/dfinity/candid/blob/master/rust/candid_derive/src/lib.rs#L36 I have to force the consumers of Sudograph to depend directly on `ic_cdk`, when otherwise they would not have to. I think the root of the problem is hardcoding  this path `::ic_cdk::export::candid`, especially the prefix `::`.

**Describe the solution you'd like**

Consider changing the hard-coded `candid_path` to `ic_cdk::export::candid`, or allow the developer to set the crate location manually in another macro. For example, serde had a very similar problem that was solved with the introduction of this `#[serde(crate="self::serde")]`, allowing the developer to specify the location of the serde crate (or something like that, I'm still trying to wrap my head around Rust's packages, crates, and modules, but the serde solution works).

**Describe alternatives you've considered**

I could always force the consumer of Sudograph to depend directly on `ic_cdk`, but it would be an extra inconvenience to them. Part of Sudograph's value proposition is being extremely simple, a sort of one-stop shop for developing on the IC using GraphQL. I have been messing with modules and crates to try and get this to work without needing changes to the `candid` libraries, but I'm not sure it's possible, and the fix on your end probably isn't too complicated.

**Additional context**

I can give more context on the serde issue if desired, I would need to dig up the issue. Please understand that this is about the ergonomics and ease of use of libraries like Sudograph, right now the `CandidType` attribute macro is somewhat tightly coupled to a specific project structure and could be made more composable.

chenyan-dfinity:
Let me try to better understand the problem here. The reason `::` prefix doesn't work is because `sudograph` exports its own version of `ic_cdk`? Note that even if we fix this, the user would still need to directly depend on `serde`, because the derive code for `Deserialize` is hardcoded to `serde`.

[...]

PR #249, opened 2021-06-21
add candid_type helper attribute

lastmjs:
**Overview**

This adds a `#[candid_path("path_to_candid")]` helper attribute to the candid derive macro, in response to this issue: https://github.com/dfinity/candid/issues/248

**Requirements**

I've tested `#[candid_path("path_to_candid")]` in Sudograph, the original project I encountered the problem with. I suppose testing `#[candid_path("path_to_candid")]` in variety of crates, with candid in different paths, would show the problem as resolved.

**Considered Solutions**

Mostly a new helper attribute such as `#[candid_path("path_to_candid")]` or `#[candid(crate="path_to_candid")]`. More information in the issue linked to above.

**Recommended Solution**

The introduction of `#[candid_path("path_to_candid")]`. This was easy to implement, and shouldn't affect existing functionality. Only users who need to specify a custom path need to use `#[candid_path("path_to_candid")]`, otherwise functionality should remain unchanged.

**Considerations**

There should be no breaking changes, performance changes should be minimal, security I am not sure on, probably similar security problems that are already present with macros, and maintenance should be minimal.


rust/candid_derive/src/lib.rs
```diff
@@ -32,11 +34,34 @@ pub(crate) fn idl_hash(id: &str) -> u32 {
...
+    custom_candid_path: &Option<proc_macro2::TokenStream>,
+) -> proc_macro2::TokenStream {
+    match custom_candid_path {
+        Some(custom_candid_path_value) => custom_candid_path_value.clone(),
+        None => quote::quote! { ::candid },
+    }
+}
+
+// TODO I am not sure I parsed this in the best way possible...any pointers would be great
+fn get_custom_candid_path(input: &syn::DeriveInput) -> Option<proc_macro2::TokenStream> {
```
chenyan-dfinity, line 57: It's better to return `Result<TokenStream>` so that we can customize the error message, see https://github.com/dfinity/candid/blob/master/rust/candid_derive/src/func.rs#L215

At the call site, we can convert `syn::Error` to a `compile_error`, see https://github.com/dfinity/candid/blob/master/rust/candid_derive/src/lib.rs#L17

chenyan-dfinity (commented):
Thanks for doing this!

rust/candid_derive/src/lib.rs
lastmjs: Awesome, I have addressed this

chenyan-dfinity (approved):
LGTM!

For the format issue, you can run `cargo fmt --all` to auto-fix format warnings.

lastmjs:
> LGTM!
> 
> For the format issue, you can run `cargo fmt --all` to auto-fix format warnings.

This is great to know!

Is there anything else I need to do for this PR? The linting errors do not seem to be related to any code that I added

chenyan-dfinity:
I will fix these and merge. Thanks!

lastmjs:
Any ETA on a new release of the candid crate with these changes? And an ic-cdk crate that references the new candid version?

lastmjs:
And do you have any advice on how to depend on a crate like this before it is published? The problem is I publish my crate that depends on ic-cdk which depends on candid, and I can't depend directly on a github version in my crate that is published to crates.io. `patch` and `replace` are also not working because of versioning issues. Any other way?

chenyan-dfinity:
I will make a release this week. It's a major version bump for ic-cdk, otherwise, you can use `patch` for crates.io. Another way is for your crate to depend directly on the github version of candid, but that's also not optimal.

chenyan-dfinity:
CDK updated here: https://github.com/dfinity/cdk-rs/pull/72. You may be able to use it before getting released.

PR #250, opened 2021-06-24
feat: move CandidType cache to CandidTypeCache trait

Daniel-Bloom-dfinity:
**Overview**
This makes it possible to move away from a global thread local static type cache.

**Requirements**
Eventual deprecation and removal of `internal::env_id` would be the end goal of this feature.

**Considered Solutions**
Most solutions to this end up needing to refactor these traits.

**Recommended Solution**
This solution forces caching for all `CandidType`s, removes the need for `unreachable` impls of `CandidType`, while still allowing `IDLValue` to override default behavior, while only minimally altering the overall interface.

**Considerations**
There are no notable security or performance considerations, but this does change the interface in a way which could break some power users. Ordinary users who are just using `#[derive(CandidType)]` should not notice any difference. Additional traits does add additional complexity.

### DIFF
--- a/rust/candid/src/ser.rs
+++ b/rust/candid/src/ser.rs
@@ -341,7 +341,7 @@ impl TypeSerialize {
     fn encode(&self, buf: &mut Vec<u8>, t: &Type) -> Result<()> {
         if let Type::Var(id) = t {
             let actual_type = self.env.rec_find_type(id)?;
-            if types::internal::is_primitive(&actual_type) {
+            if types::internal::is_primitive(actual_type) {
                 return self.encode(buf, actual_type);
             }
         }
@@ -376,14 +376,14 @@ impl TypeSerialize {
             Type::Var(_) => {
                 let idx = self
                     .type_map
-                    .get(&t)
+                    .get(t)
                     .ok_or_else(|| Error::msg(format!("var type {} not found", t)))?;
                 sleb128_encode(buf, i64::from(*idx))
             }
             _ => {
                 let idx = self
                     .type_map
-                    .get(&t)
+                    .get(t)
                     .ok_or_else(|| Error::msg(format!("type {} not found", t)))?;
                 sleb128_encode(buf, i64::from(*idx))
             }
--- a/rust/candid/src/ser.rs
+++ b/rust/candid/src/ser.rs
@@ -2,7 +2,7 @@
 
 use super::error::{Error, Result};
 use super::parser::{typing::TypeEnv, value::IDLValue};
-use super::types::{self, CandidTyping, IdlSerialize, internal::Opcode, Field, Type};
+use super::types::{self, internal::Opcode, CandidTyping, Field, IdlSerialize, Type};
 use byteorder::{LittleEndian, WriteBytesExt};
 use leb128::write::{signed as sleb128_encode, unsigned as leb128_encode};
 use std::collections::HashMap;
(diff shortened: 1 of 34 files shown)


rust/candid_common/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+/// Hash function for [Symbolic Field Ids](https://github.com/dfinity/candid/blob/master/spec/Candid.md#shorthand-symbolic-field-ids)
+#[inline]
+pub fn idl_hash(id: &str) -> u32 {
```
chenyan-dfinity, line 3: It feels a bit overkill to have a crate that just contains the hash function. The hash function is not likely to change in the future, so duplicating the code in the derive crate is not a concern.

rust/candid/src/types/mod.rs
```diff
@@ -17,31 +17,66 @@ pub mod principal;
...
+    }
+    fn add_type(&mut self, id: TypeId, ty: Type) {
+        if ty != Type::Unknown {
+            self::internal::env_id(id.clone(), ty.clone());
+        }
+        self::internal::env_add(id, ty);
+    }
+}
+
+pub trait CandidType: IdlSerialize {
```
chenyan-dfinity, line 74: What's the difference between `CandidType` and `CandidTyping`?

rust/candid/src/types/impls.rs
```diff
@@ -1,16 +1,159 @@
 use super::internal::*;
-use super::{CandidType, Compound, Serializer};
+use super::{CandidType, CandidTypeCache, CandidTyping, Compound, IdlSerialize, Serializer};
+
+macro_rules! serialize_impl {
```
chenyan-dfinity, line 4: Too many macros affects readability. Although the code is shorter, it is harder to debug.

rust/candid/src/ser.rs
```diff
@@ -28,13 +27,12 @@ impl IDLBuilder {
             value_ser: ValueSerializer::new(),
         }
     }
-    pub fn arg<'a, T: types::CandidType>(&'a mut self, value: &T) -> Result<&'a mut Self> {
+    pub fn arg<'a, T: ?Sized + CandidTyping>(&'a mut self, value: &T) -> Result<&'a mut Self> {
```
chenyan-dfinity, line 30: I was expecting that we can get rid of `types::internal::env_clear()` on line 24. If not, what's the point of this PR?

Daniel-Bloom-dfinity:
Full disclosure, this originally part of a failed attempt to remove the circular dependency between this repo and agent-rs. Essentially only proposing it for comment to see if any of the ideas are useful to salvage anything from the attempt.

rust/candid_common/src/lib.rs
Daniel-Bloom-dfinity: Agreed, this would have been more useful if the original design had panned out, since the end design would have had added a `candid_type` crate in addition that would have also depended on this new `candid_common`. As it stands, yeah this is pretty overkill.

rust/candid/src/types/mod.rs
Daniel-Bloom-dfinity: Basically I split `CandidTyping` off from `CandidType`, leaving all the implementation side `_ty` stuff in `CandidType`, but taking the caller side `ty` into `CandidTyping`. This allows `IDLValue` to override the `CandidTyping` without needing the `unreachable!` implementations.

The relationship between them is kind of like `From` and `Into`, except with the extra layer of memoization `CandidTyping` is responsible for invoking.

chenyan-dfinity: `IDLValue` is not supposed to be used for serializing native Rust values. The `unreachable` is intentional, so that users know it's an anti-pattern. Splitting the trait just to avoid this `unreachable` seems overkill. Also since `0.7.0`, `IDLValue::ty()` returns `Type::Unknown` instead of `unreachable`.

PR #257, opened 2021-07-03
chore: removes default exports from types

krpeacock:
**Overview**
This will support automatic type inference in `dfx` projects

**Changes**

- typescript bindings will no longer use default exports, instead favoring named _SERVICE export
- javascript exports named `idlFactory` instead of default export
- typescript interface no longer appends semicolon

**Considerations**

Some existing TypeScript projects may require updates after updating `dfx`

### DIFF
--- a/rust/candid/src/bindings/typescript.rs
+++ b/rust/candid/src/bindings/typescript.rs
@@ -137,7 +137,7 @@ fn pp_defs<'a>(env: &'a TypeEnv, def_list: &'a [&'a str]) -> RcDoc<'a> {
                 .append(pp_ty(ty))
                 .append(";"),
         };
-        export.append("")
+        export
     }))
 }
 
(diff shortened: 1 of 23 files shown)


rust/candid/src/bindings/typescript.rs
```diff
@@ -129,22 +129,24 @@ fn pp_defs<'a>(env: &'a TypeEnv, def_list: &'a [&'a str]) -> RcDoc<'a> {
...
+                .append(";"),
             _ => kwd("export type")
                 .append(ident(id))
                 .append(" = ")
-                .append(pp_ty(ty)),
+                .append(pp_ty(ty))
+                .append(";"),
         };
-        export.append(";")
+        export.append("")
```
chenyan-dfinity, line 140:
```suggestion
        export
```

chenyan-dfinity (approved):

PR #262, opened 2021-07-27
fix(UI): fetchRootKey for local replica

chenyan-dfinity:


tools/ui/src/candid.ts
```diff
@@ -5,7 +5,16 @@ import {
 import {Principal} from '@dfinity/principal'
 import './candid.css';
 
+function is_local(agent: HttpAgent) {
+  // @ts-ignore
+  const hostname = agent._host.hostname;
+  return hostname === '127.0.0.1' || hostname.endsWith('localhost');
```
kritzcreek, line 11: How is this not vulnerable to someone hosting `iclocalhost.app` and pwning you?

Issue #245, opened 2021-06-14
Feature Idea: Generic data


rossberg:
### Motivation

Candid cannot currently describe services or functions with a type-generic interface. That is a severe limitation that both we and users keep bumping into. For example, it is not possible to describe the interface of a generic key/value store, where the value could be any form of data (either homogeneously or heterogeneously).

Below I collect some thoughts from previous discussions for possible ways to lift this restriction.

### Idea 0: Parameterised type definitions

Summary: Allow type parameters in type definitions.

Examples:
```
type list(t) = variant { nil; cons : record {t; list(t)} }

type kv_store(t) : {
  put : (key : text, val : t) -> ();
  get : (key : text) -> (val : opt t) query;
  fill : (list : vec record {key : text; val : t}) -> ();
}
```

Since types are structural, this essentially is a simple macro mechanism that avoids having to repeat the same type many times with different instantiations. It's a prerequisite to make generic functions practical.

However, this requires additional checks to make sure that type recursion is well-founded and not diverging. If not restricted somehow, that is a significant extra burden on decoders (unless this is a text-only convenience, and types are duplicated in the serialised value, but that would be wasteful).

### Idea 1: Dynamic data type

Summary: Extend Candid with a new type `dyn`, which can carry any type of data. It is represented as a nested, self-contained Candid blob, complete with its own type description. (Type-theoretically, `dyn = exists X. X`)

This probably is the simplest approach. It essentially assigns a special type to blobs that contain nested Candid encodings.

Example:
```
service gen_kv_store : {
  put : (key : text, val : dyn) -> ();
  get : (key : text) -> (val : opt dyn) query;
  fill : (list : vec record {key : text; val : dyn}) -> ();
}
```

Pros: very simple
Cons:
* does not convey much intention (e.g., are all vals expected to have homogeneous type?)
* might be inefficient by repeating the same type many times (e.g., if the dyn vals in `fill` all share the same type, but this type is rather large; they cannot share any type defs with the outer blob either)

### Idea 2: Polymorphism encoded as `dyn`

Summary: Add syntactic sugar on top of `dyn` for expressing polymorphic functions with universally quantifed parameters. A value of parameter type is represented like `dyn`, but with the assumption that its internal type matches the instantiation of the type parameter.

Example:
```
service gen_kv_store : {
  put : (type t) (key : text, val : t) -> ();
  get : (type t) (key : text) -> (val : opt t) query;
[...]

PR #291, opened 2021-11-03
Add support for bound function arguments

rossberg:
This specs out the Candid-side of the "closure" extension I suggested [here](https://github.com/dfinity/motoko/issues/2703#issuecomment-951773868).

WDYT, does this work?

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -464,13 +464,25 @@ A *function reference* is described by its function type. For example, they allo
 <reftype> ::= func <functype> | ...
 ```
 
-##### Example
+A function reference may be represented by a simple closure, encapsulating a prefix of arguments that have previously been bound. These hidden arguments do not appear in the function type, but will be forwarded implicitly when the function itself is invoked.
+
+
+##### Examples
 
 ```
 type engine = service {
   search : (query : text, callback : func (vec result) -> ());
 }
 ```
+
+```
+type wallet = service {
+  topup : (amount : nat) -> ();
+  forward : (call : () -> ()) -> ();
+}
+```
+In the latter example, the `call` parameter is assumed to be a closure encapsulating a call to another service (including bound arguments) that the wallet executes on its own caller's behalf by invoking the function.
+
 
 #### Principal References
 
@@ -571,9 +583,9 @@ The types of these values are assumed to be known from context, so the syntax do
 <fieldval> ::= <nat> = <annval>
 
 <refval> ::=
-  | service <text>             (canister URI)
-  | func <text> . <name>       (canister URI and message name)
-  | principal <text>           (principal URI)
+  | principal <text>                      (principal URI)
+  | service <text>                        (canister URI)
+  | func <text> . <name> ( <annval>,* )?  (canister URI, message name, and possibly bound arguments)
 
 <arg> ::= ( <annval>,* )
 
@@ -1064,7 +1076,7 @@ Most Candid values are self-explanatory, except for references. There are two fo
 Likewise, there are two forms of Candid values for function references:
 
 * `ref(r)` indicates an opaque reference, understood only by the underlying system.
-* `pub(s,n)`, indicates the public method name `n` of the service referenced by `s`.
+* `pub(s,n,v*:t*)`, indicates the public method name `n` of the service referenced by `s`, possibly followed by a list of type-annotated bound argument values.
 
 #### Notation
 
@@ -1188,11 +1200,15 @@ M : <val> -> <reftype> -> i8*
 M(ref(r) : service <actortype>) = i8(0)
 M(id(v*) : service <actortype>) = i8(1) M(v* : vec nat8)
 
-M(ref(r)   : func <functype>) = i8(0)
-M(pub(s,n) : func <functype>) = i8(1) M(s : service {}) M(n : text)
+M(ref(r)         : func <functype>) = i8(0)
+M(pub(s,n)       : func <functype>) = i8(1) M(s : service {}) M(n : text)
+M(pub(s,n,v*:t*) : func <functype>) = i8(2) M(s : service {}) M(n : text) M*(v* : t*)
 
 M(ref(r) : principal) = i8(0)
 M(id(v*) : principal) = i8(1) M(v* : vec nat8)
+
+M* : <val>* -> <datatype>* -> i8*
+M*(v^N : <datatype>^N) = leb128(N) M(v : <datatype>)^N
 ```
 
 
@@ -1218,10 +1234,13 @@ R((k,v) : k:<datatype>) = R(v : <datatype>)
 R : <val> -> <reftype> -> <ref>*
 R(ref(r) : service <actortype>) = r
 R(id(b*) : service <actortype>) = .
-R(ref(r)   : func <functype>) = r
-R(pub(s,n) : func <functype>) = .
+R(ref(r) : func <functype>) = r
+R(pub(s,n,v*:t*) : func <functype>) = R*(v* : t*)
 R(ref(r) : principal) = r
 R(id(b*) : principal) = .
+
+R* : <val>* -> <datatype>* -> <ref>*
+R*(v^N : <datatype>^N) = R(v : <datatype>)^N
 ```
 
 Note:


matthewhammer:
LGTM!

My only nit is about the other companion PR, and how to compose this PR with that one in a larger motivating example.

Let's say that the wallet call wants to return some data, and we don't know what it is, or even its arity?

How would we adapt this?:

```
type wallet = service {
  topup : (amount : nat) -> ();
  forward : (call : () -> ()) -> ();
}
```

I see three options:
1. `Blob`
2. `dynamic` (see #292)
3. a variant of `dynamic` that I will call `dynamics` (e.g., see comment https://github.com/dfinity/candid/issues/245#issuecomment-959449726)

####  1. Use `Blob`


```
forward : (call : () -> `Blob`) -> `Blob`;
```

And then assume that this `Blob` holds a candid response sequence, not a single candid value.

Pros: `Blob` exists today in Candid and Motoko.
Cons: `Blob` conversions are explicit, and perhaps not ergonomic yet.  They could also have type errors.

#### 2. Use `dynamic` (see #292)

```
forward : (call : () -> `dynamic`) -> `dynamic`;
```

Pros: Could be a safer option to option 1, since conversions would have dynamic checking and auto trapping (I presume).
Cons:  Unfortunately, this API is not quite as "general" as the first, since `dynamic` cannot encode a sequence of values, but only one value.  See option 3.

####  3. Use `dynamics` (see comments to #292)

```
forward : (call : () -> `dynamics`) -> `dynamics`;
```

Pros: Actually solves the problem (I think).
Cons: Furtherest of the options from what we have today.

spec/Candid.md
```diff
@@ -571,9 +583,9 @@ The types of these values are assumed to be known from context, so the syntax do
 <fieldval> ::= <nat> = <annval>
 
 <refval> ::=
-  | service <text>             (canister URI)
-  | func <text> . <name>       (canister URI and message name)
-  | principal <text>           (principal URI)
+  | principal <text>                      (principal URI)
```
chenyan-dfinity, line 586: `principal` is now a primitive type, so we can move this to `<primval>`

spec/Candid.md
```diff
@@ -464,14 +464,26 @@ A *function reference* is described by its function type. For example, they allo
...
 ```
 type engine = service {
   search : (query : text, callback : func (vec result) -> ());
 }
 ```
 
+```
+type wallet = service {
+  topup : (amount : nat) -> ();
+  forward : (call : () -> ()) -> ();
```
chenyan-dfinity, line 481:
```suggestion
  forward : (call : func () -> ()) -> ();
```
syntactically, how do we know if the function has bound arguments? Does the closure supports return type?

spec/Candid.md
```diff
@@ -1188,11 +1200,15 @@ M : <val> -> <reftype> -> i8*
...
-M(pub(s,n) : func <functype>) = i8(1) M(s : service {}) M(n : text)
+M(ref(r)         : func <functype>) = i8(0)
+M(pub(s,n)       : func <functype>) = i8(1) M(s : service {}) M(n : text)
+M(pub(s,n,v*:t*) : func <functype>) = i8(2) M(s : service {}) M(n : text) M*(v* : t*)
 
 M(ref(r) : principal) = i8(0)
 M(id(v*) : principal) = i8(1) M(v* : vec nat8)
+
+M* : <val>* -> <datatype>* -> i8*
+M*(v^N : <datatype>^N) = leb128(N) M(v : <datatype>)^N
```
chenyan-dfinity, line 1211: this will be a breaking change? or the bound argument is stored in `R` which was not used before?

rossberg:
For the time being I think it's fine if this pattern only worked for functions of fixed arity. If we wanted variadic abstraction, maybe we could make a tuple record coercible to a parameter list and vice versa, analogous to what we do in Motoko.

Perhaps n-ary functions as a primitive in Candid were a mistake (as @nomeata always argued, though for different reasons). If we introduced such a coercion, could we retroactively pepper over it?

spec/Candid.md
rossberg: You mean, know it by the type? We don't, it's always allowed. Do you think that would cause issues for bindings?

spec/Candid.md
rossberg: Done.

spec/Candid.md
rossberg: No, R is only used if there are reference types that need it, as before. The type is determined from M.

But I think this is a breaking change in so far that a receiver not understanding the new encoding of closures would choke on it. The only way to avoid this is by indeed introducing a separate closure type as a future type, it seems. Hm, that would be ugly...

spec/Candid.md
chenyan-dfinity: I see. The existing example has a `callback` function, which can have bound arguments as well. You are adding another example to illustrate the use of bound arguments, which makes me think there may be syntactic difference for the bound args.

So any function reference can have bound arguments. The return value will be decoded according to the function signature, right? That means we need an abstraction for argument tuples. And if we use the `dynamic` type as return type, `dynamic` represents argument tuples instead of a single argument.

matthewhammer: > That means we need an abstraction for argument tuples.

Yep.  That's my thinking too, and more agreement in the comments for #292

spec/Candid.md
nomeata, line 1211: Where is the type of `v` encoded? I’d expect something like
```
 M*(v^N : <datatype>^N) = leb128(N) I(<datatype>^N)^N  M(v : <datatype>)^N
```
to match what we do in `B`

nomeata: A new type seems to be appropriate. Not breaking existing stuff is part of our value proposition, and given the complexity of this maybe it’s good if a service can say that they really only support plain function references, but not closures?

spec/Candid.md
```diff
@@ -1064,7 +1076,7 @@ Most Candid values are self-explanatory, except for references. There are two fo
 Likewise, there are two forms of Candid values for function references:
 
 * `ref(r)` indicates an opaque reference, understood only by the underlying system.
-* `pub(s,n)`, indicates the public method name `n` of the service referenced by `s`.
+* `pub(s,n,v*:t*)`, indicates the public method name `n` of the service referenced by `s`, possibly followed by a list of type-annotated bound argument values.
```
nomeata, line 1079: You don’t want to support binding arguments to references?

Why not make it a recursive definition that allows you to bind arguments to any function reference (whether it’s opaque, public, or itself a closure – after all, these are all values of the same type, so binding to them should be allowed).
```suggestion
* `closure(f,v*:t*)`, indicates the function reference `f`, followed by a list of type-annotated bound argument values.
```
```

**Review by nomeata — COMMENTED** [review]
A method call `f(x,y)` now does not just mean “encode `(x,y)` via Candid and send”, because `f` could be a closure. This requires
 * a heap representation of a Candid closure `f`, consisting of (likely)
   - service reference
   - method name 
   - original type table from the message that carried `f` (pruned to the actual used types? copied as is? Probably pruning is not possible if the bound values are of a future type we don’t understand)
   - the candid encoded value, kept as an opaque blob (no need to decode, and because of future types we probably can’t)
 * the ability to _merge_ that type table with the type table we’d already generate for the explicitly passed `(x,y)`

That’s a quite high “implementation complexity price” to be paid, so I am overall quite wairy of this.

**Comment by rossberg** [comment]
@nomeata, fair points about constructing the new call.

I would assume that the serialiser would not merge but merely extend the type table it gets from the closure. Since we allow duplication, it could do so blindly.

For the argument tuple, if it extended the type table from the closure as just said, then I agree that it suffices to just copy over the serialised blob for each value. (We could even add an leb128 with the length for each value's encoding to the `TM` function I now introduced; however, I'm not sure we want that – for security reasons, the deserialiser should validate the data anyway to the extend it can, and could separate the individual values right there.)

So yes, some work is required, but it doesn't seem too bad?

(With hindsight, I actually think it was an oversight that our function types did not allow closures from the beginning. It seems like a glaring omission for a higher-order data format.)

**Review by nomeata — COMMENTED** [review]
I see, if you put the type table from the closure first, you only have to renumber the type indices you are appending. Still more work than now (where the full type table and thus these indices  are pre-computed by the Motoko compiler.)

I still see a problem where you need to pass a  closure as an argument to a closure, because now you need to merge and renumber after all. And that's not possible: our future types prevent _any_ kind of operations on type tables or indices.

So bound arguments need their own type table, both in the closure, and in the final call.

So that leads to the nicely simple design where the bound arguments in a closure are simply complete argument sequences with their own type table (i.e. `B(args)`, and we allow the concatenation of encoded argument sequences to represent the encoding of the concatenation (yay, distributivity laws :-)).

Overall, it seems that this feature has very narrow use cases on our platform, because of the prevalent authentication via `caller`. Canister can really only ever call closures received from fully trusted users. Anything more interesting likely need system-level closures. Maybe worth satisfying the proxy-from-ttusted-user use case without candid support (raw calls), and push for system level closures instead (which then may not even need changes to Candid, since these would be `ref`s)?

File: spec/Candid.md [hunk 5]
```diff
@@ -987,6 +1004,12 @@ C[service <actortype> <: service <actortype'>](service <text>) = service <text>
 C[principal <: principal](principal <text>) = principal <text>
 ```
 
+However, functions can be converted into closures with an empty list of bound arguments:
+```
+C[func <functype> <: closure <functype'>](func <text>.<id>) = clos(func <text>.<id>, .)
```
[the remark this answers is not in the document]
rossberg: Ah, yes.

spec/Candid.md
```diff
@@ -1200,15 +1231,21 @@ M : <val> -> <reftype> -> i8*
...
+  leb128(|R(f : func <functype>) R*(v* : t*)|)
+  i8(2) M(f : func <functype>) TM*(v* : t*)
 
 M(ref(r) : principal) = i8(0)
 M(id(v*) : principal) = i8(1) M(v* : vec nat8)
 
-M* : <val>* -> <datatype>* -> i8*
-M*(v^N : <datatype>^N) = leb128(N) M(v : <datatype>)^N
+TM : <val> -> <datatype> -> i8*
+TM(v : <datatype>) = T(<datatype>) M(v : <datatype>)
```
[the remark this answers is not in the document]
rossberg: Good point, changed.

spec/Candid.md
```diff
@@ -1131,7 +1159,10 @@ T : <reftype> -> i8*
 T(func (<datatype1>*) -> (<datatype2>*) <funcann>*) =
   sleb128(-22) T*(<datatype1>*) T*(<datatype2>*) T*(<funcann>*) // 0x6a
 T(service {<methtype>*}) =
-  sleb128(-23) T*(<methtype>*)                                    // 0x69
+  sleb128(-23) T*(<methtype>*)                                  // 0x69
+T(closure (<datatype1>*) -> (<datatype2>*) <funcann>*) =
```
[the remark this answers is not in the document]
rossberg: Done.

spec/Candid.md
```diff
@@ -1200,15 +1231,21 @@ M : <val> -> <reftype> -> i8*
...
 
-M(ref(r)         : func <functype>) = i8(0)
-M(pub(s,n)       : func <functype>) = i8(1) M(s : service {}) M(n : text)
-M(pub(s,n,v*:t*) : func <functype>) = i8(2) M(s : service {}) M(n : text) M*(v* : t*)
+M(ref(r)        : func <functype>) = i8(0)
+M(pub(s,n)      : func <functype>) = i8(1) M(s : service {}) M(n : text)
+M(clos(f,v*:t*) : closure <functype>) =
+  leb128(|i8(2) M(f : func <functype>) TM*(v* : t*)|)
+  leb128(|R(f : func <functype>) R*(v* : t*)|)
+  i8(2) M(f : func <functype>) TM*(v* : t*)
```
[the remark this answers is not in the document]
rossberg: Done'

rossberg (commented):
> So bound arguments need their own type table, both in the closure, and in the final call.
>
> So that leads to the nicely simple design where the bound arguments in a closure are simply complete argument sequences with their own type table (i.e. B(args), and we allow the concatenation of encoded argument sequences to represent the encoding of the concatenation (yay, distributivity laws :-)).

Interesting point. But that would require a backwards-incompatible change to the encoding of calls, wouldn't it?

I suppose we could circumvent all this by not allowing currying but only complete binding, i.e., thunks instead of partially applied closures. But that perhaps is a bit too specialised...

spec/Candid.md
```diff
@@ -1322,9 +1324,17 @@ Deserialisation at an expected type sequence `(<t'>,*)` proceeds by
...
 * A serialised type may be headed by an other than -1 to -24 . Any such opcode is followed by an LEB128-encoded count, and then a number of bytes corresponding to this count. A type represented that way is called a *future type*.
+  ```
+  FT : i32 -> i8* -> i8*
+  FT(n, b*) = sleb128(n) leb128(|b*|) b*
+  ```
+
+* A value corresponding to a future type is called a *future value*. It is represented by two LEB128-encoded counts, *m* and *n*, followed by a *m* bytes in the memory representation M and accompanied by *n* corresponding references in R.
+  ```
+  FM : i8* -> ref* -> i8*
+  FM(b*, r*) = leb128(|b*|) leb128(r*) b*
```
nomeata, line 1336:
```suggestion
  FM(b*, r*) = leb128(|b*|) leb128(|r*|) b*
```

nomeata:
> But that would require a backwards-incompatible change to the encoding of calls, wouldn't it?

Oh, right, of course. But that means we have an impossibly result: we can't merge type tables (because of future types), but we also can't keep them separate (because it wouldn't be backwards compatible). This seems to prevent currying…

nomeata:
(For some reason I can't mark conversations as resolved here)

spec/Candid.md
```diff
@@ -1188,12 +1231,24 @@ M : <val> -> <reftype> -> i8*
...
 M(ref(r) : principal) = i8(0)
 M(id(v*) : principal) = i8(1) M(v* : vec nat8)
+
+TM : <val> -> <datatype> -> i8*
+TM(v : <datatype>) = I(<datatype>) M(v : <datatype>)
+
+TM* : <val>* -> <datatype>* -> i8*
+TM*(v^N : <datatype>^N) = leb128(N) TM(v : <datatype>)^N
 ```
+The meta-function `FM` constructs a backwards-compatible encoding for values of for [future types](#deserialisation-of-future-types).
```
crusso, line 1251:
```suggestion
The meta-function `FM` constructs a backwards-compatible encoding for values [future types](#deserialisation-of-future-types).
```

rossberg:
```suggestion
The meta-function `FM` constructs a backwards-compatible encoding for values of [future types](#deserialisation-of-future-types).
```

PR #297, opened 2021-11-22
Add parser fuzzer

guidovranken:
**Overview**
> Why do we need this feature? What are we trying to accomplish?

For finding bugs and vulnerabilities in candid.

**Requirements**
> What requirements are necessary to consider this problem solved?

This cargo fuzz-compatible fuzzing harness is required to fuzz candid.

**Considered Solutions**
> What solutions were considered?

**Recommended Solution**
> What solution are you recommending? Why?

This fuzzer integration was requested by Dfinity.

**Considerations**
> What impact will this change have on security, performance, users (e.g. breaking changes) or the team (e.g. maintenance costs)?

Maintenance: the fuzzer must constitute a valid program, and if changes to the candid API are made, it must be ensured that the fuzzer still compiles, and if not, the fuzzer code must be adapted to the new API.

### DIFF
--- a/rust/candid/fuzz/.gitignore
+++ b/rust/candid/fuzz/.gitignore
@@ -1,0 +1,3 @@
+target
+corpus
+artifacts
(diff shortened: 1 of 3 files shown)


chenyan-dfinity (approved):
A good start :)

PR #301, opened 2021-12-12
Update incorrect URL for Candid UI 

christyjacob4:
**Overview**
Updated the incorrect URL

**Requirements**
NA

**Considered Solutions**
NA

**Recommended Solution**
NA

**Considerations**
NA

### DIFF
--- a/docs/modules/candid-guide/pages/candid-howto.adoc
+++ b/docs/modules/candid-guide/pages/candid-howto.adoc
@@ -83,7 +83,7 @@ By default, the `+local+` canister execution environment binds to the `+127.0.0.
 For example, the full URL should look similar to the following but with the `+CANDID-UI-CANISTER-IDENTIFIER+` that was returned by the `+dfx canister id+` command:
 +
 ....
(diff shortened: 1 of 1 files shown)


docs/modules/candid-guide/pages/candid-howto.adoc
```diff
@@ -83,7 +83,7 @@ By default, the `+local+` canister execution environment binds to the `+127.0.0.
 For example, the full URL should look similar to the following but with the `+CANDID-UI-CANISTER-IDENTIFIER+` that was returned by the `+dfx canister id+` command:
 +
 ....
-http://127.0.0.1:8000/candid?canisterId=<CANDID-UI-CANISTER-IDENTIFIER>
+http://127.0.0.1:8000/?canisterId=<CANDID-UI-CANISTER-IDENTIFIER>
```
christyjacob4, line 86: The old URL seems to be broken and does not work with the latest version of Dfinity ( 0.8.4 )
I've updated it to use the link that works

chenyan-dfinity (approved):

PR #311, opened 2022-01-12
spec: subtype checking only for reference types

### LINKED ISSUE #295: Extensible variants
When we designed the solution for extensible records, with special rules for `opt` so that they can be used in record fields, we thought we had also solved it dually for variants, as long as the variant itself is wrapped in `opt`.

And that was certainly true in the iteration that did a dynamic value-based check: if the tag was known, decoding would work; if it was a new, unknown tag, it would fail and turn to `null`.

But later (for good reason) we switched to a type-based check, decoding as `null` any value if the type isn't a subtype of the expected type. But in the record extension case, the extended variant _isn't_ a subtype of the expected type, so now all values map to `null`. Which is not what we wanted, I believe.

This just came up in potential practice in https://forum.dfinity.org/t/enable-canisters-to-hold-icp/6153/167?u=nomeata

Not sure what to do about it, though. Besides maybe some `variant`-specific solution, e.g. some way to explicitly or implicitly declare a nullary tag as the default.

chenyan-dfinity:
Fix #295

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -909,7 +909,7 @@ Coercion into `reserved` is the constant map (this is arbitrarily using `null` a
 _ : <t> ~> null : reserved
 ```
 
-NB: No rule is needed for type `empty`, because there are no values of that type. By construction, `<v> : <t> ~> _ : empty` will not hold.
+NB: No rule is needed for type `empty`, because there are no values of that type. By construction, `_ : empty ~> _ : _` will not hold.
 
 #### Vectors
 
@@ -939,7 +939,7 @@ not (<v> : <t> ~> _ : <t'>)
 opt <v> : opt <t> ~> null : opt <t'>
 ```
 
-Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it has a suitable type:
+Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it can be decoded successfully:
 ```
 not (null <: <t'>)
 <v> : <t> ~> <v'> : <t'>
@@ -991,7 +991,7 @@ variant { <nat> = <v'> } : variant { <nat> : <t'>; _;* }
 
 #### References
 
-Function and services reference values are untyped, so the coercion function is the identity here:
+For function and services reference values, the coercion relation checks whether the given types are actually subtypes of the expected type, and fails else:
 
 ```
 func <functype> <: func <functype'>
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -994,7 +994,7 @@ variant { <nat> = <v'> } : variant { <nat> : <t'>; _;* }
 
 #### References
 
-For function and services reference values, the coercion relation checks whether the given types are actually subtypes of the expected type, and fails else:
+For function and services reference values, the coercion relation checks whether the given types are actually subtypes of the expected type, and fails otherwise:
 
 ```
 func <functype> <: func <functype'>
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -1,8 +1,8 @@
 # Candid Specification
 
-Version: 0.1.3
+Version: 0.1.4
 
-Date: February 3, 2021
+Date: January 11, 2022
 
 ## Motivation
 
@@ -885,62 +885,60 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
 
 This subtyping is implemented during the deserialisation of Candid at an expected type. As described in [Section Deserialisation](#deserialisation), the binary value is conceptually first _decoded_ into the actual type and a value of that type, and then that value is _coerced_ into the expected type.
 
-To model this, we define, for every `t1, t2` with `t1 <: t2`, a function `C[t1<:t2] : t1 -> t2`. This function maps values of type `t1` to values of type `t2`, and is indeed total.
+To model this, we define a partial coercion function `C[t1 ~> t2] : t1 -> t2`. This function maps values of type `t1` to values of type `t2`. For every `t1, t2` with `t1 <: t2`, the function is always defined, but not every `t1, t2` has a coercion.
 
-to describe these values, we re-use the syntax of the textual representation, and use the the `<annval>` syntax (i.e. `(v : t)`) if necessary to resolve overloading.
+To describe these values, we re-use the syntax of the textual representation, and use the the `<annval>` syntax (i.e. `(v : t)`) if necessary to resolve overloading.
 
 #### Primitive Types
 
 On primitve types, coercion is the identity:
 ```
-C[<t> <: <t>](x) = x    for every <t> ∈ <numtype>, bool, text, null
+C[<t> ~> <t>](x) = x     for every <t> ∈ <numtype>, bool, text, null
 ```
 
 Values of type `nat` coerce at type `int`:
 ```
-C[nat <: int](<nat>) = <nat>
+C[nat ~> int](<nat>) = <nat>
 ```
 
 Coercion into `reserved` is the constant map (this is arbitrarily using `null` as “the” value of `reserved`):
 ```
-C[<t> <: reserved](_) = null
+C[<t> ~> reserved](_) = null
 ```
-NB: No rule is needed for type `empty`, because there are no values of that type. By construction, `C[empty <: <t>]` is the unique function on the empty domain.
+NB: No rule is needed for type `empty`, because there are no values of that type. By construction, `C[empty ~> <t>]` is the unique function on the empty domain.
 
 #### Vectors
 
 Vectors coerce pointwise:
 ```
-C[vec <t> <: vec <t'>]( vec { <v>;* } ) = vec { C[<t> <: <t'>](<v>);* }
+C[vec <t> ~> vec <t'>]( vec { <v>;* } ) = vec { C[<t> ~> <t'>](<v>);* }
 ```
 
 #### Options
 
 The null type and the reserved type coerce into any option type:
 ```
-C[null <: opt <t>](null) = null
+C[null ~> opt <t>](null) = null
 ```
 
 An optional value coerces at an option type, if the constituent value has a suitable type, and else goes to `null`:
 ```
-C[opt <t> <: opt <t'>](null) = null
-C[opt <t> <: opt <t'>](opt <v>) = opt C[<t> <: <t'>](v)  if <t> <: <t'>
-C[opt <t> <: opt <t'>](opt <v>) = null                   if not(<t> <: <t'>)
+C[opt <t> ~> opt <t'>](null) = null
+C[opt <t> ~> opt <t'>](opt <v>) = opt C[t ~> t'](v)  if t ~> t'
+C[opt <t> ~> opt <t'>](opt <v>) = null               if not (t ~> t')
 ```
 
 Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it has a suitable type:
 ```
-C[<t> <: opt <t'>](<v>) = opt C[<t> <: <t'>](v)  if not (null <: <t'>) and <t> <: <t'>
+C[<t> ~> opt <t'>](<v>) = opt C[<t> ~> <t'>](v)  if not (null <: <t'>) and <t> ~> <t'>
 ```
 
 Any other type goes to `null`:
 ```
-C[reserved <: opt <t>](_) = null
-C[<t> <: opt <t'>](_) = null  if not (null <: <t'>) and not (<t> <: <t'>)
-C[<t> <: opt <t'>](_) = null  if null <: <t'> and not (null <: <t>)
+C[reserved ~> opt <t>](_) = null
+C[<t> ~> opt <t'>](_) = null  if not (null <: <t'>) and not (<t> ~> <t'>)
+C[<t> ~> opt <t'>](_) = null  if null <: <t'> and not (null ~> <t>)
 ```
-
-NOTE: These rules above imply that a Candid decoder has to be able to decide the subtyping relation at runtime.
 
 #### Records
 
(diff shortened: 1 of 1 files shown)


spec/Candid.md
```diff
@@ -885,63 +885,61 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
 
 This subtyping is implemented during the deserialisation of Candid at an expected type. As described in [Section Deserialisation](#deserialisation), the binary value is conceptually first _decoded_ into the actual type and a value of that type, and then that value is _coerced_ into the expected type.
 
-To model this, we define, for every `t1, t2` with `t1 <: t2`, a function `C[t1<:t2] : t1 -> t2`. This function maps values of type `t1` to values of type `t2`, and is indeed total.
+To model this, we define a partial coercion function `C[t1 ~> t2] : t1 -> t2`. This function maps values of type `t1` to values of type `t2`. For every `t1, t2` with `t1 <: t2`, the function is always defined, but not every `t1, t2` has a coercion.
```
nomeata, line 888: I don’t think I would model it that way, but more like we had before (before 6e62bbf277ffbe32dfadb20e7e80814c0772dba6), with the universal set of values `V` as the input, and the only change being that in `V`, references values have type annotations.

But I can’t quite put my finger on _why_ that’s better, so maybe carry on if this doesn’t match your gut feeling.

spec/Candid.md
```diff
@@ -885,63 +885,61 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 ```
 
 An optional value coerces at an option type, if the constituent value has a suitable type, and else goes to `null`:
 ```
-C[opt <t> <: opt <t'>](null) = null
-C[opt <t> <: opt <t'>](opt <v>) = opt C[<t> <: <t'>](v)  if <t> <: <t'>
-C[opt <t> <: opt <t'>](opt <v>) = null                   if not(<t> <: <t'>)
+C[opt <t> ~> opt <t'>](null) = null
+C[opt <t> ~> opt <t'>](opt <v>) = opt C[t ~> t'](v)  if t ~> t'
+C[opt <t> ~> opt <t'>](opt <v>) = null               if not (t ~> t')
```
nomeata, line 928: This doesn't quite work; we really need to check if `C[t ~> t'](v)` (note the `v` matters!) is defined. So maybe
```suggestion
C[opt <t> ~> opt <t'>](opt <v>) = v'   if C[t ~> t'](v) = v'
C[opt <t> ~> opt <t'>](opt <v>) = null if C[t ~> t'](v) is not defined.
```

spec/Candid.md
nomeata: Also `t1 ~> t2` isn’t really a meaningful relation. There is just a partial function `C` with inputs `v : t1` and `t2`. So maybe
```
C(v : t1, t2) : t2
```
would work.

spec/Candid.md
```diff
@@ -885,63 +885,61 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 ```
 
 Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it has a suitable type:
 ```
-C[<t> <: opt <t'>](<v>) = opt C[<t> <: <t'>](v)  if not (null <: <t'>) and <t> <: <t'>
+C[<t> ~> opt <t'>](<v>) = opt C[<t> ~> <t'>](v)  if not (null <: <t'>) and <t> ~> <t'>
```
nomeata, line 933:
```suggestion
C[<t> ~> opt <t'>](<v>) = v' if not (null <: <t'>) and opt C[<t> ~> <t'>](v) = v'
```

spec/Candid.md
```diff
@@ -885,63 +885,61 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 ```
 
 Any other type goes to `null`:
```
nomeata, line 936: We are now analysing values, not types
```suggestion
Any other value goes to `null`:
```

spec/Candid.md
```diff
@@ -885,63 +885,61 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 ```
-C[reserved <: opt <t>](_) = null
-C[<t> <: opt <t'>](_) = null  if not (null <: <t'>) and not (<t> <: <t'>)
-C[<t> <: opt <t'>](_) = null  if null <: <t'> and not (null <: <t>)
+C[reserved ~> opt <t>](_) = null
+C[<t> ~> opt <t'>](_) = null  if not (null <: <t'>) and not (<t> ~> <t'>)
+C[<t> ~> opt <t'>](_) = null  if null <: <t'> and not (null ~> <t>)
```
nomeata, line 940:
```suggestion
C[reserved ~> opt <t>](_) = null
C[<t> ~> opt <t'>](v) = null  if not (null <: <t'>) and C[<t> ~> <t'>](v) is undefined
C[<t> ~> opt <t'>](_) = null  if null <: <t'> and and not (null <: <t>)
```
(I think we need the last line as as, but I’m not sure, have to think more about it. Note that `null <: <t>` is just fancy pants for “an opt-like type”, and does not require a subtype check, just a peek at `<t>`).

spec/Candid.md
```diff
@@ -1001,7 +1001,7 @@ The relations above have certain properties. As in the previous section, `<v> :
 
 * Roundtripping:
   ```
-  (<v> : <t>) ⟺ C[<t> <: <t>](<v>) = <v>
+  (<v> : <t>) ⟺ C[<t> ~> <t>](<v>) = <v>
   ```
 
 * Soundness of subtyping (or, alternatively, well-definedness of coercion)
```
nomeata, line 1007: This needs to be now split into two properties:

* Well-typedness:
  ```
  ∀ <v> : <t>. C[<t> <: <t'>](<v>) : <t'> (if defined)
  ```
* Totality (on subtypes):
  ```
  <t> <: <t'> ⇒ ∀ <v> : <t> C[<t> <: <t'>](<v>) is defined
  ```

spec/Candid.md
chenyan-dfinity: I think they are equivalent, but you are right that inference rules are more natural here.

spec/Candid.md
```diff
@@ -1024,12 +1058,12 @@ The relations above have certain properties. As in the previous section, `<v> :
   ```
   does not imply
   ```
-  C[<t1> <: <t3>] = C[<t2> <: <t3>] ⚬ C[<t1> <: <t2>]
+  C[<t1> ~> <t3>] = C[<t2> ~> <t3>] ⚬ C[<t1> ~> <t2>]
```
chenyan-dfinity, line 1061: How to formulate this part?

spec/Candid.md
```diff
@@ -885,62 +885,81 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 
 Coercion into `reserved` is the constant map (this is arbitrarily using `null` as “the” value of `reserved`):
 ```
-C[<t> <: reserved](_) = null
+--------------------------
+_ : <t> ~> null : reserved
 ```
-NB: No rule is needed for type `empty`, because there are no values of that type. By construction, `C[empty <: <t>]` is the unique function on the empty domain.
+
+NB: No rule is needed for type `empty`, because there are no values of that type. By construction, `<v> : <t> ~> _ : empty` will not hold.
```
nomeata, line 912:
```suggestion
NB: No rule is needed for type `empty`, because there are no values of that type. By construction, `_ : empty ~> _ : _` will not hold.
```

spec/Candid.md
```diff
@@ -885,62 +885,81 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 
 #### Options
 
-The null type and the reserved type coerce into any option type:
+The null value coerces into any option type:
 ```
-C[null <: opt <t>](null) = null
+-----------------------------
+null : _ ~> null : opt <t>
 ```
```
nomeata, line 929: This would allow `null : reserved ~> null : opt <t>`. I guess that’s ok, we want that to hold, although previously we had a dedicated rule for `reserved`.

It might still be easier to go through the input types separately (`null`, `opt t`, `reserved`, other `t`) with their own rules.

spec/Candid.md
```diff
@@ -885,62 +885,81 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
+<v> : <t> ~> <v'> : <t'>
+----------------------------------------
+opt <v> : opt <t> ~> opt <v'> : opt <t'>
+
+not (<v> : <t> ~> _ : <t'>)
+----------------------------------------
+opt <v> : opt <t> ~> null : opt <t'>
 ```
 
 Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it has a suitable type:
```
nomeata, line 942:
```suggestion
Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it can be decoded successfully:
```
(It’s no longer just the type that matters.)

spec/Candid.md
```diff
@@ -885,62 +885,81 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
+not (null <: <t'>)
+not (<v> : <t> ~> _ : <t'>)
+--------------------------------
+<v> : <t> ~> null : opt <t'>
+
+null <: <t'>
+not (null <: <t>)
+----------------------------
+<v> : <t> ~> null : opt <t'>
+```
```
nomeata, line 962: Do we need a rule for `reserved`? I guess that depends on whether the type `reserved` has more than just the value `null`, and that is a philosophical debate that we better avoid here (in the wire format it has only one value, but in a semantic model maybe not). So to avoid the discussion, maybe explicitly say that
```
 <v> : reserved ~> null : opt <t'>
```

spec/Candid.md
```diff
@@ -970,18 +994,28 @@ C[variant { <nat> = <t>; _;* } <: variant { <nat> = <t'>; _;* }](variant { <nat>
 Function and services reference values are untyped, so the coercion function is the identity here:
```
nomeata, line 994: No longer true!
```suggestion
For function and services reference values, the coercion relation checks whether the given types are actually subtypes of the expected type, and fails else:
```

spec/Candid.md
```diff
@@ -1001,12 +1035,12 @@ The relations above have certain properties. As in the previous section, `<v> :
 
 * Roundtripping:
   ```
-  (<v> : <t>) ⟺ C[<t> <: <t>](<v>) = <v>
+  (<v> : <t>) ⟺ <v> : <t> ~> <v> : <t>
   ```
```
nomeata, line 1040:
````suggestion

* Well-typedness
  ```
  <v> : <t> ~> <v'> : <t'> ⇒ <v'> <: <t'>
  ```

````

It remains to be seen if such theorems need `<v> : <t>` as an assumption. A technicality, but still. Maybe we should state explicit that the relation is only defined on well-typed input. Or maybe it holds by construction.

File: spec/Candid.md (hunk 7 above)
Comment (nomeata) [reply spec/Candid.md:1061]: We can actually drop the assumption about subtyping, and state

NB: Transitive coherence does not hold:
```
<v1> : <t1> ~> <v2> : <t2>,  <v2> : <t2> ~> <v3> : <t3>, 
```
and
```
<v1> : <t1> ~> <v3'> : <t3> 
```
do not imply `<v3> = <v3'>`.

However, it implies `<v3> ~ <v3'>` is the smallest homomorphic, reflexive, symmetric relation that satisfies `∀ v. opt v ~ null`.

File: spec/Candid.md [hunk 14]
```diff
@@ -924,8 +924,9 @@ vec { <v>;* } : vec <t> ~> vec { <v'>;* } : vec <t'>
 
 The null value coerces into any option type:
 ```
+null <: <t>
 -----------------------------
-null : _ ~> null : opt <t>
+null : <t> ~> null : opt <t'>
```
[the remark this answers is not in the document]
chenyan-dfinity: I think it's fine. We use `null <: <t>` to represent `null`, `reserved`, `opt _` in many places.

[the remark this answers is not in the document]
nomeata: That's `null <: <t>`, the type, not `null : <t>`, the value.

The rule as written by you would accept ill-typed input like `null : int`. It's clear that this is not what we mean, but it doesn't hurt to be precise.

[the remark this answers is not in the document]
chenyan-dfinity: The premise is `null <: <t>`, so `<t>` can only be `null`, `reserved` or `opt _`. It cannot be `int`.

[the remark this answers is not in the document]
nomeata: Ah, I didn't notice that. Sorry, my bad. This works of course.

nomeata:
#316 has early explorations of updating the Coq proof for this, but stuck at various technicalities right now.

spec/Candid.md
```diff
@@ -951,37 +973,52 @@ In the following rule:
...
+----------------------------------------------------------
+variant { <nat> = <v>; } : variant { <nat> : <t>; _;* } ~>
+variant { <nat> = <v'> } : variant { <nat> : <t'>; _;* }
 ```
 
 
 #### References
 
-Function and services reference values are untyped, so the coercion function is the identity here:
+For function and services reference values, the coercion relation checks whether the given types are actually subtypes of the expected type, and fails else:
```
crusso, line 997:
```suggestion
For function and services reference values, the coercion relation checks whether the given types are actually subtypes of the expected type, and fails otherwise:
```

spec/Candid.md
```diff
@@ -885,62 +885,84 @@ service { <name> : <functype>; <methtype>;* } <: service { <name> : <functype'>;
...
 
-To model this, we define, for every `t1, t2` with `t1 <: t2`, a function `C[t1<:t2] : t1 -> t2`. This function maps values of type `t1` to values of type `t2`, and is indeed total.
-
-to describe these values, we re-use the syntax of the textual representation, and use the the `<annval>` syntax (i.e. `(v : t)`) if necessary to resolve overloading.
+This section describes the coercion, as a relation `V : T ~> V' : T'` to describe when a value `V` of type `T` can be coerced to a value `V'` of type `T'`. The fields `V`, `T` and `T'` can be understood as inputs and `V'` as the output of this relation.
+To describe these values, we re-use the syntax of the textual representation.
 
 #### Primitive Types
 
 On primitve types, coercion is the identity:
```
crusso, line 893:
```suggestion
On primitive types, coercion is the identity:
```

nomeata:
Oh, this hasn't landed yet? Completely forgot… I guess we are waiting for all implementations to catch up?

crusso:
> Oh, this hasn't landed yet? Completely forgot… I guess we are waiting for all implementations to catch up?

@chenyan-dfinity is working on it. Motoko is hopefully ready to go (though I'm worried even I don't understand that implementation any more).

nomeata:
🥳 

I have released haskell-candid 0.4 with the corresponding changes. Maybe someone can help me with https://github.com/dfinity/ic-hs/pull/118?

PR #322, opened 2022-03-08
Make encoding functions public

seniorjoinu:
**Overview**
Many APIs of candid encoding are private. This makes it harder to use this library in case of unusual scenarios implementation.
In one of my projects (ic-event-hub) I need to be able to transfer a (possibly very big) batch of objects between canisters in as few messages as possible, as efficiently as possible. I can achieve that using `ValueSerializer` and `TypeSerialize` structs, filling messages with objects sequentially. But in order to do that, I'm using a custom `candid` fork where all methods of these structs are public.

Here is the PR with some of these methods made public (enough to achieve what I want). But I strongly recommend the team managing this repo to choose more open policy in terms of method visibility and to make everything public. Yes, this will increase the risk of error making for beginners. But for experienced developers who understand the internals this is actually a lot more useful, because it allows them to program custom serialization logic on top of this library.

**Considerations**
This PR doesn't affect performance or security, but it could be useful to mark affected methods as `FOR ADVANCED USE ONLY` in the docs.

### DIFF
--- a/rust/candid/src/ser.rs
+++ b/rust/candid/src/ser.rs
@@ -83,10 +83,14 @@ impl ValueSerializer {
     pub fn get_result(&self) -> &[u8] {
         &self.value
     }
+
+    #[doc(hidden)]
     pub fn write_leb128(&mut self, value: u64) -> Result<()> {
         leb128_encode(&mut self.value, value)?;
         Ok(())
     }
+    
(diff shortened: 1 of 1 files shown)


seniorjoinu:
Done @chenyan-dfinity

chenyan-dfinity (approved):

PR #323, opened 2022-03-18
revise existing tests to match #311 (subtype checking only for refere…

crusso:
…nce types)

I think these tests are need to change in light of #311 (as verified by Motoko testing)

### DIFF
--- a/test/construct.test.did
+++ b/test/construct.test.did
@@ -54,7 +54,7 @@ assert blob "DIDL\00\00" == "(null)"                                 : (Opt) "op
 
 // vector
 assert blob "DIDL\01\6d\7c\01\00\00" == "(vec {})"                          : (vec int) "vec: empty";
-assert blob "DIDL\01\6d\7c\01\00\00"                                       !: (vec int8) "vec: non subtype empty";
+assert blob "DIDL\01\6d\7c\01\00\00"                                        : (vec int8) "vec: non subtype empty";
 assert blob "DIDL\01\6d\7c\01\00\02\01\02" == "(vec { 1; 2 })"              : (vec int) "vec";
 assert blob "DIDL\01\6d\7b\01\00\02\01\02" == "(blob \"\\01\\02\")"         : (vec nat8) "vec: blob";
 assert blob "DIDL\01\6d\00\01\00\00" == "(vec {})"                          : (Vec) "vec: recursive vector";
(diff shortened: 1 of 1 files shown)


nomeata:
I guess we should implement the new rules in Haskell and Rust to compare notes on these updated tests.

crusso:
The fact that no-one is confident about how this is meant to behave strikes me as slightly worrying ;->

nomeata:
Just because it changes every few months doesn’t mean we are not confident? ;-)

crusso:
@nomeata thanks for adjusting the tests to use a method name - most of my tests now pass but there's another bug I'll need to figure out on the motoko side:

test/construct.test.did
```diff
@@ -139,10 +139,10 @@ assert "(variant {})"                                                          !
...
 assert blob "DIDL\01\6b\01\00\7f\01\00\00\2a"                                  !: (variant {0:int}) "variant: type mismatch";
-assert blob "DIDL\01\6b\02\00\7f\01\7c\01\00\01\2a"                            !: (variant {0:int; 1:int}) "variant: type mismatch in unused tag";
+assert blob "DIDL\01\6b\02\00\7f\01\7c\01\00\01\2a"                            : (variant {0:int; 1:int}) "variant: type mismatch in unused tag";
 assert blob "DIDL\01\6b\01\00\7f\01\00\00" == "(variant {0})"                   : (variant {0;1}) "variant: ignore field";
-assert blob "DIDL\01\6b\02\00\7f\01\7f\01\00\00"                               !: (variant {0}) "variant {0;1} !<: variant {0}";
-assert blob "DIDL\01\6b\02\00\7f\01\7f\01\00\00" == "(null)"                    : (opt variant {0}) "variant {0;1} <: opt variant {0}";
+assert blob "DIDL\01\6b\02\00\7f\01\7f\01\00\00"                               : (variant {0}) "variant {0;1} <: variant {0}";
+assert blob "DIDL\01\6b\02\00\7f\01\7f\01\00\00" == "(variant {0})"            : (opt variant {0}) "variant {0;1} <: opt variant {0}";
```
chenyan-dfinity, line 145:
```suggestion
assert blob "DIDL\01\6b\02\00\7f\01\7f\01\00\00" == "(opt variant {0})"        : (opt variant {0}) "variant {0;1} <: opt variant {0}";
```

test/subtypes.test.did
```diff
@@ -0,0 +1,164 @@
...
+   == "(null)" : (opt func () -> (variant {0 : nat})) "variant {0 : bool} </: variant {0 : nat}";
+assert blob "DIDL\02\6a\00\01\01\00\6b\01\00\7e\01\00\01\01\00\01m"
+   == "(null)" : (opt func () -> (variant {1 : bool})) "variant {0 : bool} </: variant {1 : bool}";
+
+// infinite types
+
+assert blob "DIDL\02\6a\00\01\01\00\6c\01\00\01\01\00\01\01\00\01m"
+   == "(opt func \"aaaaa-aa\".m)" : (opt func () -> (EmptyRecord)) "(µ record) <: (µ record)";
+assert blob "DIDL\02\6a\00\01\01\00\6c\01\00\01\01\00\01\01\00\01m"
+   == "(null)" : (opt func () -> (empty)) "(µ record) </: empty";
```
chenyan-dfinity, line 127:
```suggestion
// assert blob "DIDL\02\6a\00\01\01\00\6c\01\00\01\01\00\01\01\00\01m"
//   == "(null)" : (opt func () -> (empty)) "(µ record) </: empty";
```

I will comment out this test for now. It's a bit implementation dependent. In Rust, we replace `µ record` with `empty` after parsing the type table, so we actually get `opt func` instead of `null`.

chenyan-dfinity (approved):

test/subtypes.test.did
crusso: Can you explain why? I don't quite follow.

test/subtypes.test.did
```diff
@@ -0,0 +1,164 @@
...
+   == "(opt func \"aaaaa-aa\".m)" : (opt func () -> (Vec)) "vec {µ vec} <: (µ vec)";
+
+// future types
+// This uses 0x67 for the “future type”; bump (well, decrement) once that
+// becomes a concrete future type
+
+assert blob "DIDL\02\6a\00\01\01\00\67\00\01\00\01\01\00\01m"
+   == "(opt func \"aaaaa-aa\".m)" : (opt func () -> (opt empty)) "(future type) <: (opt empty)";
+assert blob "DIDL\02\6a\00\01\01\00\67\00\01\00\01\01\00\01m"
+   == "(null)" : (opt func () -> (nat)) "(future type) <: (nat)";
```
chenyan-dfinity, line 164:
```suggestion
   == "(null)" : (opt func () -> (nat)) "(future type) </: (nat)";
```

test/subtypes.test.did
nomeata: In the Haskell code I think I too replace `mu record` with empty, but only after subtype checking?

I believe Motoko gets this right as spec'ed.

If we want to allow implementations to differ here we should say so explicitly in the spec. But I don't know if that's a good idea.

Could you introduce in the rust code a `MuRecord` type that you can replace it with, which does the job of preventing your code to run into these vicious cycles, while still having the right subtypes relations?

Hmm, but `mu record <: 
mu (record opt)`, so you probably really can't throw away the structure before subtype checking.

nomeata: It’s safe, but it is not according to specification. If we go that route, it would be consistent to equate all other empty types as well (`empty`, `{empty, empty}` etc.), and it should be part of the spec.

PR #327, opened 2022-03-25
fix: `deserialize_ignored_any` should be untyped

Daniel-Bloom-dfinity:
And some tests as well.

**Overview**
Prevents `IDLValueVisitor` from hitting [this panic](https://github.com/dfinity/candid/blob/master/rust/candid/src/parser/value.rs#L539).

**Requirements**
New `any_val` test passes.

**Considered Solutions**
Removing or modifying [this branch](https://github.com/dfinity/candid/blob/master/rust/candid/src/de.rs#L858).

**Recommended Solution**
This doesn't break all the other tests and doesn't appear to impact the `IgnoreAny` visitor, which is the only other visitor which uses this path.

**Considerations**
Any external visitors using the ignore pathway may experience strange enum behavior, but those are already kind of uncharted waters.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -1,19 +1,20 @@
 //! Deserialize Candid binary format to Rust data structures
 
-use super::error::{Error, Result};
 use super::{
+    error::{Error, Result},
     parser::typing::TypeEnv,
     types::{Field, Label, Type},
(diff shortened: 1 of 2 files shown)


chenyan-dfinity (approved):
LGTM. `ignored_any` is essentially an untyped deserialization.

PR #328, opened 2022-03-25
fix empty vec type

chenyan-dfinity:
Change empty vector type from `vec null` to `vec empty` to satify subtyping rules.

### DIFF
--- a/rust/candid/src/parser/value.rs
+++ b/rust/candid/src/parser/value.rs
@@ -301,7 +301,7 @@ impl IDLValue {
             IDLValue::Float32(_) => Type::Float32,
             IDLValue::Float64(_) => Type::Float64,
             IDLValue::Text(_) => Type::Text,
-            IDLValue::None => Type::Opt(Box::new(Type::Null)),
+            IDLValue::None => Type::Opt(Box::new(Type::Empty)),
             IDLValue::Reserved => Type::Reserved,
             IDLValue::Opt(ref v) => {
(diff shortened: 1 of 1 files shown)


rust/candid/src/parser/value.rs
```diff
@@ -301,15 +301,15 @@ impl IDLValue {
             IDLValue::Float32(_) => Type::Float32,
             IDLValue::Float64(_) => Type::Float64,
             IDLValue::Text(_) => Type::Text,
-            IDLValue::None => Type::Opt(Box::new(Type::Null)),
+            IDLValue::None => Type::Opt(Box::new(Type::Empty)),
```
nomeata, line 304: The line below is reasonable, but this line looks fishy.

Also, would it be worth extending the test suite with a case to catch this, to help other implementations? Or does the reference test suite not cover this code path?

chenyan-dfinity: Why is this a problem? A `null : opt t` can use any type `t`.

The reference test suite doesn't cover this code path. This is only for serializing untyped textual values.

nomeata: Ah, ok. It looked as if it would lead to `null : empty` which would be wrong.

nomeata: Hmm, no, it doesn't even look like that, I was just to hasy while reading. All good, ignore me :-)

PR #338, opened 2022-04-10
add allowing trailing data modifications of decoding functions

seniorjoinu:
**Overview**
I'm building a stable memory allocator that works with `CandidType`'s. Sometimes an allocated block, containing a candid encoded value, has some leftover free bytes at the end (alignment). I'm adding these new functions in order to achieve two things:
1. not break any existing code;
2. be able to deserialize the data from such blocks, despite trailing bytes.

I could fix it on my side, but this would require to add another 4 bytes of data to each block, which is something I don't want to do. Since candid is already able to determine the payload size of a message, it seems like a waste of memory.

Please, take a look @chenyan-dfinity, when you have time.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -19,13 +19,20 @@ use std::{collections::VecDeque, io::Cursor, mem::replace};
 /// Use this struct to deserialize a sequence of Rust values (heterogeneous) from IDL binary message.
 pub struct IDLDeserialize<'de> {
     de: Deserializer<'de>,
+    allow_trailing: bool,
 }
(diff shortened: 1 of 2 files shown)


chenyan-dfinity:
You can simply not calling `done()`, so that the trailing bytes get dropped without any check.

Also be mindful that Candid is not a great format for stable memory: 1) The special `opt` rule makes it easy to loss data when opt type changes; 2) There is no value sharing in the Candid encoding, so it can blow up the stable memory exponentially if the Rust value is shared.

seniorjoinu:
Thanks for the response!

What serialization format/library can you suggest then?
With support for unsized data ofc.

PR #358, opened 2022-06-30
Candid UI update

chenyan-dfinity:
* Fetch did file from canister metadata
* Deprecate `localhost/_/candid` endpoint
* Fetch name section from metadata (instrumented code)
* bump agent-js
* bump ic-cdk

### DIFF
--- a/tools/ui/src/candid.ts
+++ b/tools/ui/src/candid.ts
@@ -53,13 +53,16 @@ export async function fetchActor(canisterId: Principal): Promise<ActorSubclass> 
     const source = window.atob(maybeDid);
     js = await didToJs(source);
   } else {
-    try {
-      js = await getRemoteDidJs(canisterId);
-    } catch(err) {
-      if (/no query method/.test(err)) {
-        js = await getLocalDidJs(canisterId);
-      } else {
-        throw(err);
+    js = await getDidJsFromMetadata(canisterId);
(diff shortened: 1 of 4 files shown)


tools/ui/src/candid.ts
```diff
@@ -92,15 +94,31 @@ export async function getCycles(canisterId: Principal): Promise<bigint|undefined
...
+    const blob = status.get('name') as ArrayBuffer;
+    const decoded = IDL.decode([IDL.Vec(IDL.Tuple(IDL.Nat16, IDL.Text))], blob)[0] as Array<[number,string]>;
     decoded.forEach(([id, name]) => names[id] = name);
   } catch(err) {
     return undefined;
   }
 }
 
+async function getDidJsFromMetadata(canisterId: Principal): Promise<undefined | string> {
+  const status = await CanisterStatus.request({ agent, canisterId, paths: ['candid'] });
```
ericswanson-dfinity, line 113: I think it's `candid:service` https://github.com/dfinity/sdk/blob/master/src/dfx/src/lib/metadata/names.rs

chenyan-dfinity: `agent-js` has its own naming convention: https://github.com/dfinity/agent-js/blob/main/packages/agent/src/canisterStatus/index.ts#L181

PR #369, opened 2022-08-17
postMessage to playground

chenyan-dfinity:
Whenever a user execute an update call, send a heartbeat to playground

### DIFF
--- a/tools/ui/src/candid.ts
+++ b/tools/ui/src/candid.ts
@@ -87,7 +87,7 @@ function postToPlayground(id: Principal) {
   const message = {
     caller: id.toText(),
   };
-  window.parent?.postMessage(`CandidUI${JSON.stringify(message)}`, '*');
+  (window.parent || window.opener)?.postMessage(`CandidUI${JSON.stringify(message)}`, '*');
 }
 
 export async function getCycles(canisterId: Principal): Promise<bigint|undefined> {
--- a/tools/ui/src/candid.ts
+++ b/tools/ui/src/candid.ts
@@ -82,6 +82,13 @@ export function getProfilerActor(canisterId: Principal): ActorSubclass {
   return Actor.createActor(profiler_interface, { agent, canisterId });
 }
 
+function postToPlayground(id: Principal) {
+  const message = {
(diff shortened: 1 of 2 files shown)


tools/ui/src/candid.ts
```diff
@@ -82,6 +82,13 @@ export function getProfilerActor(canisterId: Principal): ActorSubclass {
   return Actor.createActor(profiler_interface, { agent, canisterId });
 }
 
+function postToPlayground(id: Principal) {
+  const message = {
+    caller: id.toText(),
+  };
+  window.parent?.postMessage(`CandidUI${JSON.stringify(message)}`, '*');
```
chenyan-dfinity, line 89: @rvanasa In playground, user can also choose to open Candid UI in a separate tab. In that case, is it possible to postMessage back to the original tab? It can get complicated if the user has multiple playground tabs open.

rvanasa: Right, this could get very complicated with multiple tabs. Perhaps we could send the previous canister ID alongside `caller` so that MP can filter these messages?

chenyan-dfinity: Okay, let's just use `window.parent` for now. What do you mean by previous canister ID?

rvanasa: I was thinking of "previous canister ID" as the original ID passed into the URL query parameters (so that MP can filter out messages from older windows). 

We could try something like `(window.parent || window.opener)?.postMessage(...)`. I tested sending a message via `window.opener` in a popped-out Candid UI window and it seems to work as expected.

rvanasa: Also, to avoid future security vulnerabilities caused by the `*` origin pattern, we could consider looping through the `allowedExternalOrigins` in `external.ts` and sending the message to each of these origins, similar to [this StackOverflow answer](https://stackoverflow.com/a/52003345).

chenyan-dfinity: So `target=_blank` makes `window.opener` to be `null`. But it's a good idea to watch for both.

PR #377, opened 2022-09-27
fix: disable serde's rc feature

roman-kashitsyn:
Some clients don't want to have serde's rc feature enabled because they find it too dangerous. This change makes it easy to use Candid for shared types without serde's rc feature.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -17,6 +17,9 @@ use serde::de::{self, Visitor};
 use std::fmt::Write;
 use std::{collections::VecDeque, io::Cursor, mem::replace};
 
+pub mod rc;
+pub mod arc;
+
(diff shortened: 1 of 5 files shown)


rust/candid/src/de.rs
```diff
@@ -17,6 +17,9 @@ use serde::de::{self, Visitor};
 use std::fmt::Write;
 use std::{collections::VecDeque, io::Cursor, mem::replace};
 
+pub mod arc;
+pub mod rc;
```
chenyan-dfinity, line 21: Is it better to put the module in `lib.rs` instead of `de.rs`, as they contain both serialize and deserialize?

rikonor (approved):

rust/candid/src/de.rs
roman-kashitsyn: Sure, done!

chenyan-dfinity (approved):

PR #378, opened 2022-09-27
refactor!: move Principal here and no re-export ic-types

lwshang:
**Overview**
Re-export `ic-types` often caused build issues in downstream projects.
We believe moving `Principal` implementation to `candid` crate makes more sense.

Also include some other minor changes:
* Remove the cdk feature which is not used by `ic-cdk` anymore. `candid_derive` also needs a major release.
* Update dependencies 

**Considerations**
This **IS** a breaking change. `candid::types::ic_types` is removed.

We need to publish:
*`candid` v0.8.0
*`candid_derive` v0.5.0

### DIFF
--- a/rust/candid/Cargo.toml
+++ b/rust/candid/Cargo.toml
@@ -20,7 +20,7 @@ lalrpop = "0.19.0"
 
 [dependencies]
 byteorder = "1.4.3"
-candid_derive = { path = "../candid_derive", version = "0.5" }
+candid_derive = { path = "../candid_derive", version = "=0.5.0" }
 codespan-reporting = "0.11"
 crc32fast = "1.3.0"
 data-encoding = "2.3.2"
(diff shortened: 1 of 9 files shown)


rust/candid/Cargo.toml
```diff
@@ -20,10 +20,11 @@ lalrpop = "0.19.0"
 
 [dependencies]
 byteorder = "1.4.3"
-candid_derive = { path = "../candid_derive", version = "=0.4.5" }
+candid_derive = { path = "../candid_derive", version = "0.5" }
```
chenyan-dfinity, line 23:
```suggestion
candid_derive = { path = "../candid_derive", version = "=0.5.0" }
```
Always need to pin the derive version.

chenyan-dfinity (approved):

PR #380, opened 2022-10-03
feat: adds candid-wasm-bindgen library for npm

krpeacock:
**Overview**
To facilitate any workflow where someone wants to dynamically generate IDL from declarations, this bundles functionality from Candid into utility methods for string -> file output using wasm-pack

Since IO is not available in a wasm browser context, I had to make changes so that `File` types were not required

**Considerations**


tools/candid-wasm-bindgen/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+# The `console_error_panic_hook` crate provides better debugging of panics by
+# logging them with `console.error`. This is great for development, but requires
+# all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
+# code size when deploying.
+console_error_panic_hook = { version = "0.1.6", optional = true }
+
+# `wee_alloc` is a tiny allocator for wasm that is only ~1K in code size
+# compared to the default allocator's ~10K. It is slower than the default
+# allocator, however.
+wee_alloc = { version = "0.4.5", optional = true }
```
chenyan-dfinity, line 26: I think we stayed away from wee_alloc, because it's too aggressive in asking for new memory pages: https://dfinity.slack.com/archives/CH4CADCJX/p1615932677024900. Do things improve since then?

chenyan-dfinity (commented):
Mostly LG. Is it better to host this in `agent-js` repo? With the new `ic-verify-bls-signature` crate, we probably need a similar npm package there.

tools/candid-wasm-bindgen/Cargo.toml
rvanasa: Relevant conversation: https://github.com/rustwasm/wee_alloc/issues/107

Likewise, here's a security alert I received a few weeks ago for `prettier-plugin-motoko`:

image

It looks like `wee_alloc` is disabled by default, so perhaps the main reason to remove the dependency is to avoid a similar Dependabot message. Should be fine either way.

chenyan-dfinity: Nice! We observed the same memory leak a year ago, and was assuming something was wrong on our platform. Because we didn't see any bug reports about it with such a popular crate...

ilbertt:
Following the [same decision made for the Rust bindgen](https://github.com/dfinity/candid/pull/674#issuecomment-3246475879), JS/TS bindgen needs to be moved closer to the `@icp-sdk/core`. A package similar to this will be created in the [dfinity/icp-js-core](https://github.com/dfinity/icp-js-core) repo

PR #394, opened 2022-11-21
adds CandidType for Uuid behind Uuid feature flag

lunchspider:
**Overview**
Adds CandidType impl for Uuid

**Requirements**
Uuid are widely used as a Ids. This makes it easier to work with

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -251,6 +251,7 @@ dependencies = [
  "sha2 0.10.2",
  "test-generator",
  "thiserror",
+ "uuid",
 ]
 
 [[package]]
@@ -1948,6 +1949,12 @@ dependencies = [
  "matches",
  "percent-encoding",
 ]
+
+[[package]]
+name = "uuid"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "422ee0de9031b5b948b97a8fc04e3aa35230001a722ddd27943e0be31564ce4c"
 
 [[package]]
 name = "vec_map"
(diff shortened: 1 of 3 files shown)


chenyan-dfinity:
Can this be done outside of candid crate, e.g., wrapping a newtype for `Uuid` and implement `CandidType` for the newtype in the user crate?

lunchspider:
yeah but then this wouldn't work
```
#[derive(CandidType)]
struct SomeStruct{
 id : Uuid
}
```
`Uuid` is widely used everywhere creating a wrapper element for uuid means user will have to implement every trait on it..
which becomes very hard if working with 2-3 external crate, primarily databases.

chenyan-dfinity:
Right, you cannot `derive`, but you can manually `impl CandidType for SomeStruct`. Then any structure using `SomeStruct` can derive `CandidType`. This can be a separate crate that other projects depend on.

Reasons for not including this in `candid` crate:

* The above workaround exists and is a common pattern in Rust.
* Avoids unnecessary dependency. Not every project using Candid requires a UUID library. It's reasonable to impl structures from `std`, but not beyond that. 
* You can hide the dependency under a feature flag, but it bloats the library. There are also legal concerns for which license we can include as the transitive dependency.
* Even for a specific feature, we don't want to decide which implementation is the canonical one by including a particular crate in Cargo.toml.

Overall, I think impl a trait for an external crate is an anti-pattern in Rust. We should use newtypes whenever possible.

lunchspider:
ohk, got you!
Thanks

PR #399, opened 2022-12-07
try to fix unsafe inside deserializer

chenyan-dfinity:


rust/candid/src/de.rs
```diff
@@ -135,6 +137,12 @@ macro_rules! check {
     }};
 }
 
+thread_local! {
```
roman-kashitsyn, line 140: I'm not sure I understand why it has to be a thread-local. Can't we compute this array at the call site?

rust/candid/src/de.rs
```diff
@@ -135,6 +137,12 @@ macro_rules! check {
...
+    // Visitor has to implement Copy
+    static ALLOWED_VISITORS: RefCell<[TypeId; 2]> = RefCell::new([TypeId::of::<&crate::parser::value::IDLValueVisitor>(), TypeId::of::<&de::IgnoredAny>()]);
```
roman-kashitsyn, line 142: > `TypeId::of::<&crate::parser::value::IDLValueVisitor>()`

Should it be a reference, not value like `TypeId::of::<crate::parser::value::IDLValueVisitor>()`?
I'm surprised that the call for a reference compiles at all, the doc says types with non-static lifetimes aren't allowed:

> A TypeId is currently only available for types which ascribe to 'static, but this limitation may be removed in the future.
>
> pub fn [of](https://doc.rust-lang.org/nightly/std/any/struct.TypeId.html#method.of)<T>() -> [TypeId](https://doc.rust-lang.org/nightly/std/any/struct.TypeId.html)
where
    T: 'static + ?[Sized](https://doc.rust-lang.org/nightly/std/marker/trait.Sized.html),

chenyan-dfinity: Yeah, we can use a value. We have our own implementation of `TypeId` that removes the static constraint. The std lib doesn't support non-static, because lifetime isn't used when computing TypeId, but we don't care about the lifetime difference here.

Issue #374, opened 2022-09-15
As a developer, I want to [Nat] support bincode/serde_json/cbor deserialize


witter-deland:
Very happy to see that Nat supports Serialize, but it doesn't work properly with bincode\serde_json\cbor
``` rust
use candid::{Deserialize, Nat};
use serde::Serialize;

#[derive(Default, Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
pub struct TestStruct {
    inner: Nat,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_serde_with_bincode() {
        let test_struct = TestStruct::default();
        let serialized = bincode::serialize(&test_struct).unwrap();
        // panicked at 'called `Result::unwrap()` on an `Err` value: DeserializeAnyNotSupported'
        let deserialized = bincode::deserialize(&serialized).unwrap();
        assert_eq!(test_struct, deserialized);
    }

    #[test]
    fn test_serde_with_json() {
        let test_struct = TestStruct::default();
        let serialized = serde_json::to_string(&test_struct).unwrap();
        // panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: sequence, expected Nat value"
        let deserialized: TestStruct = serde_json::from_str(&serialized).unwrap();
        assert_eq!(test_struct, deserialized);
    }

    #[test]
    fn test_serde_with_cbor() {
        let test_struct = TestStruct {
            inner: Nat::from(1000u64),
        };
        let serialized = serde_cbor::to_vec(&test_struct).unwrap();
        println!("serialized {:?}", serialized);
        // panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: sequence, expected Nat value"
        let deserialized: TestStruct = serde_cbor::from_slice(&serialized).unwrap();
        assert_eq!(test_struct, deserialized);
    }
}
```

ninegua:
I have a partial solution to this #400 for JSON & CBOR, but not bincode, because bincode isn't self describing.

I may be difficult to support bincode, because Nat serializes differently for small values and big values. But bincode would require the same format for both. Maybe someone else would have a better idea?

chenyan-dfinity:
I think the problem for bincode is that they don't support `deserialize_any`? `Nat` serializes the same format for both small and big values.

PR #400, opened 2022-12-08
Add nat serialization support for json and cbor, and test cases

ninegua:
A partial solution to address #374.

Note that Nat serializes as arrays in JSON and CBOR, which could be counter-intuitive.

Know caveat is that bincode isn't supported because bincode doesn't use deserialize_any.


chenyan-dfinity (approved):
LGTM

PR #414, opened 2023-03-17
Add type decoder fuzzer

venkkatesh-sekar:
**Overview**

For finding bugs and vulnerabilities in candid. Specifically, this fuzzer tries to target candid decoding against a specific CandidType

**Requirements**

A cargo fuzz-compatible fuzzing harness is required to fuzz candid.

**Considered Solutions**

- Implement a custom struct or utilise an existing type used widely.

**Recommended Solution**

-  The fuzzer tries to decode to the legacy HTTPResponse type. The type is complex enough for achieving better coverage from the fuzzer. The type is implemented directly in the fuzzer source file to avoid external dependencies. 

**Considerations**

Maintenance: the fuzzer must constitute a valid program, and if changes to the candid API are made, it must be ensured that the fuzzer still compiles, and if not, the fuzzer code must be adapted to the new API.


chenyan-dfinity (approved):

robin-kunzler (approved):
@venkkatesh-sekar Thanks for adding the fuzzer, LGTM!

PR #420, opened 2023-04-18
spec: support composite query

chenyan-dfinity:
Follow up with https://github.com/dfinity/interface-spec/pull/144

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -191,7 +191,11 @@ service : {
 A function type describes the list of parameters and results and their respective types. It can optionally be annotated to be *query*, *composite_query* or *oneway*.
 
 * `query` indicates that the function does not modify any state and can potentially be executed more efficiently (e.g., on cached state).
-* `composite_query` is a special `query` function that cannot be called in inter-canister calls. The composite query function can call other query and composite query functions on the same subnet.
+* `composite_query` is a special `query` function that has IC-specific features and limitations:
+  - `composite_query` function can only call other `composite_query` and `query` functions.
+  - `composite_query` function can only be called from other `composite_query` functions (not callable from `query` functions) and from outside of IC.
+  - `composite_query` cannot be made cross-subnets.
+  - All these limitations are temporary due to the implementation. Eventually, `query` and `composite_query` functions will become the same thing.
 * `oneway` function has no return results, and the caller doesn't have to wait for the function return.
 * Other annotations may be added in the future.
 
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -70,7 +70,7 @@ This is a summary of the grammar proposed:
 <actortype> ::= { <methtype>;* }
 <methtype>  ::= <name> : (<functype> | <id>)
 <functype>  ::= <tuptype> -> <tuptype> <funcann>*
-<funcann>   ::= oneway | query
+<funcann>   ::= oneway | query | composite_query
 <tuptype>   ::= ( <argtype>,* )
 <argtype>   ::= <datatype>
 <fieldtype> ::= <nat> : <datatype>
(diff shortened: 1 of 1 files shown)


spec/Candid.md
```diff
@@ -70,7 +70,7 @@ This is a summary of the grammar proposed:
 <actortype> ::= { <methtype>;* }
 <methtype>  ::= <name> : (<functype> | <id>)
 <functype>  ::= <tuptype> -> <tuptype> <funcann>*
-<funcann>   ::= oneway | query
+<funcann>   ::= oneway | query | composite_query
```
ggreif, line 73: I still think this is a very badly chosen name. My suggestion (that came too late to be considered in the replica naming contest) was _report_.

chenyan-dfinity: I wish it's named `query_tmp_hack`

nomeata: QUERY. As in, a larger query.

chenyan-dfinity: QUEyr. It's bigger than `query`, but not big enough (only the same subnet). The spelling is wrong, because it has strange await semantics...

spec/Candid.md
```diff
@@ -188,11 +188,16 @@ service : {
 
 #### Structure
 
-A function type describes the list of parameters and results and their respective types. It can optionally be annotated to be *query*, which indicates that it does not modify any state and can potentially be executed more efficiently (e.g., on cached state). (Other annotations may be added in the future.)
+A function type describes the list of parameters and results and their respective types. It can optionally be annotated to be *query*, *composite_query* or *oneway*.
+
+* `query` indicates that the function does not modify any state and can potentially be executed more efficiently (e.g., on cached state).
+* `composite_query` is a special `query` function that cannot be called in inter-canister calls. The composite query function can call other query and composite query functions on the same subnet.
```
ulan, line 194: > cannot be called in inter-canister calls
I think it should be "cannot be called by update methods" because it is possible to make an inter-canister composite query call.

Could you please also add that the restriction about the "update method" and "same subnet" are due to the incompleteness of current implementation and may be lifted in the future?

spec/Candid.md
ulan: @ggreif: I associate "report" with documents and graphs, not queries. If I was a developer, I would be confused by that name. I think the name has to contain the word "query" or at the very least "queyr" 😄 

@chenyan-dfinity: "because it has strange await semantics..."
There is a trade-off between strange await semantics and strange query semantics. Community voted to keep the proper query semantics sacrificing a corner case of self-call await. We also surveyed the use cases of people interested in composite queries and all of the uses cases involved calling different canisters, where await works as expected.

spec/Candid.md
ggreif: Here are my further naming suggestions
- `inquisition` (religious undertone)
- `interrogation` (too long, menacing)
- `quiz` (many questions, sounds fitting)

spec/Candid.md
ggreif: Here are my further naming suggestions
- `inquisition` (religious undertone)
- `interrogation` (too long, menacing)
- `quiz` (many questions, sounds fitting)

nomeata: We could just ask Dom. That's how naming debates have been resolved in the past :-)

crusso (approved):
But do address @Ulan's suggestions.

spec/Candid.md
crusso: I still quite like `enquiry` or `inquiry`, but never mind.

ulan: I hope this discussion is not blocking the PR.

spec/Candid.md
ulan: I am approving assuming this comment will be addressed.

ulan (approved):
Lgtm, thanks!

spec/Candid.md
chenyan-dfinity: > may be lifted in the future

Lifting all restrictions means we will have just the `query` method, and no `composite query` method, or it's the other way around?

Candid spec is quite high level, and tries to be platform agnostic. See the paragraph above about `query` method. It's only suggesting that `query` method can be more efficient. I prefer to describe `composite query` without mentioning subnet, but seems like it's not easy?

ulan: Sorry for the delay. I thought I've replied already, but don't see my reply here.

> Lifting all restrictions means we will have just the query method, and no composite query method, or it's the other way around?

Query and composite query will become the same thing:

- query will be able to call query and composite query
- composite query will be callable from an update

> Candid spec is quite high level, and tries to be platform agnostic. See the paragraph above about query method. It's only suggesting that query method can be more efficient.

It seems inconsistent that we mention limitations of composite queries without mentioning the main limitation of query: it cannot call any other methods.

If we want to stay platform agnostic, then maybe we should write something more vague: "composite query is a kind of a query with additional platform-specific features and limitations"?

That's said, my personal preference would be to be more pragmatic and mention all information that might be useful for the users (maybe with some comment that it only applies to IC): 
- composite queries can be called only from other composite queries.
- composite queries can call composite queries and queries.
- composite query calls cannot be made cross-subnets.
- queries cannot call any other methods.
All these limitations are temporary due to the implementation.

chenyan-dfinity: I made another pass. PTAL.

> composite queries can be called only from other composite queries.

It can also be called outside of IC.

> queries cannot call any other methods.

There used to be a hack that a query function can call another terminating query function. Is this being removed?

ulan: Thanks, the latest version lgtm!

> There used to be a hack that a query function can call another terminating query function. Is this being removed?

Do you mean ICQC on system and verified application subnets? There non-replicated queries can call other queries on the same subnet. That is exactly what we are packaging as composite_query and releasing on all subnets.

Right now, queries on application subnets cannot call any other queries (terminating or not).

PR #429, opened 2023-05-17
Read didc data from stdin

bitdivine:
**Overview**
`didc` fails to encode or decode large blobs because the hex is too large to pass as a command line argument.

Example:
```
$ didc decode "$(cat payload.hex)"
bash: /home/max/bin/didc: Argument list too long
$ ls -lh payload.hex 
-rw-rw-r-- 1 max max 5,7M Mai 17 11:32 payload.hex
```

**Requirements**
didc should be able to encode or decode large files.

**Considered Solutions**
The hex needs to be passed to the program but not as a command line argument.  This suggests reading input from a file.  I considered:
* Reading the data from stdin
* Reading the data from a named file

**Recommended Solution**
Read data from stdin.  Both methods work, but reading from a named file places a greater burden on a user reading from stdin, than reading from stdin for a user who wants to read from a named file.

If data is in a file, it can be read from stdin trivially:
```
didc decode < payload.hex
```
However if the input is from stdin and we are forced to read from a  named file, we have a choice of bad options:

Option 1:
```
some_pipeline | didc decode --file /dev/stdin
``` 
which is not portable across platforms.  E.g. OSX doesn't have `/dev/stdin`.

Option 2: 
```
tmpfile="$(mktemp)"
some_pipeline > "$tmpfile"
didc decode --file "$tmpfile"
rm "$tmpfile"
```
This is quite verbose and shortcuts, such as not creating a dynamic tmpfile, invite errors.

**Considerations**
I see no negative impacts.

On the positive side, didc will become more convenient to use.  Our code is full of cases where a file is catted just to pass it as an argument:
```
didc encode "$(cat "$didfile")"
```

### DIFF
--- a/tools/didc/src/main.rs
+++ b/tools/didc/src/main.rs
@@ -6,8 +6,9 @@ use candid::{
     Error, IDLArgs, TypeEnv,
 };
 use clap::Parser;
+use std::collections::HashSet;
+use std::io;
 use std::path::PathBuf;
-use std::{collections::HashSet, io};
 
 #[derive(Parser)]
 #[clap(version, author)]
--- a/tools/didc/src/main.rs
+++ b/tools/didc/src/main.rs
@@ -6,8 +6,8 @@ use candid::{
     Error, IDLArgs, TypeEnv,
 };
 use clap::Parser;
-use std::{collections::HashSet, io};
 use std::path::PathBuf;
+use std::{collections::HashSet, io};
 
 #[derive(Parser)]
 #[clap(version, author)]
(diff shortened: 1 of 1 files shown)


tools/didc/src/main.rs
```diff
@@ -232,8 +242,17 @@ fn main() -> Result<()> {
             format,
             annotate,
         } => {
+            let blob = match blob {
+                Some(blob) => blob,
+                None => io::read_to_string(io::stdin())?,
+            };
             let bytes = match format.as_str() {
```
bitdivine, line 249: Files with hex often contain whitespace.  Some editors like to add a newline at the end.  xxd output is easier to view if wrapped at 60 lines rather than being on one gigantic line.  `echo "abba" | didc decode` may complain about abba not being valid candid but it seems silly for it to complain about the trailing whitespace.

bitdivine:
Question:  How is didc tested?  I would be very happy to add tests but I don't see where this is done.  If there is no test setup for the utilities, I would perhaps recommend `bats` as used by the sdk team, for no other reason than it is already used for command line tool testing at dfinity.

chenyan-dfinity (approved):
LGTM.

didc is intentionally untested, because 1) It's supposed to be a very thin wrapper around the candid library. If the didc logic is so complicated that it requires a test, it's better to add this feature into the library, and we test there; 2) didc is mostly a debugging tool. If there are use cases that requires to use `didc`, it usually means we are lacking some features on the sdk side.

PR #431, opened 2023-05-24
Support combining actors and returning imports

ncpenke:
**Overview**

Two use cases:

- Allow modular candid files to define canister interfaces in a model where a canister can implement different interfaces
- Expose imports to facilitate implementation of custom downstream code generators

**Requirements**

- A mechanism exists to combine actors across imported files
- A mechanism exists to detect imported did file changes for code generation

**Considered Solutions**

Re-implement outside of candid, but the broader community wont' benefit from this

**Recommended Solution**

A new `check_file_with_options` API that allows passing custom options. This is extensible in the future for other use-cases. The option will allow actors to be combined in certain cases.

**Considerations**

Minimal impact. This change includes a performance bug fix. The newer functionality is implemented in a backwards compatible manner.


chenyan-dfinity:
Can you elaborate on the use case a bit more? Why is `import` not enough?

ncpenke:
> Can you elaborate on the use case a bit more? Why is `import` not enough?

`import` allows importing the types, but not the service. 

 In our use-case we have interfaces which define both the types, and service methods associated with those types. For example we have a `backup` interface which lets us take a backup of the canister data, a `logger` interface which lets us query the canister's logs. These interfaces consist of both the types, and the service methods (both update and query calls) that define the interface. We were having to manually copy/paste the methods in the final canisters that were using them, instead it would be preferable to maintain these interfaces separately and modularly include them in the final canister.

The other use-case is exposing the imports set. We have a custom rust agent that we generate off the parsed data (since the rust agent that's currently in progress wasn't read before). In order to update the auto-generated code we need to know if any of the imported files changed.

ncpenke:
Awesome, thank you!

chenyan-dfinity:
Closed in favor of #507

PR #432, opened 2023-05-24
Build releases with zig

bitdivine:
**Overview**

The release builds are not statically linked, so fail on platforms with a different libc.

**Requirements**

**Considered Solutions**
- A minimal change that links the linux build with musl
- A more radical approach by which we build with zig.  Cleaner but to benefit fully, all builds need go go via zig, which entails more testing.

**Recommended Solution**

Undecided.  Testing both.  Both work locally, seeing what works in CI.

**Considerations**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,59 +4,21 @@ on:
   push:
     tags:
       - '*'
-
+    branches:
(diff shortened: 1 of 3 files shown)


bitdivine:
Not needed any more.

PR #433, opened 2023-05-24
Build with Musl

bitdivine:
**Overview**
The release linux amd64 builds fail to run on platforms with a different libc.  In particular, dfinity servers such as ch1-spm11 run a different libc so the builds cannot be used there.

**Requirements**
The linux amd64 builds should work on all linux amd64 platforms, including:
- Ubuntu support releases that are not end of life:  https://wiki.ubuntu.com/Releases
  - Long term support, where most users live: 16.04, 18.04, 20.04, 22.04
  - Standard support: 23.04
- Other distributions

In CI we are constrained by the available images: https://github.com/actions/runner-images  We can provide only spot checks for other OSs.  In CI we can test:
  - Ubuntu 20.04, 22.04

**Considered Solutions**
- A minimal change that links the linux build with musl.  This is a tried and tested approach also used for idl2json, ic-admin and other tools.
- A more radical approach by which we build with zig.  Cleaner but to benefit fully, all builds need go go via zig, which entails more testing.  [Example of Zig used for cross platform builds](https://www.uber.com/en-NL/blog/bootstrapping-ubers-infrastructure-on-arm64-with-zig/)
- Maybe build many platforms with cross, including linux amd64, even though the builder is linux amd64.  I have no experience of using cross other than for arm and don't know what linking would look like but it might work.
- Another approach is to make a separate build for each available Ubuntu release and leave other Linux distributions unsupported, however that would be rather unkind to Fedora, Debian, Alpine, Nix and other Linux users.

**Recommended Solution**
Undecided.  Testing musl and zig as these are the solutions I know already work.  Result:  Both zig and musl builds of didc work locally.  Zig is [getting a build error on CI](https://github.com/dfinity/candid/pull/432).  Until that is fixed, calling musl directly is the only working solution on CI.

**Considerations**
I am aware of no negative consequences; a positive consequence is that the linux amd64 build works on more machines.

I would recommend adding a test that the the builds work on a wide variety of platforms, e.g. by building the release artefacts, testing them on all available platforms and only then making the release.  I have included such a test in this PR:
Screenshot from 2023-05-25 01-47-25

Note: The arm32 could theoretically be tested with something like qemu.  It would be nice to know a few distributions that it is used on before putting effort into this.


.github/workflows/release.yml
```diff
@@ -1,21 +1,21 @@
 name: Release
-
 on:
   push:
     tags:
       - '*'
-
+    branches: # TODO: Remove before merging.
+      - musl  # Pushing the feature branch should test this PR.
```
bitdivine, line 7: Here is a run: https://github.com/dfinity/candid/actions/runs/5070940415
Here is the corresponding release.  It is marked as pre-release and can be deleted: https://github.com/dfinity/candid/releases/tag/musl

.github/workflows/release.yml
bitdivine, line 1: Note: Automatically formatting has changed the indentation, making the change appear larger than it is.  Please view the PR with the "hide whitespace" option enabled:
Screenshot from 2023-05-25 02-16-22


I don't see a formatter in this repo so I am formatting with `yq . FILE.yaml`.  If I have missed something, apologies, I would be happy to use another formatter.  I would also be happy to add yq based formatting to this repo if wished.  I would import the yaml formatter from the nns-dapp repo: https://github.com/dfinity/nns-dapp/blob/main/scripts/fmt-yaml

chenyan-dfinity (approved):
LGTM.

PR #434, opened 2023-05-28
limit size of vec null/reserved

### LINKED ISSUE #294: Prevent space bombs
The Candid serialisation format has two singleton types whose encoding size is zero: `null` and `reserved`. That means that an array of arbitrary size over these types (or a super-type like `opt t`) can be represented in constant space in the serialisation format, but may blow up arbitrarily on the receiver side, creating the risk of DoS attacks.

The space savings from this representation (as opposed to a single byte) are probably not significant for practical use cases, because there normally is little value in passing data of singleton types in the first place.

We should hence change the representation. The challenge is how to do that in a sufficiently backwards-compatible manner.

chenyan-dfinity:
Fix #294 

Limit the total size of vec null/reserved. If null took 1 byte, the largest `vec null` we can send on the IC is 2MiB. This is probably the least breaking change we can make to the spec.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -170,6 +170,7 @@ struct Deserializer<'de> {
     // Indicates whether to deserialize with IDLValue.
     // It only affects the field id generation in enum type.
     is_untyped: bool,
+    zero_sized_values: usize,
     #[cfg(not(target_arch = "wasm32"))]
     recursion_depth: u16,
 }
@@ -188,6 +189,7 @@ impl<'de> Deserializer<'de> {
             gamma: Gamma::default(),
             field_name: None,
             is_untyped: false,
+            zero_sized_values: 2 * 1048576,
             #[cfg(not(target_arch = "wasm32"))]
             recursion_depth: 0,
         })
@@ -252,6 +254,16 @@ impl<'de> Deserializer<'de> {
             self.wire_type = self.table.trace_type(&self.wire_type)?;
         }
         Ok(())
+    }
+    fn is_zero_sized_type(&self, t: &Type) -> bool {
+        match t.as_ref() {
+            TypeInner::Null | TypeInner::Reserved => true,
+            TypeInner::Record(fs) => fs.iter().all(|f| {
+                let t = self.table.trace_type(&f.ty).unwrap();
+                self.is_zero_sized_type(&t)
+            }),
+            _ => false,
+        }
     }
     // Should always call set_field_name to set the field_name. After deserialize_identifier
     // processed the field_name, field_name will be reset to None.
@@ -634,10 +646,11 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
                 let expect = e.clone();
                 let wire = w.clone();
                 let len = Len::read(&mut self.input)?.0;
-                if matches!(self.table.trace_type(w)?.as_ref(), TypeInner::Null | TypeInner::Reserved) {
-                    if len > 2 * 1048576 { // 2M message size limit
-                        return Err(Error::msg("vec null/reserved too large"));
+                if self.is_zero_sized_type(w) {
+                    if self.zero_sized_values < len {
+                        return Err(Error::msg("vec length of zero sized values too large"));
                     }
+                    self.zero_sized_values -= len;
                 }
                 visitor.visit_seq(Compound::new(self, Style::Vector { len, expect, wire }))
             }
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -189,7 +189,7 @@ impl<'de> Deserializer<'de> {
             gamma: Gamma::default(),
             field_name: None,
             is_untyped: false,
-            zero_sized_values: 2 * 1048576,
+            zero_sized_values: 2_000_000,
             #[cfg(not(target_arch = "wasm32"))]
             recursion_depth: 0,
         })
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -258,13 +258,11 @@ impl<'de> Deserializer<'de> {
     fn is_zero_sized_type(&self, t: &Type) -> bool {
         match t.as_ref() {
             TypeInner::Null | TypeInner::Reserved => true,
-            TypeInner::Record(fs) => {
-                fs.is_empty()
-                    || fs.iter().all(|f| {
-                        let t = self.table.trace_type(&f.ty).unwrap();
-                        self.is_zero_sized_type(&t)
-                    })
-            }
+            TypeInner::Record(fs) => fs.iter().all(|f| {
+                let t = self.table.trace_type(&f.ty).unwrap();
+                // recursive records have been replaced with empty already, it's safe to call without memoization.
+                self.is_zero_sized_type(&t)
+            }),
             _ => false,
         }
     }
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -133,7 +133,6 @@ macro_rules! check {
         }
     }};
 }
-
 #[cfg(not(target_arch = "wasm32"))]
 macro_rules! check_recursion {
     ($this:ident $($body:tt)*) => {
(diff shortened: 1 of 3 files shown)


spec/Candid.md
```diff
@@ -1229,7 +1229,7 @@ M(_ : reserved) = .
 M : <val> -> <constype> -> i8*
 M(null : opt <datatype>) = i8(0)
 M(?v   : opt <datatype>) = i8(1) M(v : <datatype>)
-M(v^N  : vec <datatype>) = leb128(N) M(v : <datatype>)^N
+M(v^N  : vec <datatype>) = leb128(N) M(v : <datatype>)^N  // N < 2097152 if <datatype> == null or reserved
```
nomeata, line 1232: There are more types of size zero: records where all fields are zero. Maybe write 
> if `|M(v : <datatype>)| = 0`

spec/Candid.md
```diff
@@ -1247,6 +1247,9 @@ M(ref(r) : principal) = i8(0)
 M(id(v*) : principal) = i8(1) M(v* : vec nat8)
 ```
 
+Note:
+
+* Since `null` and `reserved` values take no space, to prevent unbounded sized message, we limit the length of `vec null` and `vec reserved` to be less than 2MiB.
```
nomeata, line 1252:
```suggestion
* Since `null` and `reserved` values, and records of such values, take no space, to prevent unbounded sized message, we limit the length of `vec null` and `vec reserved` to be less than 2MiB.
```

nomeata (changes requested):

nomeata:
Also, can you add a test to the official test suite that will blow up if this isn't fixed, so that other impls will be forced to address this, too?

chenyan-dfinity:
Good catch. There are also nested cases, e.g., `vec vec null` or even `vec variant { a : vec null }`. We need a global counter, and whenever we see `vec {zero-sized value}`, we subtract the counter with the length.

nomeata:
I don't see why you need a counter. To implement a is_zero_sized_type function to recognize the (possibly nested) records?

chenyan-dfinity:
We cannot locally bound the vec size. For example, if we bound the vec size to be N, `vec vec null` can have `N^2` zero values.

nomeata:
But `vec null` itself is never zero size (it has at least the value), so not sure if you need to treat it special. (If you do, then also `vec opt null` etc.)

chenyan-dfinity:
I think `vec {zero-sized value}` is the only problematic case, because it takes sub-linear space. `null` itself takes linear space if you consider both the type and value encoding. `vec opt null` also takes linear space, because `opt null` takes at least one byte.

nomeata:
I agree that `vec {zero sized value}` is the relevant case - so I don't see why you need a counter for nested cases. Maybe it'll be clearer when in the PR though.

(I fear there are other forms of bombs, e.g. deeply nested pairs with sharing and `null` leafs. But one at a time.)

chenyan-dfinity:
The ideal fix is to make `null/reserved` take 1 byte. But that breaks all existing clients. The counter is kind of simulating that behavior without the breaking change. I've updated the spec and test suite in the PR.

test/construct.test.did
```diff
@@ -60,7 +60,11 @@ assert blob "DIDL\01\6d\7b\01\00\02\01\02" == "(blob \"\\01\\02\")"         : (v
 assert blob "DIDL\01\6d\00\01\00\00" == "(vec {})"                          : (Vec) "vec: recursive vector";
 assert blob "DIDL\01\6d\00\01\00\02\00\00" == "(vec { vec {}; vec {} })"    : (Vec) "vec: tree";
 assert blob "DIDL\01\6d\00\01\00\02\00\00" == "(vec { vec {}; vec {} })"    : (vec vec opt empty) "vec: non-recursive tree";
-assert blob "DIDL\01\6d\7f\01\00\e8\07"                                     : (vec null) "vec: space bomb";
+assert blob "DIDL\01\6d\7f\01\00\e8\07"                                     : (vec null) "vec null";
+assert blob "DIDL\01\6d\7f\01\00\80\94\eb\dc\03"                           !: (vec opt nat) "space bomb: vec null <: vec opt nat";
+assert blob "DIDL\01\6d\70\01\00\80\94\eb\dc\03"                           !: (opt nat) "space bomb: send vec reserved to opt nat";
+assert blob "DIDL\04\6c\03\01\7f\02\01\03\02\6c\01\01\70\6c\00\6d\00\01\03\80\94\eb\dc\03" !: (vec record {null;record{reserved};record{}}) "space bomb: zero-sized record";
+assert blob "DIDL\02\6d\01\6d\7f\01\00\05\ff\ff\7f\ff\ff\7f\ff\ff\7f\ff\ff\7f\ff\ff\7f"    !: (vec vec null) "space bomb: vec vec null";
```
chenyan-dfinity, line 67: Without the counter, it's hard to reject this test: Each sub `vec null` is of length 2M and takes 3 bytes, the top level vec length is 5. We can easily extend the top level vec length to 2M / 3 = 699050 and still fits in the 2M message limit. The total memory on the receiver end will be way larger than the Wasm heap.

rust/candid/src/de.rs
```diff
@@ -258,10 +258,13 @@ impl<'de> Deserializer<'de> {
...
-            TypeInner::Record(fs) => fs.iter().all(|f| {
-                let t = self.table.trace_type(&f.ty).unwrap();
-                self.is_zero_sized_type(&t)
-            }),
+            TypeInner::Record(fs) => {
+                fs.is_empty()
+                    || fs.iter().all(|f| {
+                        let t = self.table.trace_type(&f.ty).unwrap();
+                        self.is_zero_sized_type(&t)
+                    })
```
nomeata, line 266: No need to special case the empty record, `.all` does the right thing there too.

rust/candid/src/de.rs
```diff
@@ -254,6 +255,19 @@ impl<'de> Deserializer<'de> {
...
+            }
+            _ => false,
+        }
+    }
```
nomeata, line 270: Also, is this safe against recursive records, which could cause an infinite loop? Or have recursive records been ruled out before somewhere?

nomeata:
I see, thanks

nomeata (dismissed):

test/construct.test.did
nomeata: Is there a clear and simple criteria underlying this, i.e. when to reject and when to accept? One that makes sense in other implementations as well?

And I'd put them into a new file. This makes it easier for an implementation to treat these tests specially (e.g. test them with a heap size bound, or not at all when just starting to implement Candid.)

rust/candid/src/de.rs
chenyan-dfinity: Yes, it's safe. Recursive records have been replaced with empty when parsing the table type. I add a comment in the code.

test/construct.test.did
chenyan-dfinity: In the spec, we have `"we limit the total vector length of such zero-sized values in a message to be 2MiB."`. So this message will be rejected when decoding the second `vec null`.

chenyan-dfinity:
cc @rossberg

rust/candid/src/de.rs
nomeata: Great, thanks!

spec/Candid.md
```diff
@@ -1247,6 +1247,9 @@ M(ref(r) : principal) = i8(0)
 M(id(v*) : principal) = i8(1) M(v* : vec nat8)
 ```
 
+Note:
+
+* Since `null`, `reserved`, `record {}`, and records of such values, take no space, to prevent unbounded sized message, we limit the total vector length of such zero-sized values in a message to be 2MiB. For example, if there are both `vec null` and `vec record {}` in a message, the length of both vectors combined cannot exceed 2MiB.
```
nomeata, line 1252: Maybe it's cleaner to give the length not in bytes, but rather elements? Because the problem is precisely that such elements have different sizes in different encodings.

```suggestion
* Since `null`, `reserved`, `record {}`, and records of such values, take no space, to prevent unbounded sized message, we limit the total vector length of such zero-sized values in a message to be 2465792 elements. For example, if a message contains two vectors, one at type `vec null` and one at type `vec record {}`, then the length of both vectors combined cannot exceed 2465792 elements.
```

chenyan-dfinity: Sure. Where does the number `2465792` come from? It's more than 2MiB?

nomeata: Did I miscalculate? Indeed I did, fixed :-D

chenyan-dfinity: Yeah, let's make it 2000000. Easier for the human eyes.

nomeata (approved):
LGTM

crusso:
It's probably fine, but 2000,000 is the limit on cross subnet message and ingress message. I think message within a subnet can actually be 3MB, unless that's been fixed. Is there someway of bounding this by the size of the raw candid data itself?

For example, for the extended candid used by Motoko stable vars, we'll want to either ignore or greatly increase the bound.

chenyan-dfinity:
The bound is for `vec null/reserved` only. It doesn't apply to other vectors. In practice, nobody really uses `vec null/reserved`?

crusso (approved):
LG(or at least pragmatic)TM

nomeata:
We forgot to bump the candid spec version. Should we have, @chenyan-dfinity?

chenyan-dfinity:
bump here #435. The new version also includes the composite query annotation.

Issue #443, opened 2023-06-30
Rust support arrays larger than 32 items


0kenx:
Use Rust const generics to support arrays larger than 32 items

PR #444, opened 2023-06-30
use const generics for array

0kenx:
Solves #443


chenyan-dfinity (commented):
Thanks, this is cool! Please run `cargo fmt`, then I will merge.

chenyan-dfinity (approved):

PR #462, opened 2023-08-27
spec: allow record {} <: record {null}

chenyan-dfinity:
In the current spec, we only allow `record {} <: record {opt t; reserved}`, i.e. users can omit `opt` fields or `reserved` fields in the argument, but cannot omit `null` fields.

In the Rust implementation, however, we accidentally allowed `null` fields to be optional as well. And this bug has become a feature that people start to use it in the wild...The most prominent use is to decode an empty blob `DIDL\0\0` into a null value, which corresponds to `()` in Rust.

This PR tries to make that bug into a feature to allow optional `null` fields in the record.

Another option is to make `()` in Rust map to empty message in Candid, and have a dedicated `candid::Null` type in Rust. But that's also a breaking change.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -60,7 +60,7 @@ impl<'de> IDLDeserialize<'de> {
                 return T::deserialize(&mut self.de);
             } else {
                 return Err(Error::msg(format!(
-                    "No more values on the wire, the expected type {expected_type} is not opt, reserved or null"
+                    "No more values on the wire, the expected type {expected_type} is not opt, null, or reserved"
                 )));
             }
         }
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -53,14 +53,14 @@ impl<'de> IDLDeserialize<'de> {
         if self.de.types.is_empty() {
             if matches!(
                 expected_type.as_ref(),
-                TypeInner::Opt(_) | TypeInner::Reserved
+                TypeInner::Opt(_) | TypeInner::Reserved | TypeInner::Null
             ) {
                 self.de.expect_type = expected_type;
                 self.de.wire_type = TypeInner::Reserved.into();
                 return T::deserialize(&mut self.de);
             } else {
                 return Err(Error::msg(format!(
-                    "No more values on the wire, the expected type {expected_type} is not opt or reserved"
+                    "No more values on the wire, the expected type {expected_type} is not opt, reserved or null"
                 )));
             }
         }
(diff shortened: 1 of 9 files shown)


chenyan-dfinity:
cc @rossberg

crusso:
I'm not opposed - might even be more consistent with subtyping since nulls are options too, but curious what @nomeata and @rossberg think.

Would need a new version of Candid I think. And updating of Motoko unless it has the same behaviour already.

nomeata:
I vaguely remember that we discussed this before, and I believe I was in favor. Maybe someone finds the conversation.

crusso:
> I vaguely remember that we discussed this before, and I believe I was in favor. Maybe someone finds the conversation.

I couldn't find it on my mobile. So didn't look that hard

crusso (approved):
LGTM (on mobile)

rust/candid/src/de.rs
```diff
@@ -548,7 +548,8 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
     {
         self.unroll_type()?;
         check!(
-            *self.expect_type == TypeInner::Null && *self.wire_type == TypeInner::Null,
+            *self.expect_type == TypeInner::Null
+                && matches!(*self.wire_type, TypeInner::Null | TypeInner::Reserved),
```
ggreif, line 552: So this was a different bug, not accepting `Reserved`?

rust/candid/src/de.rs
```diff
@@ -53,14 +53,14 @@ impl<'de> IDLDeserialize<'de> {
...
-                TypeInner::Opt(_) | TypeInner::Reserved
+                TypeInner::Opt(_) | TypeInner::Reserved | TypeInner::Null
             ) {
                 self.de.expect_type = expected_type;
                 self.de.wire_type = TypeInner::Reserved.into();
                 return T::deserialize(&mut self.de);
             } else {
                 return Err(Error::msg(format!(
-                    "No more values on the wire, the expected type {expected_type} is not opt or reserved"
+                    "No more values on the wire, the expected type {expected_type} is not opt, reserved or null"
```
ggreif, line 63:
```suggestion
                    "No more values on the wire, the expected type {expected_type} is not opt, null or reserved"
```

rust/candid/src/de.rs
chenyan-dfinity: It's the same. When skipping null, we assign the wire type to be `reserved`.

ggreif (approved):
Looks good, comments added.

PR #471, opened 2023-10-04
feat!: split candid crate

lwshang:
Currently, the `candid` crate mainly consists of two parts:
* Candid data encoding/decoding
* Parser and derived functionality (bindgen)

The second part evolves fast which forces us to bump major version frequently.
Such version bumps propagated to the downstream projects (agent-rs, cdk-rs, sdk, ic monorepo, etc).

By splitting out `candid_parser`, we expect to have a stable `candid` crate.

Also, I move `Principal` into a separate `ic_principal` crate. Therefore, we can provide a lightweight dependency for those who only need `Principal`.

## Notable changes
* Some private or crate public methods are now public since they are called cross-crate
* Add `parse_idl_args()` and `parse_idl_value()` because in `candid_parser` we cannot implement `FromStr` for `IDLArgs` and `IDLValue` which are defined in `candid`.
* `TypeEnv` is defined in `candid`, so we can't implement type method `ast_to_type` in `candid_parser`. Instead I turned it into a function which takes `TypeEnv` as the first argument. Similarly, `size_helper()` and `size()` are changed.

## Checklist

- [x] Overhaul the `Cargo.toml` files
- [x] Changelog


rust/candid_parser/src/error.rs
```diff
@@ -0,0 +1,126 @@
+//! `candid::Result<T> = Result<T, candid::Error>>`
```
chenyan-dfinity, line 1:
```suggestion
//! `candid_parser::Result<T> = Result<T, candid_parser::Error>>`
```

rust/candid/Cargo.toml
```diff
@@ -38,22 +35,10 @@ thiserror = "1.0.20"
 anyhow = "1.0"
 binread = { version = "2.1", features = ["debug_template"] }
 
-lalrpop-util = { version = "0.20.0", optional = true }
```
chenyan-dfinity, line 41: On line 21, `codespan-reporting = "0.11"` can be removed

rust/candid/src/bindings/mod.rs
```diff
@@ -2,19 +2,3 @@
 // This module assumes the input are type checked, it is safe to use unwrap.
 
 pub mod candid;
```
chenyan-dfinity, line 4: Can we try to remove the `bindings` modules, and make `candid` module as something like `pretty_printer`?

chenyan-dfinity (commented):
Mostly LG. Can you also update the changelog?

rust/candid/Cargo.toml
lwshang: The `error` submodule still depends on this crate.

Cargo.toml
```diff
@@ -1,7 +1,9 @@
 [workspace]
 members = [
     "rust/candid",
+    "rust/candid_parser",
     "rust/candid_derive",
+    "rust/ic_principal",
```
chenyan-dfinity, line 6: Do you recall why we put `Principal` back to `candid` a while ago? Also do we want to give it a more general name, e.g. `ic_types`, to accommodate more types in the future?

chenyan-dfinity (commented):
Let's merge this to a `beta` branch. I have a few other breaking changes to make after this. Let's aim for an official release early November.

Cargo.toml
lwshang: We used to have the [`ic-types` crate](https://crates.io/crates/ic-types) which contained `principal` and `hash_tree` modules.
The modules were moved to `candid` and `agent-rs` repo respectively. And `ic-types` is [deprecated](https://github.com/dfinity/ic-types/pull/27).

IMO, it's better to have `ic_principal` crate which contains only `Principal`. It will be a stable crate on the top of the dependency tree. We won't have to bump its version because of changes in other parts.

lwshang:
I just pushed the `beta` branch which is identical to this `lwshang/split` branch.

Issue #474, opened 2023-10-08
Add II login to Candid UI for authenticated calls


Web3NL:
Requires a canister to implement `http_request` method and serve `/.well-known/ii-alternative-origins` with `https://a4gq6-oaaaa-aaaab-qaa4q-cai.raw.icp0.io` as an origin, as I currently understand.

If this is a good idea and there is support for this feature, I'm willing to start working on a PR.

Web3NL:
[See this demo](https://m4ul7-aqaaa-aaaal-qcewq-cai.raw.icp0.io/?id=mvxad-wyaaa-aaaal-qcexa-cai) and [forked repo](https://github.com/Web3NL/candid/tree/feat-authenticated-calls)

For testing I have a canister with:

`http_request` query method with `update = ?true` (see [upgrade to update](https://m4ul7-aqaaa-aaaal-qcewq-cai.raw.icp0.io/?id=mvxad-wyaaa-aaaal-qcexa-cai))

`http_request_update` and serves `/.well-known/ii-alternative-origins` with my self deployed candid ui canister as an alternative origin

`hello` query method that only answers if `caller = allowed`

setter and getter for `var allowed: Principal`

**To test:**
Open the [demo](https://m4ul7-aqaaa-aaaal-qcewq-cai.raw.icp0.io/?id=mvxad-wyaaa-aaaal-qcexa-cai)

Query the `hello` method. This should give an error

Login with your II and copy your principal id
 
Paste in `_set_allowed_principal` and submit

Query the `hello` method

**Next steps**
Some more testing (hopefully local Docker testing with Selenium for local alternative origins testing)
UI touches
Gather feedback

PR #475, opened 2023-10-09
Feat: authenticated calls in Candid UI

Web3NL:
See [#474](https://github.com/dfinity/candid/issues/474) and [demo link](https://m4ul7-aqaaa-aaaal-qcewq-cai.raw.icp0.io/?id=mvxad-wyaaa-aaaal-qcexa-cai)

**Overview**
Allow authenticated calls via Internet Identity (and possibly pem files)

**Requirements**
New deps: `auth-client` and `identity` packages from `agent-js`
Modifications and additions to `tools/ui`

**Considered Solutions**
Added II to Candid UI canister

**Recommended Solution**
Also recommend uploading pem files (with warning to developer about security)

**Considerations**
Should be easy to maintain, modifications of existing files are kept to a minimum
New deps version bumps

**TODO**
Remove `dfx.json`, `canister_ids.json` and `target_test_canister` from branch (left them for demo purpose)
Browser compatibility check
Testing preferably with Docker and Selenium (for local II alternative origins testing over https)
UI enhancements
Gather feedback


rvanasa:
Looks good so far! Thanks for the contribution!

tools/ui/src/candid.ts
```diff
@@ -77,6 +80,14 @@ export async function fetchActor(canisterId: Principal): Promise<ActorSubclass>
   }
   const dataUri = 'data:text/javascript;charset=utf-8,' + encodeURIComponent(js);
   const candid: any = await eval('import("' + dataUri + '")');
+
+  authClient = authClient ?? (await AuthClient.create())
```
rvanasa, line 84: Perhaps we could deactivate the idle timeout (similar to the NNS dapp):

```suggestion
  authClient = authClient ?? (await AuthClient.create({
    idleOptions: {
      disableIdle: true,
      disableDefaultIdleCallback: true,
    },
  }))
```

v2i0s2h2 (commented):
sir

maheshVishwakarma1998 (commented):
ser

tools/ui/dfx.json
```diff
@@ -14,6 +14,11 @@
           "name": "candid:service"
         }
       ]
+    },
+    "target": {
```
chenyan-dfinity, line 18: We don't need this canister.

chenyan-dfinity (approved):
Thanks for your contribution! 

We don't need the `target_test_canister` here. I can merge once it's removed.

Web3NL:
Thank you!

tools/ui/target_test_canister/AlternativeOrigin.mo
```diff
@@ -0,0 +1,15 @@
+import Text "mo:base/Text";
```
chenyan-dfinity, line 1: Also remove this directory.

Web3NL:
Here we go!

Web3NL:
Many thnx!

Will this make it into dfx 0.15.2? I would be so proud 😄

chenyan-dfinity:
@Web3NL Actually, there is one more UI fix we need to do. We use Candid UI in the playground with small width. The II button looks big there. Is there an easy way to make it smaller with small window width? (cc @rvanasa if you have a quick fix)

Screenshot 2023-10-17 at 12 13 13 PM

Web3NL:
Will have a look tomorrow i hope

ggreif:
@Web3NL Does the merged version work with `ic0.app`-domain anchors too?

chenyan-dfinity:
Yes, I think it supports both `raw` and `icp0.io`

Web3NL:
> @Web3NL Does the merged version work with `ic0.app`-domain anchors too?

yes, i added a form to choose between ic0.app / icp0.io. Same for raw / non-raw

Web3NL:
> @Web3NL Does the merged version work with `ic0.app`-domain anchors too?

Excuse me, I corrected my last response.

Does this answer your question, or were you refering to the default identity provider, which now is always `identity.internetcomputer.org` and never `identity.ic0.app`

The `derivationOrigin` (canister for which we want to obtain a Principal) may be `icp0.io` or `ic0.app`

rvanasa:
> Actually, there is one more UI fix we need to do. We use Candid UI in the playground with small width. The II button looks big there. Is there an easy way to make it smaller with small window width?

From a responsive UX standpoint, the following three improvements would help a lot on smaller screen sizes:

- Using `@media` CSS queries to change the login form / button layout for different screen widths
- Moving checkboxes inside of the `<label>` tag (which makes it possible to toggle the checkbox by clicking the label)
- Shortening "Login with Internet Identity" to just "Login" (maybe depending on screen width)

@Web3NL, here is a partial solution in case you want to borrow any of this code: https://github.com/dfinity/candid/pull/480

Web3NL:
I will have a look, thnx for tagging.

And I think this feature deserves some documention on the portal on how a canister should serve the `ii-alternative-origins` file.

I will attend to that 🙂

ggreif:
> yes, i added a form to choose between ic0.app / icp0.io. Same for raw / non-raw
> Note, the obtained principal is independent of derivation origin domain, but does change if you include / exclude raw

I am asking because my `ic0.app`-based anchor didn't work in the example canister.

Web3NL:
Ah, I see.

Is this because `identityProvider` is set to `identity.internetcomputer.org`?

If so, then there should also be an option to choose `identity.ic0.app` as the provider

chenyan-dfinity:
Deployed to the main net!

Another possible improvement is to only show the `Login` button when `alternativeOrigins` is presented in the destination canister response.

PR #500, opened 2023-12-05
refactor candid ui auth

chenyan-dfinity:
* Allow customizing II url and derivationOrigin via parameter `ii` and `origin`. Report error on the login button if there is an error in the parameters.
* Remove `derivationOrigin` by default

### DIFF
--- a/tools/ui/src/auth/auth.ts
+++ b/tools/ui/src/auth/auth.ts
@@ -2,135 +2,124 @@ import { Principal } from "@dfinity/principal"
 import { authClient } from "../candid"
 import { refresh_actor } from "../index"
 import { dfinityLogo, copyIcon } from "./icons"
-import { identityProvider } from "./identityProvider"
 
 export async function renderAuth() {
-  ;(await authClient?.isAuthenticated()) ? insertLogout() : insertLoginForm()
+  const is_logged = await authClient?.isAuthenticated();
+  is_logged ? insertLogout() : insertLoginForm();
 }
 
-type DerivationOriginOptions = {
-  canisterId: Principal
-  domain: string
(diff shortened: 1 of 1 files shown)


frederikrothenberger:
> if there is an easy way to detect that alternativeOrigin has been set in the destination canister, we can enable derivationOrigin without the parameter.

There is not, unfortunately.

frederikrothenberger (approved):

tools/ui/src/auth/auth.ts
```diff
@@ -2,135 +2,145 @@ import { Principal } from "@dfinity/principal"
...
-
-function derivationOrigin(options: DerivationOriginOptions): string {
-  if (options.domain !== "icp0.io" && options.domain !== "ic0.app") {
-    throw new Error("Invalid domain")
-  }
-
-  return options.raw
-    ? "https://" + options.canisterId.toString() + ".raw." + options.domain
-    : "https://" + options.canisterId.toString() + "." + options.domain
+  authClient?.logout();
```
rvanasa, line 144: `AuthClient.logout()` is apparently an async function ([source](https://github.com/dfinity/agent-js/blob/40d98004cd7cb1da7fe62794762b9ffc24a4d21e/packages/auth-client/src/index.ts#L555)):

```suggestion
  await authClient?.logout();
```

rvanasa (approved):
LGTM (suggested a fix for a subtle corner case).

chenyan-dfinity:
cc @Web3NL

chenyan-dfinity:
Make sense. Users can still customize the derivation url with the `origin` parameter. This sounds like a good default option.

PR #502, opened 2023-12-06
feat: Make it easy to convert data to IDLValue and candid text format

bitdivine:
**Overview**
It is not well documented how to convert types that implement `CandidType` into an `IDLValue`.

**Requirements**
It is easy to convert instances of Rust types into `IDLValues` using the best available method.

**Considered Solutions**
- Provide a Candid library method for the conversion. (implemented in this PR)
- Document the current approach.  However the current approach seems inefficient  and can probably be improved.

**Recommended Solution**
Provide a library method for converting data to Candid.

The current method - serializing and parsing - can probably be improved greatly, so it seems preferable to offer a conversion method in the candid library and then improve that method rather than document a method that is sub-optimal and then expect developers to update their code when a better method is documented.

**Considerations**
- I think that this will have no negative influence on performance.
- This makes conversion of data to text-format candid a one-liner:  `IDLValue::try_from_candid_type(data).unwrap().to_string()`

### DIFF
--- a/rust/candid/src/types/value.rs
+++ b/rust/candid/src/types/value.rs
@@ -1,6 +1,6 @@
 use crate::types::number::{Int, Nat};
 use crate::types::{Field, Label, Type, TypeEnv, TypeInner};
-use crate::{Error, Result};
+use crate::{CandidType, Decode, Encode, Error, Result};
 use serde::de;
 use serde::de::{Deserialize, Visitor};
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


chenyan-dfinity (approved):
`annotate_types` is a best-effort implementation, notably subtyping is not supported. A better way is to use `from_bytes_with_types` which implements everything from the official deserializer.

bitdivine:
@chenyan-dfinity  Thank you!

PR #504, opened 2023-12-13
spec: allow import to merge service definitions

chenyan-dfinity:
There is a unspecified behavior when `import` contains a main service. This PR supports merging the imported main service via `import_service`.

Better keyword suggestions are welcome.

Related discussion https://github.com/dfinity/candid/pull/431


chenyan-dfinity:
cc @rossberg

spec/Candid.md
```diff
@@ -64,7 +64,7 @@ The purpose of an IDL is defining the signature, and thereby the *type* of an ac
 This is a summary of the grammar proposed:
 ```
 <prog>  ::= <def>;* <actor>?
-<def>   ::= type <id> = <datatype> | import <text>
+<def>   ::= type <id> = <datatype> | import <text> | import* <text>
```
crusso, line 67: How about `import service` instead of `import*`?

spec/Candid.md
```diff
@@ -519,20 +519,25 @@ type B = A;  // error: cyclic type definition
 In order to allow splitting interface definitions up into multiple files or share common definitions between multiple interfaces, *import* declarations are provided.
 
 ```
-<def>   ::= ... | import <text>
+<def>   ::= ... | import <text> | import* <text>
 ```
 
-An import refers to another interface file by URL. The semantics is that of textual inclusion, except that definitions from the imported file must not refer to definitions from the importing file.
+There are two forms of import: `import` and `import*`. Both `import` and `import*` refer to another interface file by URL. The type definitions from the imported file are textually included in the importing file. The definitions from the imported file must not refer to definitions from the importing file.
```
crusso, line 525: So a natural question is what happens if either or both services have service arguments? It's not clear what the semantics  should be.

crusso: An alternative might to add `and`/`or` as in Motoko to take the glb/lub of two service types.

crusso: (but then you'd have to define the glb/lub, if those even exist for Candid's weird subtyping rules).

chenyan-dfinity: Good catch! I think we can disallow service arguments in the imported file. By definition, service constructor refers to a Wasm module, while service refers to the running canister. Importing a service constructor would imply some kind of static linking, which we don't support anyway.

I think we cannot define glb/lub, the opt rule makes every `opt T` a bottom type...

spec/Candid.md
chenyan-dfinity: `import_service` then

crusso: Why not just use the keyword combo, to avoid introducing a new keyword? Prettier too

chenyan-dfinity: Make sense. Updated.

spec/Candid.md
```diff
@@ -537,7 +538,7 @@ service : A
 ```
 File `B.did`:
 ```
-import "A.did";  // Cannot use import* because of method name duplication
+import "A.did";  // Cannot use import_service because of method name duplication
```
crusso, line 541:
```suggestion
import "A.did";  // Cannot use `import service` because of method name duplication
```

crusso (approved):
LGTM

PR #514, opened 2024-01-23
Bump agent-js and remove workaround for Candid UI

rvanasa:
Updates agent-js to the latest version and removes a hotfix used for the latest dfx release. 

Pending the next release of `@dfinity/agent`.

### DIFF
--- a/tools/ui/package-lock.json
+++ b/tools/ui/package-lock.json
@@ -8,11 +8,11 @@
       "name": "ui",
       "version": "0.1.0",
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


chenyan-dfinity (approved):

chenyan-dfinity:
Do we want to bump to the latest agent-js as well?

rvanasa:
Should be ready to merge now. Had to wait for a follow-up adjustment to the logic which shipped with `@dfinity/agent` 0.21.4. 

Confirmed that everything works as expected locally and on Gitpod after a hard page reload.

PR #516, opened 2024-01-26
make zero-sized value counting more fine grained

chenyan-dfinity:
Update spec to limit the number of zero-sized values in a message, instead of the sum length of `vec {zero-sized value}`.


rust/candid/src/de.rs
```diff
@@ -337,16 +334,13 @@ impl<'de> Deserializer<'de> {
...
-            TypeInner::Record(fs) => fs.iter().all(|f| {
-                let t = self.table.trace_type(&f.ty).unwrap();
-                // recursive records have been replaced with empty already, it's safe to call without memoization.
-                self.is_zero_sized_type(&t)
-            }),
-            _ => false,
+    fn dec_zero_sized_value(&mut self) -> Result<()> {
+        //eprintln!("{}", self.zero_sized_values);
+        if self.zero_sized_values == 0 {
+            return Err(Error::msg("Too many zero sized values"));
```
crusso, line 340:
```suggestion
            return Err(Error::msg("Too many zero sized values"));
```
```suggestion
            return Err(Error::msg("Too many zero-sized values"));
```

crusso (approved):
Looks reasonable to me. Nice to have some examples too.

rust/candid/src/de.rs
```diff
@@ -337,16 +334,13 @@ impl<'de> Deserializer<'de> {
...
-        match t.as_ref() {
-            TypeInner::Null | TypeInner::Reserved => true,
-            TypeInner::Record(fs) => fs.iter().all(|f| {
-                let t = self.table.trace_type(&f.ty).unwrap();
-                // recursive records have been replaced with empty already, it's safe to call without memoization.
-                self.is_zero_sized_type(&t)
-            }),
-            _ => false,
+    fn dec_zero_sized_value(&mut self) -> Result<()> {
+        //eprintln!("{}", self.zero_sized_values);
```
mraszyk, line 338: should this comment be dropped?

spec/Candid.md
```diff
@@ -1255,7 +1255,7 @@ M(id(v*) : principal) = i8(1) M(v* : vec nat8)
 
 Note:
 
-* Since `null`, `reserved`, `record {}`, and records of such values, take no space, to prevent unbounded sized message, we limit the total vector length of such zero-sized values in a messagev (on the wire) to be 2,000,000 elements. For example, if a message contains two vectors, one at type `vec null` and one at type `vec record {}`, then the length of both vectors combined cannot exceed 2,000,000 elements.
+* Since `null`, `reserved` and `record {}` take no space, to prevent unbounded sized message, we limit the total number of such zero-sized values in a message (on the wire) to be 2,000,000 elements. For example, a value of type `record { null; reserved; record { record {} } }` contains 3 zero-sized values; a value of type `vec opt record { null; null }` with length 100 contains at most 200 zero-sized values (it is 200 when the vector contains no `null` values); if a message contains two vectors, one at type `vec null` and one at type `vec record {}`, then the length of both vectors combined cannot exceed 2,000,000 elements.
```
mraszyk, line 1258: > a value of type `record { null; reserved; record { record {} } }` contains 3 zero-sized values
Could you please list the zero-sized values explicitly? I believe to have counter 4:
- `null`
- `reserved`
- `record {}`
- `record { record {} }`

mraszyk: And in fact also the entire `record { null; reserved; record { record {} } }` is a 5th zero-sized value there.

rust/candid/src/de.rs
```diff
@@ -337,16 +334,13 @@ impl<'de> Deserializer<'de> {
         }
         Ok(())
     }
-    fn is_zero_sized_type(&self, t: &Type) -> bool {
```
mraszyk, line 340: Conceptually, we should check this condition for every (nested) value and limit all such (nested) zero-sized values at 2M in total.

spec/Candid.md
chenyan-dfinity: I will try to rephrase the spec. We only count the occurrences of `null`, `reserved` and `record {}` to simulate the effect that if these values were not zero sized. `record { record {} }` would not be zero sized, if `record {}` weren't zero sized.

rust/candid/src/de.rs
chenyan-dfinity: See above. That would be overcounting, e.g., `record { null; null }` would be counted as 3 instead of 2.

PR #519, opened 2024-02-02
spec: add a cost model for values on the wire

chenyan-dfinity:
The goal is to have a metering mechanism for deserialization.


crusso:
On mobile so hard to review.
Is this just counting every value decoded? That seems reasonable to me, except for how to choose the limit? Maybe the number of bits in 2MB? I guess for bignums/blob/text and other variable sized values it needs to depend on the size of the value.

spec/Candid.md
```diff
@@ -1253,9 +1253,43 @@ M(ref(r) : principal) = i8(0)
...
 ```
 
-Note:
+##### Wire cost model
+
+Some values on the wire take fewer space than the values on the host language, e.g., `null`, `reserved` and `record { record {} }` take no space on the wire. To prevent sending messages of exponential size, we introduce a cost model `C` for the values on the wire.
+
+```
+C : <val> -> <primtype> -> nat
+C(n : nat)      = 1
```
crusso, line 1262: The cost for nat shouldn't be constant Can we not use the natural bit or byte size for each value?

chenyan-dfinity: Yeah, we should. I was undecided whether the cost should be `log2 n` or the actual bytes in LEB128. One can keep padding `\x80` in the LEB128 encoding with a small `n`. Also, should we charge the same for length encoding in the constructor types? It was currently 1.

chenyan-dfinity:
> Is this just counting every value decoded?

Ideally, we are counting the decoding cost. I guess byte size in the decoded value is a good approximation, although the value size depends on the runtime representation. The cost model tries to give a runtime agnostic cost measure. This also doesn't count the cost of backtracking in the opt rule.

> how to choose the limit? Maybe the number of bits in 2MB?

Maybe this can vary based on the language? My current thinking is to not set a fixed number in the spec, and allow different implementations (or even different canister endpoints) to set their own limit. But this introduces UB...

spec/Candid.md
```diff
@@ -1318,17 +1352,17 @@ Deserialisation at an expected type sequence `(<t'>,*)` proceeds by
 
  * checking for the magic number `DIDL`
  * using the inverse of the `T` function to decode the type definitions `(<t>,*)`
- * check that `(<t>,*) <: (<t'>,*)`, else fail
```
mraszyk, line 1321: Why is the subtyping check dropped here?

test/spacebomb.test.did
```diff
@@ -18,20 +18,6 @@ assert blob "DIDL\04\6c\03\01\7f\02\01\03\02\6c\01\01\70\6c\00\6d\00\01\03\80\94
 assert blob "DIDL\02\6d\01\6d\7f\01\00\05\ff\ff\3f\ff\ff\3f\ff\ff\3f\ff\ff\3f\ff\ff\3f"    !: () "vec vec null (extra argument)";
 assert blob "DIDL\03\6c\01\d6\fc\a7\02\01\6d\02\6c\00\01\00\80\ad\e2\04"                   !: () "vec record {} (extra argument)";
 
-// Messages with exactly 2_000_000 zero-length elements should succeed
```
mraszyk, line 21: By dropping those tests, the above comment "Messages with more than 2_000_000 zero-length elements in vectors should be rejected" makes no sense anymore.

spec/Candid.md
```diff
@@ -1253,9 +1253,43 @@ M(ref(r) : principal) = i8(0)
...
+C(n : nat)      = |leb128(n)|
+C(i : int)      = |sleb128(i)|
+C(n : nat<N>)   = N / 8
+C(i : int<N>)   = N / 8
+C(z : float<N>) = N / 8
+C(b : bool)     = 1
+C(t : text)     = 1 + |t|
+C(_ : null)     = 1
+C(_ : reserved) = 1
+C(_ : empty) undefined
```
mraszyk, line 1271:
```suggestion
C(_ : empty)    = undefined
```

spec/Candid.md
```diff
@@ -1253,9 +1253,43 @@ M(ref(r) : principal) = i8(0)
...
+
+C : <val> -> <constype> -> nat
+C(null : opt <datatype>) = 1
+C(?v   : opt <datatype>) = 1 + C(v : <datatype>)
+C(v^N  : vec <datatype>) = 1 + sum_i C(v[i] : <datatype>)
+C(kv*  : record {<fieldtype>*}) = 1 + sum_i C(kv : <fieldtype>*[i])
+C(kv   : variant {<fieldtype>*}) = C(kv : <fieldtype>*[i])
```
mraszyk, line 1278:
```suggestion
C(kv   : variant {<fieldtype>*}) = 1 + C(kv : <fieldtype>*[i])
```

mraszyk: I see that one is added further below in the equation
```
C((k,v) : k:<datatype>) = 1 + C(v : <datatype>)
```
resolving this thread.

rust/candid/src/de.rs
```diff
@@ -714,6 +732,12 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
         V: Visitor<'de>,
     {
         self.unroll_type()?;
+        if matches!(
+            self.wire_type.as_ref(),
+            TypeInner::Null | TypeInner::Reserved | TypeInner::Opt(_)
+        ) {
+            self.dec_value_cost(1)?;
```
mraszyk, line 739: Why do we have an if around here?

spec/Candid.md
chenyan-dfinity: This was an out-dated algorithm. We only check subtyping for reference types, which is covered by the `~>` derivation below. This is needed to support extensible variant.

rust/candid/src/de.rs
chenyan-dfinity: We are only counting wire types. Deserialization can reach this function when the expected type is opt, regardless of the wire type.

rust/candid/src/de.rs
```diff
@@ -1030,12 +1058,14 @@ impl<'de, 'a> de::MapAccess<'de> for Compound<'a, 'de> {
...
                         match e.id.get_id().cmp(&w.id.get_id()) {
                             Ordering::Equal => {
+                                self.de.dec_value_cost(1)?;
                                 self.de.set_field_name(e.id.clone());
                                 self.de.expect_type = expect.pop_front().unwrap().ty;
                                 self.de.wire_type = wire.pop_front().unwrap().ty;
                             }
                             Ordering::Less => {
                                 // by subtyping rules, expect_type can only be opt, reserved or null.
+                                self.de.value_cost += 1; // correct the later decrement from serialize_option of producing None.
```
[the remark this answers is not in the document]
chenyan-dfinity: Yeah, I know. This whole way of implementing counting is fragile: The decoding algorithm is driven by the expected type, so for each expected type, we need to consider all possible wire types, it's easy to make mistakes. My hope is that we can catch the counting bugs with testing. The invariant we are testing is that the cost of decoding at type `T` equals the cost of decoding the same bytes untyped.

chenyan-dfinity:
> > This also doesn't count the cost of backtracking in the opt rule.

> Could we please include it (as well as any other work incurring cost)?

I try to avoid that for a few reasons: 1) This would require giving a cost model for the derivation rule, which restricts the way people can implement the deserialization algorithm; 2) The backtracking cost is usually very small: you fail early when checking the types, and then skipping the wire type is exactly the wire cost we defined here.

chenyan-dfinity:
Closed in favor of #524

PR #524, opened 2024-02-15
add metering for deserialization

chenyan-dfinity:


rust/candid/src/de.rs
```diff
@@ -327,26 +334,26 @@ impl<'de> Deserializer<'de> {
...
-        match t.as_ref() {
-            TypeInner::Null | TypeInner::Reserved => true,
-            TypeInner::Record(fs) => fs.iter().all(|f| {
-                let t = self.table.trace_type(&f.ty).unwrap();
-                // recursive records have been replaced with empty already, it's safe to call without memoization.
-                self.is_zero_sized_type(&t)
-            }),
-            _ => false,
+    fn add_cost(&mut self, cost: usize) -> Result<()> {
+        if let Some(n) = self.decoding_cost {
```
mraszyk, line 350: The name decoding_cost is misleading - it is the remaining decoding "capacity".

rust/candid/src/de.rs
```diff
@@ -1143,27 +1190,31 @@ impl<'de, 'a> de::VariantAccess<'de> for Compound<'a, 'de> {
             *self.de.expect_type == TypeInner::Null && *self.de.wire_type == TypeInner::Null,
             "unit_variant"
         );
+        self.de.add_cost(1)?;
```
mraszyk, line 1193: Why not start with this as always?

rust/candid/src/de.rs
```diff
@@ -236,7 +245,7 @@ struct Deserializer<'de> {
     // Indicates whether to deserialize with IDLValue.
     // It only affects the field id generation in enum type.
     is_untyped: bool,
-    zero_sized_values: usize,
+    decoding_cost: Option<usize>,
```
mraszyk, line 248: Why would this ever be None?

mraszyk: I see - that makes sense.

spec/Candid.md
```diff
@@ -1318,9 +1314,12 @@ Deserialisation at an expected type sequence `(<t'>,*)` proceeds by
 
  * checking for the magic number `DIDL`
  * using the inverse of the `T` function to decode the type definitions `(<t>,*)`
- * check that `(<t>,*) <: (<t'>,*)`, else fail
  * using the inverse of the `M` function, indexed by `(<t>,*)`, to decode the values `(<v>,*)`
- * use the coercion function `C[(<t>,*) <: (<t'>,*)]((<v>,*))` to understand the decoded values at the expected type.
+ * use the coercion function `v : t ~> v' : t'` to understand the decoded values at the expected type.
+
+Note on implementation:
```
crusso, line 1320: So I guess this is admitting defeat for having an abstract cost model?
I can see that the formulation using inference rules makes this hard.
Would it still be hard if we specified the decoder as a monadic function, returning a cumulative cost and optional value, making the backtracking and skips explicit?

crusso: Or would that be too ugly or prescriptive?

rust/candid/src/de.rs
```diff
@@ -138,29 +132,109 @@ impl<'de> IDLDeserialize<'de> {
...
+    /// C(b : bool)     = 1
+    /// C(t : text)     = 1 + |t|
+    /// C(_ : null)     = 1
+    /// C(_ : reserved) = 1
+    /// C(_ : empty)    = undefined
+    ///
+    /// C : <val> -> <constype> -> nat
+    /// C(null : opt <datatype>)  = 2
+    /// C(?v   : opt <datatype>)  = 2 + C(v : <datatype>)
+    /// C(?v   : opt <datatype'>) = 2 + C(v : <datatype>) * 50 + 10  // when v cannot be converted to <datatype'>
```
crusso, line 186:
```suggestion
    /// C(?v   : opt <datatype'>) = 2 + C(v : <datatype>) * 50 + 10  // when v cannot be converted to <datatype'>
```
where do 50 and 10 come from?

rust/candid/src/de.rs
```diff
@@ -138,29 +132,109 @@ impl<'de> IDLDeserialize<'de> {
...
+    /// C(v^N  : vec <datatype>)  = 2 + 3 * N + sum_i C(v[i] : <datatype>)
+    /// C(kv*  : record {<fieldtype>*})  = 2 + sum_skipped_i C(kv : <fieldtype>*[skipped_i]) * 50 + sum_expected_i C(kv : <fieldtype>*[expected_i])
```
crusso, line 188: How do you know which are skipped/expected without having the expected type as an argument?

crusso (commented):
Added some comments. Maybe we could discuss a bit tomorrow, or @mraszyk can fill me in on how we got here.

spec/Candid.md
chenyan-dfinity: I think it's possible to have an abstract cost model by assigning weights for each derivation rule. But I see several problems with this approach: 
1) It forces the implementation to strictly follow the derivation rules; 
2) Types in the host language is not completely captured by the model. For example, `[Nat8]` vs `Blob` in Motoko, and `Vec<(K, V)>` vs `HashMap<K,V>` in Rust. They map to the same Candid type, but have different cost in the host language. Even with the same `record` type, it can be more expensive to construct a record in Rust than in Motoko (or vice versa);
3) Even if we have a cost model, it would be hard to come up with a threshold in the spec. It's application dependent (canister endpoints and inter-canister calls may need different bounds, and stable memory won't need any limit). Plus there is always a gap between the cost model and the real implementation.

Looking from a different angle, this is a resource problem originating from a specific platform and a specific implementation. Given infinite resource, the Candid spec doesn't need to reject these large payload. Just like we don't specify stack depth in the spec, we can leave the concrete cost metering to the implementation, so that it matches better with the platform and language runtime.

rust/candid/src/de.rs
chenyan-dfinity: These are some arbitrary numbers. 50x is the multiply when we skip values in the decoder. `10` accounts for the cost of restoring states in the backtracking.

rust/candid/src/de.rs
chenyan-dfinity: Yeah, ideally it should be `C(v : t ~> t')`, but it's also verbose. Maybe I can define just `C(v:t)`, and say that skipping values cost 50x more.

crusso (approved):
Discussed offline

PR #535, opened 2024-03-15
Adjust Candid UI login button

rvanasa:
This PR makes the following changes:

* Adjusts the wording of the login button tooltips when using Candid UI in the local replica
* Shows corresponding error messages in the dev console
* Adjusts the icon animation to be more subtle and to only occur when the button is enabled

### DIFF
--- a/tools/ui/src/auth/auth.ts
+++ b/tools/ui/src/auth/auth.ts
@@ -59,7 +59,7 @@ async function insertLoginForm() {
     }
     if (origin) {
       if (!is_valid_url(origin)) {
-        throw new Error("Please provide a valid URL in `origin` query parameter");
+        throw new Error("Please provide a valid URL in the `origin` query parameter");
       }
       console.warn("`derivationOrigin` is enabled. Remember to disable alternative origins in the production canister.");
     }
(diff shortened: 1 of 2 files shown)


tools/ui/src/auth/auth.ts
```diff
@@ -47,10 +47,10 @@ async function insertLoginForm() {
       provider = "https://identity.internetcomputer.org";
     }
     if (!provider) {
-      throw new Error("Please provide internet identity url in ii parameter");
+      throw new Error("Please provide a URL to your local Internet Identity service using the `ii` query parameter");
```
chenyan-dfinity, line 50: If we have a doc for deploying II locally, we can also link it here.

chenyan-dfinity (approved):

tools/ui/src/auth/auth.ts
rvanasa: Seems like we don't yet have any documentation, but the AI chatbot is actually able to answer questions about it since it was mentioned in a Discord conversation at one point. 

I'll keep this in mind in case we can eventually include a link to proper documentation.

PR #536, opened 2024-03-16
pretty printing vec nat8 to blob

chenyan-dfinity:


rust/candid/src/pretty/candid.rs
```diff
@@ -98,6 +98,7 @@ pub fn pp_ty_inner(ty: &TypeInner) -> RcDoc {
         Var(ref s) => str(s),
         Principal => str("principal"),
         Opt(ref t) => kwd("opt").append(pp_ty(t)),
+        Vec(ref t) if matches!(t.as_ref(), Nat8) => str("blob"),
```
crusso, line 101: This will mess with Motoko again.

PR #543, opened 2024-04-10
feat: add infinite type test

mraszyk:
This PR adds an infinite type test corresponding to this didc invokation:
```
$ didc decode 4449444c016c020778080001002a00000000000000
Error: Fail to decode argument 0 from table0 to unknown

Caused by:
    0: input: 4449444c016c02077808000100_2a00000000000000
       table: type table0 = empty
       wire_type: empty, expect_type: empty
    1: Subtyping error: Cannot decode empty type
```


mraszyk:
An analogous test already exists: https://github.com/dfinity/candid/blob/ae4d0f794a808b88f8b83da7eb96ca87b3d751d4/test/construct.test.did#L93

PR #554, opened 2024-05-31
feat: make error clonable

nathanosdev:
**Overview**
This makes `candid::Error` clonable, which makes it easier for downstream clients to also make their `Error` types cloneable.

I've made a related PR for [agent-rs](https://github.com/dfinity/agent-rs/pull/563).

**Considered Solutions**
I can also work around by using `Arc<T>` in downstream clients.

**Considerations**
This shouldn't have any impact on security or performance. Anyone that was interacting directly with the `Custom` variant may have a breaking change since it is wrapped in an `Arc<T>` now.

### DIFF
--- a/rust/candid/src/error.rs
+++ b/rust/candid/src/error.rs
@@ -1,18 +1,18 @@
 //! `candid::Result<T> = Result<T, candid::Error>>`
 
 use serde::{de, ser};
-use std::{io, num::ParseIntError};
+use std::{io, num::ParseIntError, sync::Arc};
 use thiserror::Error;
 
 pub type Result<T = ()> = std::result::Result<T, Error>;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


chenyan-dfinity (commented):
This is a breaking change. It's quite labor intensive to bump the candid major version in the mono repo. So we try to avoid breaking changes as much as possible.

I can see two options: 1) we can let this PR to sleep for a while, until we make a major release (which can take quite a long time); 2) Candid error is not really structured. For the downstream users, it's enough to convert to `err.to_string()`, and clone as needed.

nathanosdev:
I agree that making a major bump just for this feature is not worth the effort, so I'm fine with leaving this to sleep and in the meantime I will just `to_string()` downstream.

ilbertt:
In addition to the previous discussion, in Rust the errors are not typically clonable. Closing it.

PR #555, opened 2024-06-10
spec: candid type selector

chenyan-dfinity:


spec/Type-selector.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Design
+
+Motivated by Cascading Style Sheets ([CSS](https://en.wikipedia.org/wiki/CSS)), we define a configuration language to specify the customization of a Candid interface. The configuration file enables the separation of the Candid interface and its configuration, including layout, binding, validation, etc.
+
+There are two parts in the config language: 1) A way to quickly select particular type nodes in Candid types; 2) Attach semantic meaning to the selected type nodes.
+
+### Type selector
+
+For the first part, we define a type selector `<selector>` similar to CSS selector.
```
crusso, line 24:
```suggestion
For the first part, we define a type selector `<selector>` similar to a CSS selector.
```

spec/Type-selector.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+<selector> := <scopes> (. <path>)? | <path>
+<path> := <name> (. <name>)*
+<scopes> := <scope> (. <scope>)*
+<scope> := func:<name> | func:init | arg:<num> | ret:<num>
```
crusso, line 30: Isn't this missing the `variant` and `record` selectors used in <paths>? As in `tree.variant.branch.record.left` below?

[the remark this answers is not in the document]
crusso: Oh, it's explained further down.

crusso (changes requested):
I'm not saying do this now, but it is possible to give a  semantics to this? Eg mapping a candid AST to  an extended Candid AST with nodes decorated with application specific sets of attributes?

spec/Type-selector.md
crusso: Is this resolved? Or did I misunderstand something?

crusso (approved):
I'll approve but see unresolved issue...

spec/Type-selector.md
chenyan-dfinity: Scope only refers to top level functions and arguments. `record` and `variant` is only part of the path. It should become clear when we get the full semantics specified out...

crusso (approved):

PR #568, opened 2024-08-21
Optimize the performance of deserializing types containing options

frankdavid:
The `recoverable_visit_some` method needs to clone the `Deserializer` in order to backtrack in case of errors. This cloning is a very expensive operation which this PR optimizes.

1. Wrap the `TypeEnv` in an `Rc`. Since the `TypeEnv` is a heavy object and is barely updated after the `Deserializer` is constructed, it makes sense to keep a single instance of it so cloning becomes cheap.
2. Make the happy path in `recoverable_visit_some` as cheap as possible by recursing on self instead of the clone. In case of an error, we reset the state from the clone.


chenyan-dfinity (approved):
Thanks! This is great improvements!

PR #576, opened 2024-09-27
display canister logs in candid UI

chenyan-dfinity:
* Fetch canister logs at the start and after each update call
* refactor candid ui to skip certification
* bump package.json dependencies


ggreif (approved):
LGTM! Low-risk change IIUC.

PR #577, opened 2024-10-17
feat(didc-js): Javascript multi purpose library with candid utils

keplervital:
**Overview**
This library adds similar tools as `didc` but aims to be used in a Javascript environment.

**Requirements**
- Can encode textual candid into a hex string.
- Can decode a hex string into candid textual representation.
- Can extract the methods of a candid service IDL.

**Considered Solutions**
We've also considered a similar approach as the candid ui which leverages a canister to expose a query method, but we ruled that out because we want to use this as the user types in a form which would result in many network requests.

**Recommended Solution**
Leverage `candid_parser` to create a javascript package exposing a compiled WebAssembly that uses those crates.

**Considerations**
This is a new library, there's no consumer yet.


chenyan-dfinity:
Please consider using https://github.com/dfinity/candid/blob/master/rust/candid_parser/src/bindings/javascript.rs instead of `idl2json` to derive JSON object. `idl2json` is incomplete and isn't compatible with agent-js's candid format.

keplervital:
> Please consider using https://github.com/dfinity/candid/blob/master/rust/candid_parser/src/bindings/javascript.rs instead of `idl2json` to derive JSON object. `idl2json` is incomplete and isn't compatible with agent-js's candid format.

The JS binding use case is different, this binding gives back JavaScript code iirc, which is not what I’m after.

The target format aims to be a displayable value that we want back, but if you’re not happy with that idl2json lib because it’s incomplete, I’m happy to keep only candid as a target format since it already gives me enough for what I’m building.

chenyan-dfinity:
Right, it gives back JS code, and if you eval the code, you get a JS value that can be displayed. Candid UI uses that to display JSON format. If it sounds too hacky, we can just remove idl2json.

Cargo.toml
```diff
@@ -22,3 +23,10 @@ anyhow = "1.0"
 rand = "0.8"
 arbitrary = "1.3"
 console = "0.15"
+wasm-bindgen = "0.2"
+console_error_panic_hook = "0.1"
+wasm-bindgen-console-logger = "0.1"
+log = "0.4"
+getrandom = "0.2"
+serde-wasm-bindgen = "0.5"
+js-sys = "0.3"
```
chenyan-dfinity, line 32: Move these inside the didc-js crate. It's not used by the rest of the projects.

tools/README.md
```diff
@@ -2,3 +2,4 @@
 
 - Didc -- multiple-purpose candid tool, including the "DFINITY Interface Description Compiler"
 - Candiff -- relates candid values that change over time
+- @dfinity/didc -- JavaScript library counter part of the didc tool, meant to be used in JavaScript projects.
```
chenyan-dfinity, line 5: If it's published to npm, add a link here.

chenyan-dfinity (approved):

PR #578, opened 2024-10-25
fix(didc-js): wasm-pack build targets bundler mode

keplervital:
Using `wasm-pack` with bundler mode is more appropriate for most use cases since this library is through bundlers.


tools/didc-js/README.md
```diff
@@ -1,68 +1,3 @@
 # @dfinity/didc
 
 A multi-purpose Candid tool for JavaScript projects, including encoding and decoding Candid values. 
-
-## Usage
```
chenyan-dfinity, line 5: Why removing the usage?

chenyan-dfinity (approved):

tools/didc-js/README.md
keplervital: Because that's already included inside [wasm-package/README.md](https://github.com/dfinity/candid/pull/578/files#diff-991b05c0305831073a0d082d60b25cd03ab1379cf2299fa907d356dae484e688) which is the one that get's published to npmjs.

PR #583, opened 2024-12-02
Add a test regarding structs with a new Option field

randombit:
This checks that adding an Option field to a struct has the expected behavior, where

* For an old writer and new reader, the struct deserializes with a field of None
* For a new writer and an old reader, the struct deserializes with the value of the Option ignored.

**Overview**
This PR just adds a test. Within the crypto team we were not sure if this situation (adding an `Option` to a Candid struct) worked as we expected (CRP-2642). This test is to confirm our understanding. But it also seems like a nice test to have for the future.

### DIFF
--- a/rust/candid/tests/serde.rs
+++ b/rust/candid/tests/serde.rs
@@ -194,6 +194,56 @@ fn test_option() {
 }
 
 #[test]
+fn test_struct_with_option() {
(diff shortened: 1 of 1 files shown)


randombit:
CI failure looks unrelated

randombit:
There may well be some other test that covers this, I did look but I could not immediately find it

PR #590, opened 2024-12-18
perf: Introduce TypeKey instead of String keys for named types

frankdavid:
**Overview**
We can achieve serious performance improvements by introducing a new `TypeKey` struct instead of `String` for candid types represented by `Type::Var`. When deserializing a candid message, we need to store information about these types (e.g. in the `TypeEnv` struct). Previously, we used `BTreeMap` with `String` keys. This is slow and a significant amount of time is spent looking up types by name.

**Optimization 1: distinguish indexed (integer) keys from String keys**
When decoding type information from the message header in `binary_parser.rs`, the types are indexed. Previously these indexed types were assigned a `String` key by calling `format!("table{index}")`. Working with integer keys is much faster.

**Optimization 2: use `HashMap` instead of `BTreeMap`**
Looking up values in a HashMap is much faster (especially if the keys are small).

**Optimization 3: precompute hash of `String`-based `TypeKey`s**

**Breaking changes**
This is a breaking change:
  - In `Type::Var(var)` `var` now has type `TypeKey` instead of `String`. Calling `var.as_str()` returns `&str` and `var.to_string()` returns a `String`. The string representation of indexed variables remains `table{index}` to maintain compatibility with previous versions.
  - `TypeEnv` now contains a `HashMap` instead of `BTreeMap`. Code that relied on the iteration order of the map (e.g. `env.0.iter()`) should make use of the newly added `TypeEnv::to_sorted_iter()` method which returns types sorted by their keys.


lwshang (dismissed):

adamspofford-dfinity (approved):

crusso:
This looks ok to me. However, I do recall @chenyan-dfinity being averse to introducing breaking changes to the API but I'm not sure how serious a problem this is.

Another approach might have been to introduce a second, Indexed Var constructors, used by the binary decoder to index into an array, since the binary decoder should actually know the size of the type table. This would be even more efficient than using a hash table, I expect.

frankdavid:
The `TypeEnv` data structure is indexed by the `TypeKey` (named or indexed), so just a new `TypeInner` variant would have not sufficed. 

I did benchmark using a combination of a `HashMap` (for named types) + a `Vec` (for indexed types) to implement `TypeEnv`/`TypeMap` but it resulted in negligible performance gains and made the code quite a bit more complex so I decided against it.

lwshang (approved):

PR #601, opened 2025-03-26
feat: Typescript Native interface and Wrapper

giorgiopiatti-caffeinated:
### Goal
The goal is to generate a typescript interface of a canister, which is more readable by AI models and follows Typescript standard types.

This introduces new functionality, is not changing existing functionality. It wraps the existing js/ts solution.

The new type mapping changes how optionals are mapped, instead of using arrays ([] for None and [Value] for Some(Value)) we use:
- `T | null` for bottom-most optional
-  `Optional<T>` for nested optionals
-  `optionalProperty?` as part of record definition e.g.
`interface {
    optionalProperty?: string;
    other: string;
}`

### How to run
We expose 2 new target options to `bind`: `ts-native-wrapper`and `ts-native-interface`, which can be used: 
- didc bind --target ts-native-interface example.did > example.d.ts
- didc bind --target ts-native-wrapper example.did > example.ts

### How to run tests
The test are based on Typescript type checker, we check that the generated wrapper correctly translates data between the old interface and the new typescript interface. Currently the generation of the new interface is not part of the tests, as is similar to the old interface.

`cd tests/typescript-native && ./test.sh <test-case>`

where `<test-case>` is a `.did` file, currently we have `recursive.did`, `example_1.did`, `small.did`.

### Future Extensions

- [ ] Add arguments names of candid functions. Motoko can now generate candid with named args see [https://github.com/dfinity/motoko/pull/4943](https://github.com/dfinity/motoko/pull/4943)
- [ ] Add comments present in original file


luc-blaeser (approved):
I like the clean idiomatic TS interface.
Maybe add the test that Nathan suggested

giorgiopiatti-dfinity:
Depends on fix from #604

PR #604, opened 2025-04-12
fix: typescript interface for func and service should match javascript

giorgiopiatti-caffeinated:
There is a mismatch between the IDL defined in javascript and the types defined in typescript. There are some hardcoded conversion that always map a func and a service to something containing principal.
```
 Func(_) => str("[Principal, string]"),
 Service(_) => str("Principal"), 
```
Suppose we have the following candid definition
```
type f = func (int8) -> (int8);
type g = f;
type h = func (f) -> (f);
type o = opt o;
service : { f : (nat) -> (h); g : f; h : g; o : (o) -> (o) }
```
Given the above candid this is incorrect to me
`export type h = ActorMethod<[[Principal, string]], [Principal, string]>;`
Correct
`export type h = ActorMethod<[f], f>;`


ggreif:
@christoph-dfinity should be aware of this...

giorgiopiatti-dfinity:
Closing, as the issue is more complicated see #606

PR #609, opened 2025-05-08
feat: release v0.10.14 with Decoding Utility Functions

stiegerc:
In ic-cdk `0.17.1` macros such as query had direct support for `decoding_quota` and `skipping_quota`, for [0.18.0](https://docs.rs/ic-cdk/0.18.0/ic_cdk/attr.query.html) this was dropped and support for a general user supplied decoding function was added instead.

So if you wanted to use candid but with specific value for `decoding_quota`, you are expected to write your own function. Since the argument is a fixed `Vec<u8>`, it's not possible to provide a generic function for different quotas, except through [const generics](https://practice.course.rs/generics-traits/const-generics.html#:~:text=Const%20generics%20are%20generic%20arguments,type%20T%20and%20N%3A%20usize.). Support for this was recently [added to ic-cdk](https://github.com/dfinity/cdk-rs/pull/599).

A frequent use-case with ic-cdk `0.17.1`
```Rust
 #[query(decoding_quota = 10000)]
fn my_query(arg: T) -> U {
   ..
}
```
could be translated with the functions added here as
```Rust
use candid::decode_args_with_decoding_quota as decoding_quota;

#[query(decode_with = "decoding_quota::<10000,_>")]
fn my_query(arg: T) -> U {
   ..
}
```

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -5,8 +5,6 @@ jobs:
   runBenchMark:
     name: run benchmark
     runs-on: ubuntu-latest
-    env:
-      RUSTC_VERSION: 1.81.0
     steps:
     - uses: actions/checkout@v4
     - name: Checkout base branch
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -5,6 +5,8 @@ jobs:
   runBenchMark:
     name: run benchmark
     runs-on: ubuntu-latest
+    env:
+      RUSTC_VERSION: 1.82.0
     steps:
     - uses: actions/checkout@v4
     - name: Checkout base branch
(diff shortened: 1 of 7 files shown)


stiegerc:
With `ic-cdk` as it is, this is the only generic solution I can see. Unfortunately const generics are currently quite limited; I would have preferred a solution with a const `DecoderConfig` as a parameter or similar.

lwshang:
The `decode_one_with_decoding_quota` has been tested manually in the mono-repo.

The others were tested in a local project.

We have to mark the doc code blocks with `ignore`. This is intended to not introduce `ic-cdk` as a dependency . Because `ic-cdk` depends on `candid`, we don't want to have cyclic dependencies.

lwshang (approved):

PR #617, opened 2025-06-11
ci: turn on terminal colors

maksymar:
**Overview**
Improved logs readability by turning on terminal colors.

Before
image
image

After
image
image

**Considerations**
May affect scripts that process raw output text.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,53 +5,57 @@ on:
       - master
   pull_request:
 
+env:
+  CARGO_TERM_COLOR: always # Force Cargo to use colors
+  TERM: xterm-256color
+
 jobs:
   rust:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
-    - name: Install stable toolchain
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -5,53 +5,57 @@ on:
...
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Build
+        run: |
+          cargo build
+      - name: Run tests
+        run: |
+          cargo test --no-default-features --color always -- --color always
+          cargo test --color always -- --color always
+          cargo test --features all --color always -- --color always
```
ilbertt, line 39: Isn't the `CARGO_TERM_COLOR` env enough? Feel free to resolve

maksymar: unfortunately, no. there are different parts of cargo outputs that controlled by env or those flags.
cargo build output is controlled by env. test groups by the first `--color always` and if I'm not mistaken `ok` and `fail` are by the second `--color always`. but you can try to simplify it, I may be mistaken.

PR #619, opened 2025-06-12
feat: support line comments

ilbertt:
**Overview**
Adds support for **line** comments that are "associated" to types, fields and services in Candid. "Associated" means that there's no empty line between the last line of the comment and the underlying type, field or service.

**Considerations**
Still needs support for block comments.


christoph-dfinity:
Summarizing our pairing session from this morning:

- Approach using the side table for documentation comments from #621 does what we need to do for parsing
- Currently codegen uses `candid::compile` which pretty prints the semantic representation (`candid/src/types/internal.rs`) rather than the syntactic representation (`candid_parser/src/types.rs`). This is problematic in a few ways (eg. it loses the distinction between vec nat8 and blob), but it also means to print the documentation comments we need to cram then into the semantic types or store them off to the side somehow.
- I suggested exploring if we can make the code generators produce the syntactic representation, and Luca said he's going to spend some time looking into that.

christoph-dfinity:
> One problem I see is that we would need to pull the candid_parser as a dependency inside the candid and candid_derive crates, in order to use the IDL* types.

Hmm I see... I think I'd prefer moving the IDL* types into the candid crate (assuming that's the one at the "bottom" of the hierarchy)

PR #622, opened 2025-06-17
refactor!: move `IDL*` types into the candid crate, under the `parser` feature flag

ilbertt:
**Overview**
Move the parser's `IDL*` types under the `candid` crate, in order to share it between that crate and the `candid_parser` crate.

The main reason is that we want to use the `IDL*` types in the bindings generator, instead of bringing around the `TypeEnv`, which holds inner `Type`s that are not meant for codegen.

**Considerations**
Introduces breaking changes.

### DIFF
--- a/rust/candid/Cargo.toml
+++ b/rust/candid/Cargo.toml
@@ -45,8 +45,9 @@ candid_parser = { path = "../candid_parser" }
 bignum = ["dep:num-bigint", "dep:num-traits"]
 printer = ["dep:pretty"]
 value = ["bignum", "printer"]
+parser = []
 default = ["serde_bytes", "printer", "bignum"]
-all = ["default", "value", "ic_principal/arbitrary"]
+all = ["default", "value", "ic_principal/arbitrary", "parser"]
 
 [[test]]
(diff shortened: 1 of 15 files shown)


christoph-dfinity (approved):
Looks good to me. One bikeshed comment would be that `src/types/syntax.rs` rather than `src/types/parser.rs` might be a nicer name, as the idea is that it describes the syntax of candid files. Not just the parser is a consumer of this format, but also everyone that wants to generate a .did file.

ilbertt:
> Looks good to me. One bikeshed comment would be that `src/types/syntax.rs` rather than `src/types/parser.rs` might be a nicer name, as the idea is that it describes the syntax of candid files. Not just the parser is a consumer of this format, but also everyone that wants to generate a .did file.

@christoph-dfinity good point. I've renamed both the module and the feature flag.

ilbertt:
Since this PR introduces breaking changes, I'll wait before merging it to check if the changes make sense in the branch I'm working on: https://github.com/dfinity/candid/compare/luca/idl-types-in-parser

PR #623, opened 2025-06-20
feat!: remove `didc test` subcommand

ilbertt:
**Overview**
Removes the outdated and undocumented `didc test` subcommand.
Additionally, updates the `didc` readme with the latest subcommands.

**Considerations**
Breaking change.

PR #625, opened 2025-06-20
feat!: use IDL types in parser

ilbertt:
**Overview**
Refactor the `candid_parser` and the `candid_derive` crates to use IDL* types. This allows us to enrich the IDL* types e.g. with comments later on, without affecting serialization and deserialization performance.

**Requirements**
Blocked by:
- #622
- #623

**Considerations**
Breaking change.


rust/candid/src/types/syntax.rs
```diff
@@ -12,6 +17,129 @@ pub enum IDLType {
     ServT(Vec<Binding>),
     ClassT(Vec<IDLArgType>, Box<IDLType>),
     PrincipalT,
+    /// Used for Rust recursive types.
+    KnotT(TypeId),
```
ilbertt: We don't need a `KnotT` here, we can find the type from the `ENV` `RefCell` before converting it into an `IDLType`

ilbertt: Resolved

ilbertt:
> Notes from our 1n1/pairing session earlier:
> 
> * We shouldn't have a `KnotT` constructor on the syntax `IDLType`
> * The `check_*` functions should stay as they are in `candid_parser::typing`, and we'll want to produce a merged `IDLProg` by "concatenating" the imported did files (not as text, but by concatenating the defs and service methods). That should also let us not need an extra `IDLEnv`
> * Let's try to have the derive code operate on the TypeContainer as it used to, and only convert to `IDLType` at the last moment (attaching the comments we collected along the way).

@christoph-dfinity I've addressed all these points

lwshang:
Luca and I discussed the upcoming changes offline.

We've decided to proceed with the current PR structure for now, given the dependent work (e.g., #625). However, we'll plan to move the types into a separate, privately-dependent crate (could be named `candid_syntax`) before the final release.

Since development in this repo has been slow, we can continue using the master branch as our active development branch and aren't expecting any patch releases soon.

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -381,15 +385,15 @@ pub mod test {
     }
     fn pp_encode<'a>(args: &'a candid::IDLArgs, tys: &'a [candid::types::Type]) -> RcDoc<'a> {
         let vals = value::pp_args(args);
-        let tys = super::pp_rets(tys);
+        let tys = super::pp_rets(&[]);
```
christoph-dfinity: Intentional change? Looks surprising

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -381,15 +385,15 @@ pub mod test {
...
         let items = [tys, vals];
         let params = concat(items.iter().cloned(), ",");
         str("IDL.encode").append(enclose("(", params, ")"))
     }
 
     fn pp_decode<'a>(bytes: &'a [u8], tys: &'a [candid::types::Type]) -> RcDoc<'a> {
         let hex = pp_hex(bytes);
-        let tys = super::pp_rets(tys);
+        let tys = super::pp_rets(&[]);
```
christoph-dfinity: Same here, was this intentional?

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -404,31 +408,29 @@ import { Principal } from './principal';
...
+        crate::check_prog(&mut env, &idl_prog).unwrap();
+        let idl_merged_prog = IDLMergedProg::from(idl_prog);
+        res += &super::compile(&idl_merged_prog);
         for (i, assert) in test.asserts.iter().enumerate() {
             let mut types = Vec::new();
             for ty in assert.typ.iter() {
                 types.push(crate::typing::ast_to_type(&env, ty).unwrap());
             }
-            let host = HostTest::from_assert(assert, &env, &types);
+            let host = HostTest::from_assert(assert, &env, &[]);
```
christoph-dfinity: More empty slices

christoph-dfinity (changes requested):
Looking very good! Spotted a couple places that might need changes.

rust/candid_parser/src/bindings/javascript.rs
ilbertt: I'm waiting for #623 to remove that piece of code

rust/candid_parser/src/bindings/javascript.rs
ilbertt: Same as above, I'm waiting for https://github.com/dfinity/candid/pull/623 to remove that piece of code

rust/candid_parser/src/bindings/javascript.rs
ilbertt: See above

rust/candid_parser/src/test.rs
ilbertt, line 1: This struct and enum were not used

rust/candid_parser/src/typing.rs
```diff
@@ -151,29 +155,6 @@ fn check_defs(env: &mut Env, decs: &[Dec]) -> Result<()> {
...
-    }
-    for (id, ty) in env.0.iter() {
-        let mut seen = BTreeSet::new();
-        if has_cycle(&mut seen, env, ty)? {
-            return Err(Error::msg(format!("{id} has cyclic type definition")));
-        }
-    }
-    Ok(())
-}
-
```
ilbertt, line 176: Moved into a `TypeEnv` method

lwshang (approved):
LGTM

rust/candid/src/types/subtype.rs
ilbertt, line 1: We should not change this code, because if we incur in `Future` or `Unknown`, this explodes now.

PR #626, opened 2025-06-20
feat: makes optional types optional in TypeScript

krpeacock:
**Overview**
A longstanding request, this will make our JavaScript / TypeScript bindings more idiomatically appropriate. Must be paired with a change to the JS agent implementation

fixes SDK-2183


rust/candid_parser/src/bindings/typescript.rs
```diff
@@ -36,7 +41,29 @@ fn pp_ty<'a>(env: &'a TypeEnv, ty: &'a Type, is_ref: bool) -> RcDoc<'a> {
...
         }
         Principal => str("Principal"),
-        Opt(ref t) => str("[] | ").append(enclose("[", pp_ty(env, t, is_ref), "]")),
+        Opt(ref t) => {
+            // if the inner type is an option, we treat it as a nested option
+            if t.as_ref().is_opt() {
+                println!("pp_ty: {ty:?}");
+                println!("t: {t:?}");
+
+                str("[] | ").append(enclose("[", pp_ty(env, t, is_ref, current_type_id), "]"))
```
ilbertt: How about we create a separate function for this? Something like `pp_opt(RcDoc) -> RcDoc`.

This way, we avoid duplicating the logic around, like it's done in line 59.

rust/candid_parser/src/bindings/typescript.rs
```diff
@@ -36,7 +41,29 @@ fn pp_ty<'a>(env: &'a TypeEnv, ty: &'a Type, is_ref: bool) -> RcDoc<'a> {
...
+            // for self referential check, check if the id is the same as the type that is inside the optional
+            else if let (Some(current_id), TypeInner::Var(ref var_id)) =
+                (current_type_id, t.as_ref())
+            {
+                if current_id == var_id {
+                    // Self-referential optional type like "type o = opt o"
+                    // In TypeScript, this becomes a recursive array type
+                    str("[] | [").append(ident(current_id)).append("]")
+                } else {
+                    pp_ty(env, t, is_ref, current_type_id).append(str(" | undefined"))
```
ilbertt: Same here, let's turn this into a function and remove the duplicated code from line 64.

rust/candid_parser/tests/assets/actor.did
```diff
@@ -1,11 +1,17 @@
...
 service :
 {
   f : (nat) -> (h);
   g : f;
   h : g;
-  o : (o) -> (o);
+  recursiveOpt : (recursiveOpt) -> (recursiveOpt);
+  normalOpt : (normalOpt) -> (normalOpt);
+  nestedOpt : (nestedOpt) -> (nestedOpt);
+  doubleNestedOpt : (doubleNestedOpt) -> (doubleNestedOpt);
```
ilbertt: How about we create a new test file specifically for testing these `opt` cases? Something like `optionals.did`

rust/candid_parser/src/bindings/typescript.rs
```diff
@@ -3,7 +3,12 @@ use candid::pretty::utils::*;
 use candid::types::{Field, Function, Label, SharedLabel, Type, TypeEnv, TypeInner};
 use pretty::RcDoc;
 
-fn pp_ty<'a>(env: &'a TypeEnv, ty: &'a Type, is_ref: bool) -> RcDoc<'a> {
+fn pp_ty<'a>(
+    env: &'a TypeEnv,
+    ty: &'a Type,
+    is_ref: bool,
+    current_type_id: Option<&'a str>,
```
ilbertt: Maybe we can rename it?
```suggestion
    parent_type_id: Option<&'a str>,
```
Sounds more appropriate for the use we make of it

ilbertt (changes requested):

rust/candid_parser/src/bindings/typescript.rs
nathanosdev: We can simplify the logic to remove the duplicate code without extracting to a function

```rust
        Opt(ref t) => {
            // if the inner type is an option, we treat it as a nested option
            if t.as_ref().is_opt() {
                return str("[] | ").append(enclose("[", pp_ty(env, t, is_ref, current_type_id), "]"));
            }

            // for self referential check, check if the id is the same as the type that is inside the optional
            match (current_type_id, t.as_ref()) {
                (Some(current_id), TypeInner::Var(var_id)) if current_id == var_id => {
                    // Self-referential optional type like "type o = opt o"
                    // In TypeScript, this becomes a recursive array type
                    str("[] | [").append(ident(current_id)).append("]")
                }
                _ => {
                    pp_ty(env, t, is_ref, current_type_id).append(str(" | undefined"))
                }
            }
        }
```

rust/candid_parser/tests/assets/actor.did
krpeacock: Sounds good!

rust/candid_parser/tests/assets/actor.did
```diff
@@ -1,11 +1,9 @@
 type f = func (int8) -> (int8);
 type g = f;
 type h = func (f) -> (f);
-type o = opt o;
```
ilbertt, line 4: Why was this type removed? I think it introduces a good test case for this PR. I know we have introduced the optionals test case, but I don't see any reason why we shouldn't keep this. One more test is never bad, and we avoid changing the goldenfiles for the other languages.

rust/candid_parser/src/bindings/typescript.rs
```diff
@@ -36,7 +41,26 @@ fn pp_ty<'a>(env: &'a TypeEnv, ty: &'a Type, is_ref: bool) -> RcDoc<'a> {
...
+                    "]",
+                ));
+            }
+
+            // for self referential check, check if the id is the same as the type that is inside the optional
+            match (parent_type_id, t.as_ref()) {
+                (Some(current_id), TypeInner::Var(var_id)) if current_id == var_id => {
+                    // Self-referential optional type like "type o = opt o"
+                    // In TypeScript, this becomes a recursive array type
+                    str("[] | [").append(ident(current_id)).append("]")
```
ilbertt, line 59: To keep it consistent with the code above:

```suggestion
                    str("[] | ").append(.append(enclose("[", ident(current_id), "]")
```

I would have rather created a function like `pp_opt` that does this, so that we avoid duplicating the code. If we use a dedicated function, the code also looks more self-explanatory, to show that those two cases are handled in the same way.

ilbertt (changes requested):

rust/candid_parser/src/bindings/typescript.rs
ilbertt: The extracted function would look like this:

```rust
fn pp_nested_opt<'a>(doc: RcDoc<'a>) -> RcDoc<'a> {
    str("[] | ").append(enclose("[", doc, "]"))
}

// then, on line 47:
return pp_nested_opt(pp_ty(env, t, is_ref, parent_type_id));

// and, on line 59:
pp_nested_opt(ident(current_id))
```

nathanosdev: Agreed, I like this.

rust/candid_parser/tests/assets/ok/example.d.ts
```diff
@@ -3,14 +3,17 @@ import type { ActorMethod } from '@dfinity/agent';
 import type { IDL } from '@dfinity/candid';
 
 export type A = B;
-export type B = [] | [A];
-export type List = [] | [{ 'head' : bigint, 'tail' : List }];
+export type B = A | undefined;
```
giorgiopiatti-caffeinated, line 6: This conversion is loosing type information, since the type id defined recursively, should be something along 
export type B = Some<A> | None.

Nested optional can arises even indirectly.

rust/candid_parser/tests/assets/ok/cyclic.d.ts
```diff
@@ -2,8 +2,8 @@ import type { Principal } from '@dfinity/principal';
 import type { ActorMethod } from '@dfinity/agent';
 import type { IDL } from '@dfinity/candid';
 
-export type A = [] | [B];
-export type B = [] | [C];
+export type A = B | undefined;
```
giorgiopiatti-caffeinated, line 5: This conversion is loosing type information, since the type id defined recursively, should be something along
```
export type A = Some<B> | None;
export type B = Some<C> | None;
```

Nested optional can arises even indirectly.

giorgiopiatti-caffeinated (changes requested):
I've looked through a few of the conversions I found some issues... 
The current conversion loses information when there are nested optionals, or type structures of nested optionals. 

I've attempted something similar as part of #601, which I then closed and maintaining somewhere else (DM to know more, is not a public repo).

PR #628, opened 2025-06-25
ci: add `tools.yaml` CI action

ilbertt:
Additionally, we add the `rust-toolchain.toml` file to set the Rust defaults and remove the deprecated `actions-rs/toolchain` step from `bench` and `rust` actions.

We also bump `pnpm` version to v10 in the `didc-js` workspace.

**Considerations**
We should
- refactor the `release` action as well, but this currently goes out of scope
- move to a standard JS repo structure with a `package.json` in the root, but this currently goes out of scope
- add tests to `didc` and run them in the CI

### DIFF
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -12,7 +12,15 @@ env:
   WORKING_DIR: tools/didc-js
 
 jobs:
-  build-and-test:
+  build-didc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build didc
+        run: cargo build -p didc --release
+
+  build-and-test-didc-js:
     runs-on: ubuntu-latest
     defaults:
       run:


.github/workflows/tools.yml
```diff
@@ -0,0 +1,33 @@
...
+      run:
+        shell: bash
+        working-directory: ${{ env.WORKING_DIR }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: "Setup PNPM"
+        uses: dfinity/ci-tools/actions/setup-pnpm@f241586372f71ae0875545bf5ce2aade82ebc0b5
+        with:
+          working-directory: ${{ env.WORKING_DIR }}
```
nathanosdev: Is the built-in `working-directory` not working here?
```suggestion
        working-directory: ${{ env.WORKING_DIR }}
```

ilbertt: Apparently, the working-directory does not work for `uses` steps

nathanosdev (approved):

PR #631, opened 2025-06-25
feat: add support for doc comments

ilbertt:
**Overview**
Inspired from #621.

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**


rust/candid_derive/src/func.rs
ilbertt, line 1: We'll add support for deriving comments in #633

PR #634, opened 2025-07-02
feat: support doc comments in Motoko bindgen

christoph-dfinity:
**Overview**
Adds support for reflecting doc comments from Candid declaration files to the Motoko generated bindings.

Internally, it adds the `IDLMergedProg` which is constructed when parsing the Candid declarations and is passed to the Motoko bindings generator. It containts the syntax types along with the associated comments, and it's traversed in parallel with the inner types by the binding generator.

Follow-up PRs will add support for doc comments in the other bindings generators.

Closes #625.


rust/candid/src/types/syntax.rs
```diff
@@ -14,6 +19,77 @@ pub enum IDLType {
     PrincipalT,
 }
 
+impl IDLType {
```
christoph-dfinity: Remember to get rid of these

rust/candid/src/types/syntax.rs
```diff
@@ -123,8 +199,126 @@ pub struct IDLProg {
...
+    }
+
+    pub fn decs(&self) -> Vec<Dec> {
+        self.typ_decs.iter().map(|b| Dec::TypD(b.clone())).collect()
+    }
+
+    pub fn actor(&self) -> Option<&IDLActorType> {
+        self.main_actor.as_ref()
+    }
+
```
christoph-dfinity:
```suggestion
```

Remember to remove this one

rust/candid/src/types/syntax.rs
ilbertt: Yep, just done

rust/candid/src/types/syntax.rs
ilbertt: Done

PR #635, opened 2025-07-02
feat: support line comments as doc comments

ilbertt:
**Overview**
Adds support for line comments (`// ...`) as doc comments in these places:
- above services:
  ```
  // This is a valid doc comment for the service
  service : {
    greet : (text) -> (text);
  }
  ```
- above actor methods:
  ```
  service : {
    // This is a valid doc comment for greet
    greet : (text) -> (text);
  }
  ```
- above type declarations:
  ```
  // This is a valid doc comment for type A
  type A = record {
    my_field : text;
  };
  ```
- above record and variant fields:
  ```
  type A = record {
    // This is a valid doc comment for my_field
    my_field : text;
  };

  type B = record {
    // This is a valid doc comment for nat element
    nat;
    text;
  }

  type C = variant {
    // This is a valid doc comment for my_variant_field
    my_variant_field : nat;
  };
  ```

Closes #631.
Closes #621.
Closes #620.
Closes #619.

**Considerations**
This doesn't break anything, as comments are invisible to the parser.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,46 @@
 
 * Non-breaking changes:
   + Supports parsing the arguments' names for `func` and `service` (init args).
+  + Supports collecting line comments as doc comments in the following cases:
+    - above services:
+      ```
+      // This is a valid doc comment for the service
+      service : {
+        greet : (text) -> (text);
+      }
+      ```
+    - above actor methods:
+      ```
+      service : {
+        // This is a valid doc comment for greet
+        greet : (text) -> (text);
+      }
+      ```
+    - above type declarations:
+      ```
+      // This is a valid doc comment for type A
+      type A = record {
+        my_field : text;
+      };
+      ```
+    - above record and variant fields:
+      ```
+      type A = record {
+        // This is a valid doc comment for my_field
+        my_field : text;
+      };
+
+      type B = record {
+        // This is a valid doc comment for nat element
+        nat;
+        text;
+      }
+
+      type C = variant {
+        // This is a valid doc comment for my_variant_field
+        my_variant_field : nat;
+      };
+      ```
 
 ### candid_derive
 


rust/candid/src/types/syntax.rs
```diff
@@ -94,6 +94,7 @@ impl IDLArgType {
 pub struct TypeField {
     pub label: Label,
     pub typ: IDLType,
+    pub doc_comment_lines: Vec<String>,
```
christoph-dfinity: With the rest of the file picking rather short names for their fields, let's maybe stay consistent and go with `docs`, or if we want to be more explicit just `doc_comments`?

rust/candid_parser/src/token.rs
```diff
@@ -1,10 +1,13 @@
...
+
 use lalrpop_util::ParseError;
 use logos::{Lexer, Logos};
 
 #[derive(Logos, Debug, Clone, PartialEq, Eq, Ord, PartialOrd)]
 #[logos(skip r"[ \t\r\n]+")]
-#[logos(skip r"//[^\n]*")] // line comment
 pub enum Token {
+    #[regex(r"//[^\n]*")]
+    DocComment,
```
christoph-dfinity: Rename to `LineComment` (it only becomes a doc-comment when the parser picks it up in certain places)

rust/candid_parser/src/token.rs
```diff
@@ -118,22 +121,40 @@ fn parse_number(lex: &mut Lexer<Token>) -> String {
     }
 }
 
+fn parse_doc_comment(lex: &Lexer<Token>) -> String {
+    lex.slice().trim_start_matches("///").trim().to_string()
```
christoph-dfinity:
```suggestion
    lex.slice().trim_start_matches("//").trim().to_string()
```

christoph-dfinity (changes requested):
Looks good, just some minor nits!

rust/candid/src/types/syntax.rs
ilbertt: I generally prefer to keep names verbose because they self-explain their purpose, but yeah I can shorten it

rust/candid_parser/src/token.rs
ilbertt: Done

rust/candid_parser/src/token.rs
ilbertt: I've already replaced it with a constant

christoph-dfinity (approved):
Nice!

PR #637, opened 2025-07-09
feat: support doc comments in Rust bindings generation

ilbertt:
**Overview**
Adds support for doc comments in the Rust bindings generator.

**Considerations**
Not all doc comments are supported due to how the Rust code is _nominalized_, as you can notice from the `rust/candid_parser/tests/assets/ok/example.rs` as compared to the `rust/candid_parser/tests/assets/ok/example.mo` Motoko one.

### DIFF
--- a/rust/candid/src/types/syntax.rs
+++ b/rust/candid/src/types/syntax.rs
@@ -146,7 +146,7 @@ pub struct IDLInitArgs {
     pub args: Vec<IDLArgType>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct IDLMergedProg {
     typ_decs: Vec<Binding>,
     main_actor: Option<IDLActorType>,
@@ -179,6 +179,10 @@ impl IDLMergedProg {
 
     pub fn decs(&self) -> Vec<Dec> {
         self.typ_decs.iter().map(|b| Dec::TypD(b.clone())).collect()
+    }
+
+    pub fn insert_binding(&mut self, id: String, typ: IDLType, docs: Vec<String>) {
+        self.typ_decs.push(Binding { id, typ, docs });
     }
 
     pub fn resolve_actor(&self) -> Result<Option<IDLActorType>> {
@@ -227,6 +231,25 @@ impl IDLMergedProg {
             typ,
             docs: top_level_docs,
         }))
+    }
+
+    pub fn resolve_actor_methods(&self) -> Result<Vec<Binding>> {
+        let typ = match self.resolve_actor()? {
+            Some(IDLActorType { typ, .. }) => typ,
+            None => return Ok(vec![]),
+        };
+
+        if let IDLType::ServT(methods) = typ {
+            Ok(methods)
+        } else if let IDLType::ClassT(_, inner) = typ {
+            if let IDLType::ServT(methods) = inner.as_ref() {
+                Ok(methods.to_vec())
+            } else {
+                Ok(vec![])
+            }
+        } else {
+            Ok(vec![])
+        }
     }
 
     // NOTE: We don't worry about cyclic type definitions, as we rule those out earlier when checking the type decs
--- a/rust/candid/src/types/syntax.rs
+++ b/rust/candid/src/types/syntax.rs
@@ -233,25 +233,6 @@ impl IDLMergedProg {
         }))
     }
 
-    pub fn resolve_actor_methods(&self) -> Result<Vec<Binding>> {
-        let typ = match self.resolve_actor()? {
-            Some(IDLActorType { typ, .. }) => typ,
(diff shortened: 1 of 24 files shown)


rust/candid_parser/src/bindings/rust.rs
```diff
@@ -372,31 +443,71 @@ fn test_{test_name}() {{
...
+            .pop_state(old, StateElem::TypeStr("std::result::Result"));
+        let old = self.state.push_state(&StateElem::Label("Ok"));
+        let ok = self.pp_ty(ok, is_ref);
+        self.state.pop_state(old, StateElem::Label("Ok"));
+        let old = self.state.push_state(&StateElem::Label("Err"));
+        let err = self.pp_ty(err, is_ref);
+        self.state.pop_state(old, StateElem::Label("Err"));
+        let body = ok.append(", ").append(err);
+        RcDoc::text(result).append(enclose("<", body, ">"))
+    }
```
ilbertt, line 474: Just extracted from `pp_ty`, nothing changed

rust/candid_parser/src/bindings/rust.rs
```diff
@@ -861,16 +1035,22 @@ impl NominalState<'_> {
                     } else {
                         path_to_var(path)
                     };
+                    let new_syntax =
+                        IDLType::RecordT(syntax_fields.map(|s| s.to_vec()).unwrap_or_default());
```
christoph-dfinity: This is quadratic in the depth of nesting for the type, as `.to_vec()` will perform a full clone.

rust/candid_parser/src/bindings/rust.rs
```diff
@@ -884,7 +1064,10 @@ impl NominalState<'_> {
                             } else {
                                 path.push(TypePath::VariantField(id.to_string()));
                             }
-                            let ty = self.nominalize(env, path, ty);
+                            let syntax_field = syntax_fields.and_then(|s| {
+                                s.iter().find(|f| f.label == **id).map(|f| f.typ.clone())
+                            });
```
christoph-dfinity: Also quadratic

rust/candid/src/types/syntax.rs
```diff
@@ -146,7 +146,7 @@ pub struct IDLInitArgs {
     pub args: Vec<IDLArgType>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
christoph-dfinity: Ideally we shouldn't be cloning this, as it's a very large structure. It's fine for the binding generator to consume this value (if it has to), as we don't need the merged prog afterwards.

christoph-dfinity (changes requested):
The solution you came up with for normalization is unfortunately quadratic in the depth of the nesting.
You could get around that by either remembering a mapping from `GeneratedName -> Vec<Path>` and then having a function that resolves a `Vec<Path>` against `IDLMergedProg`, or `GeneratedName -> &IDLType` (notice the reference type).
That would also avoid modifying the merged prog with synthetic bindings, which I think would be cleaner.

rust/candid_parser/src/bindings/rust.rs
ilbertt: Fixed

rust/candid_parser/src/bindings/rust.rs
ilbertt: Fixed

rust/candid/src/types/syntax.rs
ilbertt: Fixed

rust/candid_parser/src/bindings/rust.rs
```diff
@@ -372,31 +443,71 @@ fn test_{test_name}() {{
...
+    ) -> RcDoc<'b> {
         let old = self.state.push_state(&StateElem::TypeStr("variant"));
-        let fields: Vec<_> = fs.iter().map(|f| self.pp_variant_field(f)).collect();
+        let fields: Vec<_> = fs
+            .iter()
+            .map(|f| {
+                let (docs, syntax_field) = find_field(syntax, &f.id);
+                docs.append(self.pp_variant_field(f, syntax_field))
+            })
+            .collect();
```
christoph-dfinity, line 504: No need to collect this into a vector (allocation), as you're immediately turning it back into an iterator on the next line.

christoph-dfinity (approved):
Very nice! Just one small nit.

rust/candid_parser/src/bindings/rust.rs
christoph-dfinity: The implementation of `concat` requires `Clone` on the Iterator itself as you pointed out, so this doesn't work here. Nevermind...

PR #638, opened 2025-07-09
feat: support doc comments in TypeScript bindings generator

ilbertt:
**Overview**
Adds support for doc comments in the TypeScript bindings generator.

### DIFF
--- a/rust/candid_parser/src/bindings/typescript.rs
+++ b/rust/candid_parser/src/bindings/typescript.rs
@@ -1,7 +1,50 @@
-use super::javascript::{ident, is_tuple};
+use super::javascript::{ident, is_tuple_fields};
 use candid::pretty::utils::*;
+use candid::types::syntax::{self, IDLMergedProg, IDLType};
 use candid::types::{Field, Function, Label, SharedLabel, Type, TypeEnv, TypeInner};
 use pretty::RcDoc;
+
+const DOC_COMMENT_PREFIX: &str = "/**";
+const DOC_COMMENT_LINE_PREFIX: &str = " * ";
+const DOC_COMMENT_SUFFIX: &str = " */";
+
+fn find_field<'a>(
+    fields: Option<&'a [syntax::TypeField]>,
+    label: &'a Label,
+) -> (RcDoc<'a>, Option<&'a syntax::IDLType>) {
+    let mut docs = RcDoc::nil();
+    let mut syntax_field_ty = None;
+    if let Some(bs) = fields {
+        if let Some(field) = bs.iter().find(|b| b.label == *label) {
+            docs = pp_docs(&field.docs);
+            syntax_field_ty = Some(&field.typ);
+        }
+    };
+    (docs, syntax_field_ty)
+}
+
+fn pp_ty_rich<'a>(
+    env: &'a TypeEnv,
+    ty: &'a Type,
+    syntax: Option<&'a IDLType>,
+    is_ref: bool,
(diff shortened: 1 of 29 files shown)


rust/candid_parser/tests/assets/ok/example.mo
```diff
@@ -68,6 +68,7 @@ module {
     };
   };
   public type node = { head : Nat; tail : list };
+  /// Doc comment for res type
```
ilbertt: This change was cherry-picked from #637. We won't see it anymore if we merge that PR first

rust/candid_parser/tests/assets/example.did
```diff
@@ -27,6 +27,7 @@ type nested = record {
   40:nat;
   variant{ A; 0x2a; B; C };
 };
+// Doc comment for res type
```
ilbertt: This change was cherry-picked from #637. We won't see it anymore if we merge that PR first

rust/candid_parser/src/bindings/typescript.rs
```diff
@@ -108,10 +132,69 @@ fn pp_label(id: &SharedLabel) -> RcDoc {
...
+}
+
+fn pp_opt<'a>(
+    env: &'a TypeEnv,
+    ty: &'a Type,
+    syntax: Option<&'a IDLType>,
+    is_ref: bool,
+) -> RcDoc<'a> {
+    str("[] | ").append(enclose("[", pp_ty_rich(env, ty, syntax, is_ref), "]"))
 }
```
ilbertt, line 198: The logic inside these functions is the same that was in the match branches of the `pp_ty` before. The only addition is the usage of the syntax passed to `pp_field` (in `pp_record` and `pp_variant`) and the usage of `pp_ty_rich` (in `pp_opt`)

rust/candid_parser/src/bindings/typescript.rs
```diff
@@ -173,40 +287,53 @@ fn pp_defs<'a>(env: &'a TypeEnv, def_list: &'a [&'a str]) -> RcDoc<'a> {
...
+fn pp_actor<'a>(env: &'a TypeEnv, ty: &'a Type, syntax: Option<&'a IDLType>) -> RcDoc<'a> {
+    let service_doc = kwd("export interface _SERVICE");
     match ty.as_ref() {
-        TypeInner::Service(ref serv) => {
-            kwd("export interface _SERVICE").append(pp_service(env, serv))
-        }
-        TypeInner::Var(id) => kwd("export interface _SERVICE extends")
+        TypeInner::Service(_) => service_doc.append(pp_ty_rich(env, ty, syntax, false)),
+        TypeInner::Var(id) => service_doc
+            .append(kwd("extends"))
```
christoph-dfinity, line 302: Wouldn't this miss a space between `SERVICE` and `extends` after the refactor? If so let's also add a test to check.

ilbertt: Apparently, `append` does the trick as you can see from https://github.com/dfinity/candid/pull/638/files#diff-7b35686a5cb40a21e761a435c404d1262fed9ecd922ebe3e4db8d3bd58457ea5L16, which hasn't changed after the refactor

christoph-dfinity: It's actually because `kwd` appends a space, and so the `service_doc` document here already has a trailing space. Thanks for checking!

christoph-dfinity (approved):
Fire away!

PR #639, opened 2025-07-10
refactor: changes the implementation of `concat` to not require a clone-able iterator

christoph-dfinity:
Also refactors its callsites for clarity / to no longer allocate intermediate `Vec`s

### DIFF
--- a/rust/candid/src/pretty/utils.rs
+++ b/rust/candid/src/pretty/utils.rs
@@ -7,7 +7,8 @@ fn is_empty(doc: &RcDoc) -> bool {
     use pretty::Doc::*;
     match &**doc {
         Nil => true,
-        FlatAlt(t1, t2) | Union(t1, t2) => is_empty(t1) && is_empty(t2),
+        FlatAlt(t1, t2) => is_empty(t1) || is_empty(t2),
+        Union(t1, t2) => is_empty(t1) && is_empty(t2),
         Group(t) | Nest(_, t) | Annotated((), t) => is_empty(t),
         _ => false,
     }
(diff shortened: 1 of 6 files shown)


rust/candid/src/pretty/utils.rs
```diff
@@ -7,7 +7,8 @@ fn is_empty(doc: &RcDoc) -> bool {
     use pretty::Doc::*;
     match &**doc {
         Nil => true,
-        FlatAlt(t1, t2) | Union(t1, t2) => is_empty(t1) && is_empty(t2),
+        FlatAlt(t1, t2) => is_empty(t2) || is_empty(t1),
+        Union(t1, t2) => is_empty(t1) && is_empty(t2),
```
ilbertt, line 11: Was this a bug?

christoph-dfinity: Yes, for example `RcDoc::text("Hello").flat_alt(RcDoc::nil())` would be considered non empty by the previous behaviour.

What that document actually means is "Display 'Hello', unless you're layed out on a single line, then display empty". => Which means it's empty

The whole premise of `is_empty` is a hack, but I didn't want to try and remove it, as that would be a much more involved change.

christoph-dfinity:
> Shouldn't we check if all the other usages of concat are now aligned with this change?

I made updates to the other binding generators, but split out the changes to the candid bindings into their separate commit. I don't know how many conflicts they cause for your other PR so I didn't want to make too many changes there.

christoph-dfinity:
Thanks for the report! I'll revert the change.

PR #641, opened 2025-07-10
fix: support doc comments in `vec` inner types

ilbertt:
**Overview**
Adds support for doc comments in `vec` inner types for Motoko and TypeScript bindings generators.
Rust already supports them.


christoph-dfinity (approved):
LGTM

PR #648, opened 2025-07-14
fix: make special opt subtyping rule warning more explicit

ilbertt:
**Overview**
Explains better the special opt subtyping rule warning, adding a link to the spec.

### DIFF
--- a/rust/candid/src/types/subtype.rs
+++ b/rust/candid/src/types/subtype.rs
@@ -69,7 +69,7 @@ fn subtype_(
             Ok(())
         }
         (_, Opt(_)) => {
-            let msg = format!("FIX ME! {t1} <: {t2} via special opt rule.\nThis means the sender and receiver type has diverged, and can cause data loss.");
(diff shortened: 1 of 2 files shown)


christoph-dfinity (approved):

rust/candid/src/types/subtype.rs
```diff
@@ -69,7 +69,7 @@ fn subtype_(
             Ok(())
         }
         (_, Opt(_)) => {
-            let msg = format!("FIX ME! {t1} <: {t2} via special opt rule.\nThis means the sender and receiver type has diverged, and can cause data loss.");
+            let msg = format!("WARNING: {t1} <: {t2} due to special subtyping rules involving optional types/fields (see https://github.com/dfinity/candid/blob/master/spec/Candid.md#upgrading-and-subtyping). This means the two interfaces have diverged, which could cause data loss.");
```
mraszyk: Using a "master" link bears the risk of getting dead, pinning down to a commit bears the risk of getting outdated. I guess the former is better, but pointing it out nonetheless.

ilbertt: Fixed

PR #640, opened 2025-07-10
feat: support doc comments in Candid bindings generator

ilbertt:
**Overview**
Adds support for doc comments in the Candid bindings generator.

PR #612, opened 2025-06-10
feat: support argument names

ilbertt:
**Overview**
Adds support for named arguments: if a `.did` Candid declaration file declares functions, classes or services with named arguments, those names are parsed by the `candid-parser` and preserved in the generated bindings.
At the same time, the names of the arguments are preserved if the Candid declaration file is generated from a Rust canister.

**Requirements**
To consider this feature complete, the Motoko compiler should preserve the actor methods names from Motoko canisters to the Candid declaration files.

**Considerations**
This change should be backwards compatible.

Picked up from https://github.com/ilbertt/candid/tree/luca/arg-names

PR #652, opened 2025-07-25
feat: introduce new methods to avoid breaking API change in candid crate

lwshang:
This PR refactors recent changes (#612 and #640) to avoid a major version bump in `candid` crate.
> * Breaking changes:
>   + `pp_args` and `pp_init_args` now require a `&[ArgType]` parameter. The `pp_rets` function has been added, with the signature of the old `pp_args`.
>   + `pretty::candid::pp_label` now takes a `&Label` parameter, instead of a `&SharedLabel`.

The previously altered API (`pp_args`, `pp_init_args`, `pp_label`) has been restored to maintain backward compatibility. 

The intended new behavior is now exposed via new, non-breaking methods: `pp_named_args`, `pp_named_init_args` and `pp_label_raw`.

### DIFF
--- a/rust/candid/src/pretty/candid.rs
+++ b/rust/candid/src/pretty/candid.rs
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 
 use crate::pretty::utils::*;
-use crate::types::{ArgType, Field, FuncMode, Function, Label, Type, TypeEnv, TypeInner};
+use crate::types::{
+    ArgType, Field, FuncMode, Function, Label, SharedLabel, Type, TypeEnv, TypeInner,
+};
 use pretty::RcDoc;
 
 static KEYWORDS: [&str; 30] = [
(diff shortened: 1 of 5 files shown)


rust/candid/src/types/subtype.rs
ilbertt, line 1: Even if behind the `printer` feature flag, `pp_args` accepting `&[ArgType]` is a breaking change. Same for the `pp_args` when printer feature is disabled. See:

- https://github.com/dfinity/candid/compare/2024-05-14...master#diff-07230204a6f223a07a547ad828596f33c98c0d14c1079ced96eb66f6f75b016dL280
- https://github.com/dfinity/candid/compare/2024-05-14...master#diff-07230204a6f223a07a547ad828596f33c98c0d14c1079ced96eb66f6f75b016dL291

rust/candid/src/pretty/candid.rs
```diff
@@ -147,16 +156,19 @@ fn pp_fields(fs: &[Field], is_variant: bool) -> RcDoc {
 }
 
 pub fn pp_function(func: &Function) -> RcDoc {
-    let args = pp_args(&func.args);
-    let rets = pp_rets(&func.rets);
+    let args = pp_named_args(&func.args);
+    let rets = pp_args(&func.rets);
```
ilbertt: I would keep the `pp_rets` function, even just as a wrapper of `pp_args` because it feels so weird to print return types using the `pp_args` function.

ilbertt (commented):
Makes sense! Just a couple of comments

rust/candid/src/types/subtype.rs
lwshang: The links didn't show me the expected diff. I assume you're referring to the `pp_args` method in `subtype.rs` (in the bottom). Since it's not a public API, we should be able to change its signature without issue.

ilbertt: Oh, GitHub messing up with the UI. Anyway, yes, you are right. I've resolved this comment

PR #663, opened 2025-07-29
build: update crates versions to fix rust/bench builds

ilbertt:
**Overview**
Pick changes to `Cargo.toml` files from #661 to make rust/bench crate build properly.

**Considerations**
Bench and lint will be fixed in #661.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,12 @@ anyhow = "1.0"
 rand = "0.8"
 arbitrary = "1.3"
 console = "0.15"
+
+[workspace.lints.clippy]
+all = { level = "deny", priority = 0 }
+uninlined_format_args = { level = "allow", priority = 1 }
+
+[workspace.lints.rust]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -23,3 +23,12 @@ anyhow = "1.0"
 rand = "0.8"
 arbitrary = "1.3"
 console = "0.15"
+
+[workspace.lints.clippy]
+all = { level = "deny", priority = 0 }
```
adamspofford: This is usually not a good idea. -Dwarnings in CI does everything this does (you can't merge without fixing the lints), but this makes developing more annoying since you can't build *locally* without fixing the lints.

ilbertt: I see, I'll remove it.

For the record, we've had it since my https://github.com/dfinity/candid/pull/632

adamspofford-dfinity (approved):

PR #664, opened 2025-07-29
chore: Release 2025-07-29

adamspofford-dfinity:
candid: 0.10.14 -> 0.10.16
candid_derive: 0.6.6 -> 0.10.16
candid_parser: 0.1.4 -> 0.2.1
didc: 0.4.0 -> 0.5.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,15 @@
 
 ## Unreleased
 
-### Candid
+## 2025-07-29
+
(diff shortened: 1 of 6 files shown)


ilbertt (changes requested):
Looks good.

Can you update the `rust/bench/Cargo.lock` file as well? Running the following commands should be enough:

```shell
cd rust/bench
cargo build --release --target wasm32-unknown-unknown
```

PR #665, opened 2025-07-30
improving the JavaScript target with exporting of the IDL type object…

lastmjs:
…s and deprecations of the factory functions

**Overview**

**Requirements**

**Considered Solutions**

**Recommended Solution**

**Considerations**

### DIFF
--- a/rust/candid_parser/src/bindings/javascript.rs
+++ b/rust/candid_parser/src/bindings/javascript.rs
@@ -200,11 +200,17 @@ fn pp_defs<'a>(
     env: &'a TypeEnv,
     def_list: &'a [&'a str],
     recs: &'a BTreeSet<&'a str>,
+    export: bool,
 ) -> RcDoc<'a> {
-    let recs_doc = lines(
-        recs.iter()
-            .map(|id| kwd("const").append(ident(id)).append(" = IDL.Rec();")),
-    );
+    let export_prefix = if export { str("export ") } else { RcDoc::nil() };
+
+    let recs_doc = lines(recs.iter().map(|id| {
+        export_prefix
+            .clone()
+            .append(kwd("const"))
+            .append(ident(id))
+            .append(" = IDL.Rec();")
+    }));
     let defs = lines(def_list.iter().map(|id| {
         let ty = env.find_type(id).unwrap();
         if recs.contains(id) {
@@ -212,7 +218,9 @@ fn pp_defs<'a>(
                 .append(".fill")
                 .append(enclose("(", pp_ty(ty), ");"))
         } else {
-            kwd("const")
+            export_prefix
+                .clone()
+                .append(kwd("const"))
                 .append(ident(id))
                 .append(" = ")
                 .append(pp_ty(ty))
@@ -237,38 +245,84 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
     }
 }
 
+fn pp_deprecation_comment<'a>() -> RcDoc<'a> {
+    str("/**").append(RcDoc::hardline())
+        .append(" * @deprecated Import IDL types directly from this module instead of using this factory function.")
+        .append(RcDoc::hardline())
+        .append(" */")
+        .append(RcDoc::hardline())
+}
+
 pub fn compile(env: &TypeEnv, actor: &Option<Type>) -> String {
     match actor {
         None => {
             let def_list: Vec<_> = env.0.iter().map(|pair| pair.0.as_ref()).collect();
             let recs = infer_rec(env, &def_list).unwrap();
-            let doc = pp_defs(env, &def_list, &recs);
-            doc.pretty(LINE_WIDTH).to_string()
+            let import_doc = str("import { IDL } from '@dfinity/candid';");
+            let doc = pp_defs(env, &def_list, &recs, true);
+            let result = import_doc
+                .append(RcDoc::hardline())
+                .append(RcDoc::hardline())
+                .append(doc)
+                .pretty(LINE_WIDTH)
+                .to_string();
+
+            result
         }
         Some(actor) => {
             let def_list = chase_actor(env, actor).unwrap();
             let recs = infer_rec(env, &def_list).unwrap();
-            let defs = pp_defs(env, &def_list, &recs);
             let init = if let TypeInner::Class(ref args, _) = actor.as_ref() {
                 args.as_slice()
             } else {
                 &[][..]
             };
-            let actor = kwd("return").append(pp_actor(actor, &recs)).append(";");
-            let body = defs.append(actor);
-            let doc = str("export const idlFactory = ({ IDL }) => ")
-                .append(enclose_space("{", body, "};"));
-            // export init args
+
+            let import_doc = str("import { IDL } from '@dfinity/candid';");
+            let defs = pp_defs(env, &def_list, &recs, true);
+            let actor = pp_actor(actor, &recs);
+
+            let idl_service = str("export const idlService = ")
+                .append(actor.clone())
+                .append(";");
+
+            let idl_init_args = str("export const idlInitArgs = ")
+                .append(pp_rets(init))
+                .append(";");
+
+            let idl_factory_return = kwd("return").append(actor).append(";");
+            let idl_factory_body = pp_defs(env, &def_list, &recs, false).append(idl_factory_return);
+            let idl_factory_doc = pp_deprecation_comment()
+                .append(str("export const idlFactory = ({ IDL }) => "))
+                .append(enclose_space("{", idl_factory_body, "};"));
+
             let init_defs = chase_types(env, init).unwrap();
             let init_recs = infer_rec(env, &init_defs).unwrap();
-            let init_defs_doc = pp_defs(env, &init_defs, &init_recs);
+            let init_defs_doc = pp_defs(env, &init_defs, &init_recs, false);
             let init_doc = kwd("return").append(pp_rets(init)).append(";");
             let init_doc = init_defs_doc.append(init_doc);
-            let init_doc =
-                str("export const init = ({ IDL }) => ").append(enclose_space("{", init_doc, "};"));
+            let init_doc = pp_deprecation_comment()
+                .append(str("export const init = ({ IDL }) => "))
+                .append(enclose_space("{", init_doc, "};"));
             let init_doc = init_doc.pretty(LINE_WIDTH).to_string();
-            let doc = doc.append(RcDoc::hardline()).append(init_doc);
-            doc.pretty(LINE_WIDTH).to_string()
+
+            let result = import_doc
+                .append(RcDoc::hardline())
+                .append(RcDoc::hardline())
+                .append(defs)
+                .append(RcDoc::hardline())
+                .append(idl_service)
+                .append(RcDoc::hardline())
+                .append(RcDoc::hardline())
+                .append(idl_init_args)
+                .append(RcDoc::hardline())
+                .append(RcDoc::hardline())
+                .append(idl_factory_doc)
+                .append(RcDoc::hardline())
+                .append(RcDoc::hardline())
+                .append(init_doc);
+
+            result.pretty(LINE_WIDTH).to_string()
         }
     }
 }
--- a/rust/candid_parser/tests/assets/ok/actor.js
+++ b/rust/candid_parser/tests/assets/ok/actor.js
@@ -1,3 +1,24 @@
+import { IDL } from '@dfinity/candid';
+
+export const o = IDL.Rec();
+export const f = IDL.Func([IDL.Int8], [IDL.Int8], []);
+export const h = IDL.Func([f], [f], []);
+export const g = f;
+o.fill(IDL.Opt(o));
+
+export const idlService = IDL.Service({
+  'f' : IDL.Func([IDL.Nat], [h], []),
+  'g' : f,
+  'h' : g,
+  'h2' : h,
+  'o' : IDL.Func([o], [o], []),
+});
+
+export const idlInitArgs = [];
+
+/**
+ * @deprecated Import IDL types directly from this module instead of using this factory function.
+ */
 export const idlFactory = ({ IDL }) => {
   const o = IDL.Rec();
   const f = IDL.Func([IDL.Int8], [IDL.Int8], []);
@@ -12,4 +33,8 @@ export const idlFactory = ({ IDL }) => {
     'o' : IDL.Func([o], [o], []),
   });
 };
+
+/**
+ * @deprecated Import IDL types directly from this module instead of using this factory function.
+ */
 export const init = ({ IDL }) => { return []; };
--- a/rust/candid_parser/tests/assets/ok/class.js
+++ b/rust/candid_parser/tests/assets/ok/class.js
@@ -1,3 +1,19 @@
+import { IDL } from '@dfinity/candid';
+
+export const List = IDL.Rec();
+List.fill(IDL.Opt(IDL.Tuple(IDL.Int, List)));
+export const Profile = IDL.Record({ 'age' : IDL.Nat8, 'name' : IDL.Text });
+
+export const idlService = IDL.Service({
+  'get' : IDL.Func([], [List], []),
+  'set' : IDL.Func([List], [List], []),
+});
+
+export const idlInitArgs = [IDL.Int, List, Profile];
+
+/**
+ * @deprecated Import IDL types directly from this module instead of using this factory function.
+ */
 export const idlFactory = ({ IDL }) => {
   const List = IDL.Rec();
   List.fill(IDL.Opt(IDL.Tuple(IDL.Int, List)));
@@ -7,6 +23,10 @@ export const idlFactory = ({ IDL }) => {
     'set' : IDL.Func([List], [List], []),
   });
 };
+
+/**
+ * @deprecated Import IDL types directly from this module instead of using this factory function.
+ */
 export const init = ({ IDL }) => {
   const List = IDL.Rec();
   List.fill(IDL.Opt(IDL.Tuple(IDL.Int, List)));
--- a/rust/candid_parser/tests/assets/ok/comment.js
+++ b/rust/candid_parser/tests/assets/ok/comment.js
@@ -1,2 +1,4 @@
-const id = IDL.Nat8;
+import { IDL } from '@dfinity/candid';
+
+export const id = IDL.Nat8;
 
--- a/rust/candid_parser/tests/assets/ok/cyclic.js
+++ b/rust/candid_parser/tests/assets/ok/cyclic.js
@@ -1,3 +1,22 @@
+import { IDL } from '@dfinity/candid';
+
+export const A = IDL.Rec();
+export const C = A;
+export const B = IDL.Opt(C);
+A.fill(IDL.Opt(B));
+export const Z = A;
+export const Y = Z;
+export const X = Y;
+
+export const idlService = IDL.Service({
+  'f' : IDL.Func([A, B, C, X, Y, Z], [], []),
+});
+
+export const idlInitArgs = [];
+
+/**
+ * @deprecated Import IDL types directly from this module instead of using this factory function.
+ */
 export const idlFactory = ({ IDL }) => {
   const A = IDL.Rec();
   const C = A;
@@ -8,4 +27,8 @@ export const idlFactory = ({ IDL }) => {
   const X = Y;
   return IDL.Service({ 'f' : IDL.Func([A, B, C, X, Y, Z], [], []) });
 };
+
+/**
+ * @deprecated Import IDL types directly from this module instead of using this factory function.
+ */
 export const init = ({ IDL }) => { return []; };
--- a/rust/candid_parser/tests/assets/ok/empty.js
+++ b/rust/candid_parser/tests/assets/ok/empty.js
@@ -1,3 +1,23 @@
+import { IDL } from '@dfinity/candid';
+
+export const T = IDL.Rec();
+T.fill(IDL.Tuple(T));
+
+export const idlService = IDL.Service({
+  'f' : IDL.Func([IDL.Record({})], [IDL.Variant({})], []),
+  'g' : IDL.Func([T], [IDL.Variant({ 'a' : T })], []),
+  'h' : IDL.Func(
+      [IDL.Tuple(T, IDL.Empty)],
+      [IDL.Variant({ 'a' : T, 'b' : IDL.Record({}) })],
+      [],
+    ),
+});
+
+export const idlInitArgs = [];
+
+/**
+ * @deprecated Import IDL types directly from this module instead of using this factory function.
+ */
 export const idlFactory = ({ IDL }) => {
   const T = IDL.Rec();
   T.fill(IDL.Tuple(T));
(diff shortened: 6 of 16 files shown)


rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -237,38 +245,84 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
+}
+
 pub fn compile(env: &TypeEnv, actor: &Option<Type>) -> String {
     match actor {
         None => {
             let def_list: Vec<_> = env.0.iter().map(|pair| pair.0.as_ref()).collect();
             let recs = infer_rec(env, &def_list).unwrap();
-            let doc = pp_defs(env, &def_list, &recs);
-            doc.pretty(LINE_WIDTH).to_string()
+            let import_doc = str("import { IDL } from '@dfinity/candid';");
```
ilbertt: I would extract this into a function (e.g. `pp_imports`) so that we can reuse it:

```rust
fn pp_imports<'a>() -> RcDoc<'a> {
    str("import { IDL } from '@dfinity/candid';")
        .append(RcDoc::hardline())
        .append(RcDoc::hardline())
}
```

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -237,38 +245,84 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
-            let init_defs_doc = pp_defs(env, &init_defs, &init_recs);
+            let init_defs_doc = pp_defs(env, &init_defs, &init_recs, false);
             let init_doc = kwd("return").append(pp_rets(init)).append(";");
             let init_doc = init_defs_doc.append(init_doc);
-            let init_doc =
-                str("export const init = ({ IDL }) => ").append(enclose_space("{", init_doc, "};"));
+            let init_doc = pp_deprecation_comment()
+                .append(str("export const init = ({ IDL }) => "))
+                .append(enclose_space("{", init_doc, "};"));
             let init_doc = init_doc.pretty(LINE_WIDTH).to_string();
```
ilbertt: I would now extract this into a function as well, e.g. `pp_init_args`

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -237,38 +245,84 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
+
+            let idl_init_args = str("export const idlInitArgs = ")
+                .append(pp_rets(init))
+                .append(";");
+
+            let idl_factory_return = kwd("return").append(actor).append(";");
+            let idl_factory_body = pp_defs(env, &def_list, &recs, false).append(idl_factory_return);
+            let idl_factory_doc = pp_deprecation_comment()
+                .append(str("export const idlFactory = ({ IDL }) => "))
+                .append(enclose_space("{", idl_factory_body, "};"));
```
ilbertt: Same: I would now extract this into a function.

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -237,38 +245,84 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
+            let defs = pp_defs(env, &def_list, &recs, true);
+            let actor = pp_actor(actor, &recs);
+
+            let idl_service = str("export const idlService = ")
+                .append(actor.clone())
+                .append(";");
+
+            let idl_init_args = str("export const idlInitArgs = ")
+                .append(pp_rets(init))
+                .append(";");
```
ilbertt: We can also extract this into a function

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -237,38 +245,84 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
+
+            let idl_factory_return = kwd("return").append(actor).append(";");
+            let idl_factory_body = pp_defs(env, &def_list, &recs, false).append(idl_factory_return);
+            let idl_factory_doc = pp_deprecation_comment()
+                .append(str("export const idlFactory = ({ IDL }) => "))
```
ilbertt: I would now rename the `IDL` parameter to something like:

```suggestion
                .append(str("export const idlFactory = ({ IDL: IDLModule }) => "))
```

to avoid variable shadowing and make sure we are using the IDL passed as argument inside the `idlFactory` and the `init` functions, and **not** the top level imported `IDL` obect instead.

This requires us to pass a new argument (e.g. `idl_module_name: 'a str`) around all the nested functions until `pp_ty`, but I think it's worth it. Maybe you can find a better way to avoid passing this argument around.

ilbertt (changes requested):
Just to confirm: the use case you are trying to solve is not about being able to programmatically access the args and return types inside functions, right? Otherwise, we may need to export a `serviceMethods` object, similar to what @nathanosdev was doing in https://github.com/dfinity/candid/pull/610/files#diff-c0a840e21a25105098350b8bb524fa19d48e84949aed98ccbf945da739a1ed79R263-R274.

Btw, we can now make use of the doc comments (similar to what we added in the `ts` target in https://github.com/dfinity/candid/pull/638). I can add them in a separate PR though, so that we keep this one small.

rust/candid_parser/tests/assets/ok/actor.js
```diff
@@ -1,3 +1,24 @@
+import { IDL } from '@dfinity/candid';
+
+export const o = IDL.Rec();
+export const f = IDL.Func([IDL.Int8], [IDL.Int8], []);
+export const h = IDL.Func([f], [f], []);
+export const g = f;
+o.fill(IDL.Opt(o));
+
+export const idlService = IDL.Service({
```
nathanosdev, line 9: There's no corresponding `.d.ts` change for `idlService` and `idlInitArgs`, or any of the newly exported properties above. The objects will be available at runtime to JavaScript, but TypeScript won't recognize them at compile time.

rust/candid_parser/src/bindings/javascript.rs
nathanosdev: I think this is fine as-is to avoid the complexity.

Avoiding variable shadowing is a stylistic choice and this function is deprecated now anyway. I don't think it's worth introducing additional complexity for a stylistic choice in a deprecated function.

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -237,38 +245,84 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
     }
 }
 
+fn pp_deprecation_comment<'a>() -> RcDoc<'a> {
+    str("/**").append(RcDoc::hardline())
+        .append(" * @deprecated Import IDL types directly from this module instead of using this factory function.")
```
ilbertt: We should also add a note that this is possible only on @dfinity/candid v3.0.1, something like:

```suggestion
        .append(" * @deprecated Since `@dfinity/candid` v3.0.1, you can import IDL types directly from this module instead of using this factory function.")
```

rust/candid_parser/src/bindings/javascript.rs
lastmjs: I went down this road and decided against it because of Rust ownership issues.

rust/candid_parser/src/bindings/javascript.rs
lastmjs: I went down this road and decided against it because of Rust ownership issues.

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Which part exactly? I really don't think it's worth breaking these out into functions, from my experience it's not as simple or as beneficial as it seems.

rust/candid_parser/src/bindings/javascript.rs
lastmjs: We should also add these to the TypeScript types

rust/candid_parser/src/bindings/javascript.rs
ilbertt: Let's keep it as it is for now

rust/candid_parser/tests/assets/ok/actor.js
lastmjs: `or any of the newly exported properties above` what are you referring to here?

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/tests/assets/ok/actor.js
lastmjs: I believe I have finished this now, as the `.d.ts` files already seem to be exporting the Candid types themselves, and I've just added the `idlService` and `idlInitargs` variables.

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -238,39 +246,87 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
+
+            let import_doc = str("import { IDL } from '@dfinity/candid';");
+            let defs = pp_defs(env, &def_list, &recs, true);
+            let actor = pp_actor(actor, &recs);
+
+            let idl_service = str("export const idlService = ")
+                .append(actor.clone())
+                .append(";");
+
+            let idl_init_args = str("export const idlInitArgs = ")
```
lastmjs, line 277: Would a better name be `idlInitParams`? Generally arguments are the actual runtime values, and parameters are the variables in a function or method signature.

The `IDL` type objects are really parameters describing the types of the values that will be passed into a service during initialization and post upgrade.

Right?

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -238,39 +246,87 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
     }
 }
 
+pub fn pp_deprecation_comment<'a>() -> RcDoc<'a> {
+    str("/**").append(RcDoc::hardline())
+        .append(" * @deprecated Since `@dfinity/candid` v3.2.1, you can import IDL types directly from this module instead of using this factory function.")
```
ilbertt: See https://js.icp.build/core/changelog#301---2025-07-22:
```suggestion
        .append(" * @deprecated Since `@dfinity/candid` v3.0.1, you can import IDL types directly from this module instead of using this factory function.")
```

rust/candid_parser/src/bindings/javascript.rs
ilbertt: I would still extract this into a function or constant, so that we can change it in one place when needed (probably soon). Shouldn't give any issues with Rust ownerships

rust/candid_parser/tests/assets/ok/actor.js
ilbertt: I think this is fine now

rust/candid_parser/src/bindings/javascript.rs
ilbertt: Params makes more sense for me as well, but as far as I know we've always referred to the initialization parameters as the "init args". I believe the reason is that those params are passed to the init function exposed by the canister, so you can also consider them "runtime values".

In any case, I would not change the terminology to avoid confusion.

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -238,39 +246,87 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
     }
 }
 
+pub fn pp_deprecation_comment<'a>() -> RcDoc<'a> {
```
ilbertt: Not really important, but since we use it in the typescript bindings as well, this could be called something like:
```suggestion
pub fn pp_idl_factory_deprecation_comment<'a>() -> RcDoc<'a> {
```

ilbertt (commented):
I've updated the `next` branch now. Can you merge it into your branch (again)?

For the rest, I just left a few minor comments. I think we are really close to complete it!

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/src/bindings/javascript.rs
lastmjs: I'm confused. It will only be with the next release that has this code in it that will allow developers to import the IDL types directly from this module instead of using the factory functions.

As of `v3.0.1` the IDL types are still stuck in the generated code without being able to export them.

lastmjs: Unless you would like me to change the wording to keep `v3.0.1` and just explain that factory functions are no longer needed in general because of the changes in the version? So ignore being able to export from the module? It's just that this module is where these functions were used from, so without the `export` keywords I don't really feel like the factory functions are deprecated.

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/src/bindings/javascript.rs
lastmjs: I believe this is the last comment to resolve, let me know what you think

ilbertt: The agent-js `v3.0.1` (aka [@dfinity/candid@3.0.1](https://www.npmjs.com/package/@dfinity/candid/v/3.0.1)) was the first version to have the fix for `instanceof`. This fix allows devs to import IDL types from wherever, including the JS bindings generated by `dfx`/`didc`, which are these bindings. The agent-js version is not linked to this repo.

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -230,39 +238,91 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
         None => {
             let def_list: Vec<_> = env.to_sorted_iter().map(|pair| pair.0.as_str()).collect();
             let recs = infer_rec(env, &def_list).unwrap();
-            let doc = pp_defs(env, &def_list, &recs);
-            doc.pretty(LINE_WIDTH).to_string()
+            let import_doc = pp_imports();
+            let doc = pp_defs(env, &def_list, &recs, true);
+            let result = import_doc
+                .append(RcDoc::hardline())
+                .append(RcDoc::hardline())
```
ilbertt:
```suggestion
            let doc = pp_defs(env, &def_list, &recs, true);
            let result = pp_imports()
```

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -230,39 +238,91 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
     }
 }
 
+fn pp_imports<'a>() -> RcDoc<'a> {
+    str("import { IDL } from '@dfinity/candid';")
```
ilbertt, line 245:
```suggestion
    str("import { IDL } from '@dfinity/candid';")
        .append(RcDoc::hardline())
        .append(RcDoc::hardline())
```

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -230,39 +238,91 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
+            let init_doc = pp_idl_factory_deprecation_comment()
+                .append(str("export const init = ({ IDL }) => "))
+                .append(enclose_space("{", init_doc, "};"));
             let init_doc = init_doc.pretty(LINE_WIDTH).to_string();
-            let doc = doc.append(RcDoc::hardline()).append(init_doc);
-            doc.pretty(LINE_WIDTH).to_string()
+
+            let result = import_doc
+                .append(RcDoc::hardline())
+                .append(RcDoc::hardline())
```
ilbertt:
```suggestion
            let result = pp_imports()
```

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -230,39 +238,91 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
             let init = types.as_slice();
-            let actor = kwd("return").append(pp_actor(actor, &recs)).append(";");
-            let body = defs.append(actor);
-            let doc = str("export const idlFactory = ({ IDL }) => ")
-                .append(enclose_space("{", body, "};"));
-            // export init args
-            let init_defs = chase_types(env, init).unwrap();
+            let init_types: Vec<Type> = init.iter().map(|a| a.clone()).collect();
+
+            let import_doc = pp_imports();
```
ilbertt:
```suggestion
```

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -230,39 +238,91 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
                 Vec::new()
             };
             let init = types.as_slice();
-            let actor = kwd("return").append(pp_actor(actor, &recs)).append(";");
-            let body = defs.append(actor);
-            let doc = str("export const idlFactory = ({ IDL }) => ")
-                .append(enclose_space("{", body, "};"));
-            // export init args
-            let init_defs = chase_types(env, init).unwrap();
+            let init_types: Vec<Type> = init.iter().map(|a| a.clone()).collect();
```
ilbertt: This Vec is not necessary, you can pass the `init` slice around. Feel free to rename it to `init_types`
```suggestion
            let init = types.as_slice();
```

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -230,39 +238,91 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
+            let idl_factory_body = pp_defs(env, &def_list, &recs, false).append(idl_factory_return);
+            let idl_factory_doc = pp_idl_factory_deprecation_comment()
+                .append(str("export const idlFactory = ({ IDL }) => "))
+                .append(enclose_space("{", idl_factory_body, "};"));
+
+            let init_defs = chase_types(env, &init_types).unwrap();
             let init_recs = infer_rec(env, &init_defs).unwrap();
-            let init_defs_doc = pp_defs(env, &init_defs, &init_recs);
-            let init_doc = kwd("return").append(pp_rets(init)).append(";");
+            let init_defs_refs: Vec<&str> = init_defs.iter().map(|s| *s).collect();
```
ilbertt: Not necessary, you can pass the `init_defs` vec around
```suggestion
```

ilbertt (changes requested):

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Removed the deprecation comment

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

ilbertt (approved):
I plan to test these changes in a fresh project that uses both agent-js v2 and v3, to check that we don't have regressions. Once done, I'll merge the PR.

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -193,19 +193,27 @@ fn pp_defs<'a>(
     env: &'a TypeEnv,
     def_list: &'a [&'a str],
     recs: &'a BTreeSet<&'a str>,
+    export: bool,
 ) -> RcDoc<'a> {
-    let recs_doc = lines(
-        recs.iter()
-            .map(|id| kwd("const").append(ident(id)).append(" = IDL.Rec();")),
-    );
+    let export_prefix = if export { str("export ") } else { RcDoc::nil() };
```
ilbertt, line 198: We should add an empty line only if there are type definitions:

```rust
    let export_prefix = if export { str("export ") } else { RcDoc::nil() };

    let recs_doc = lines(recs.iter().map(|id| {
        export_prefix
            .clone()
            .append(kwd("const"))
            .append(ident(id))
            .append(" = IDL.Rec();")
    }));
    let mut defs = lines(def_list.iter().map(|&id| {
        let ty = env.find_type(&id.into()).unwrap();
        if recs.contains(id) {
            ident(id)
                .append(".fill")
                .append(enclose("(", pp_ty(ty), ");"))
        } else {
            export_prefix
                .clone()
                .append(kwd("const"))
                .append(ident(id))
                .append(" = ")
                .append(pp_ty(ty))
                .append(";")
        }
    }));
    if !def_list.is_empty() {
        defs = defs.append(RcDoc::hardline())
    }
    recs_doc.append(defs)
```

rust/candid_parser/src/bindings/javascript.rs
```diff
@@ -230,39 +238,72 @@ fn pp_actor<'a>(ty: &'a Type, recs: &'a BTreeSet<&'a str>) -> RcDoc<'a> {
...
             let init_doc = init_defs_doc.append(init_doc);
             let init_doc =
                 str("export const init = ({ IDL }) => ").append(enclose_space("{", init_doc, "};"));
             let init_doc = init_doc.pretty(LINE_WIDTH).to_string();
-            let doc = doc.append(RcDoc::hardline()).append(init_doc);
-            doc.pretty(LINE_WIDTH).to_string()
+
+            let result = pp_imports()
+                .append(defs)
+                .append(RcDoc::hardline())
```
ilbertt: See my comment above
```suggestion
```

ilbertt (commented):
If there are no type definitions, the generated file has one more empty line between import statements and export const idlService. Not a big deal, can be left as it is.

ilbertt:
> I plan to test these changes in a fresh project that uses both agent-js v2 and v3, to check that we don't have regressions. Once done, I'll merge the PR.

Update: I've tested the generated bindings with both agent-js v2 and v3, locally and on mainnet, and I couldn't find any regression.

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

rust/candid_parser/src/bindings/javascript.rs
lastmjs: Done

PR #668, opened 2025-08-04
refactor: pretty printing combinator for lists/tuples

christoph-dfinity:
Gets rid of the dodgy `is_empty` function that performed full traversals of documents to tell if they were empty.

Fixes #667

PR #669, opened 2025-08-04
fix: fixes pretty printing the concatenation of an empty list of documents

### LINKED ISSUE #667: Error: Broken function concat
The `concat` function is broken in version 0.10.16. It inserts a separator even for an empty list, see the example:

```rust
        let t = concat([1, 2].iter().map(|_x| RcDoc::text("2")), ",")
            .pretty(LINE_WIDTH)
            .to_string();

        println!("{t}");
        let t = concat([1].iter().map(|_x| RcDoc::text("1")), ",")
            .pretty(LINE_WIDTH)
            .to_string();

        println!("{t}");
        let v: Vec<u8> = Vec::new();
        let t = concat(v.iter().map(|_x| RcDoc::text("0")), ",")
            .pretty(LINE_WIDTH)
            .to_string();
        println!("{t}");
```

before it was producing:
```
2,
2,
1,

```

It now produces:
```
2,
2,
1,
,
```

christoph-dfinity:
Fixes #667

Alternative to #668, that doesn't constitute a breaking change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@
 ## Unreleased
 
 ### Candid
+
 * Non-breaking changes:
-  + Fixes a regression in pretty printing when concatenating an empty list of documents (#667)
+  + Fixes a regression in pretty printing when concatenating an empty list of documents
 
 ## 2025-07-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Candid
+* Non-breaking changes:
+  + Fixes a regression in pretty printing when concatenating an empty list of documents (#667)
+
 ## 2025-07-29
 
 ### Candid 0.10.16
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Candid
+* Non-breaking changes:
+  + Fixes a regression in pretty printing when concatenating an empty list of documents (#667)
```
ilbertt: I like to reference PRs/issues in the changelog, but we've never done it before in this changelog. Just to keep it consistent with the rest of the changelog:
```suggestion
### Candid

* Non-breaking changes:
  + Fixes a regression in pretty printing when concatenating an empty list of documents
```
(this suggestion also adds a new line after `##Candid`)

ilbertt (approved):
Great! Thanks for the fix.

Regarding #668, we can merge it into the `next` branch, where we are already collecting the breaking changes for v0.11

PR #674, opened 2025-08-21
feat: cdk v0.18 support

q-uint:
### Update Candid Rust bindings for ic-cdk v0.18
- Switches generated inter-canister calls from `ic_cdk::call(...)` to the new builder-pattern API:
```
Ok(Call::bounded_wait(self.0, "...").await?.candid()?)
Ok(Call::bounded_wait(self.0, "...").with_arg(...).await?.candid()?)
Ok(Call::bounded_wait(self.0, "...").with_args(&(...)).await?.candid()?)
```
- Removes deprecated `ic_cdk::api::call` usage.

See the ic-cdk v0.18 [migration guide](https://github.com/dfinity/cdk-rs/blob/main/ic-cdk/V18_GUIDE.md) for details.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -306,6 +306,7 @@ dependencies = [
  "pretty",
  "rand",
  "serde",
+ "serde_json",
  "test-generator",
  "thiserror",
  "toml",
--- a/rust/candid_parser/Cargo.toml
+++ b/rust/candid_parser/Cargo.toml
@@ -32,6 +32,7 @@ lalrpop-util = "0.20.0"
 logos = "0.14"
 convert_case = "0.6"
 handlebars = "6.0"
+serde_json = "1.0.74"
 toml = { version = "0.8", default-features = false, features = ["parse"] }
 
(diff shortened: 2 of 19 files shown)


ilbertt:
I had an interesting discussion with the team. We are thinking of reorganizing the bindings generation. I will update you soon.

adamspofford-dfinity:
Hi @q-uint, and thank you for the PR.

The version incompatibility issue mentioned above is actually revealing a weakness of Candid’s bindgen design.

Candid generates code for other libraries in other repos, so it must know what versions of other crates it's generating it for. In other words, a new major version of the CDK may require a new major version of Candid in the same way that a new major version of Candid requires a new major version of the CDK.

To avoid needing to release these libraries in lockstep, we’ve instead decided to factor out the code that generates functions from candid_parser, and move it into the agent and CDK repos, leaving the code that generates types intact.

This moves the codegen that depends on a package’s version into the same repo as the package.

This PR would be a duplicate of that work, so it is being closed, but thank you for your contribution, and you'll be credited when the feature is announced.

PR #678, opened 2025-09-18
fix: subtyping and coercion rules for optional types

mraszyk:
This PR fixes the specification of subtyping and coercion rules for optional types.

The changes to the specification are implemented in a separate [PR](https://github.com/dfinity/candid/pull/679).

# Subtyping rules

## Subtyping fix 1

The specification claims that the rules allow "*any* type to be regarded
as a subtype of an option".

Hence, the pre-condition of the rule
```

```
is always false and this rule can be dropped.

## Subtyping fix 2

The existing subtyping rule
```
not (null <: <datatype'>)

<datatype> <: <datatype'>

-----------------------------

<datatype> <: opt <datatype'>
```
should be fixed to
```
not (null <: <datatype>)

<datatype> <: <datatype'>

-----------------------------

<datatype> <: opt <datatype'>
```
according to the specification "It can also be specialised to its
constituent type, unless that type is itself optional". Also the fact
that `nat <: opt opt nat` cannot be derived otherwise.

## Subtyping fix 3

The pre-condition `not (null <: <datatype'>)` in the existing
subtyping rule

```
not (null <: <datatype'>)

```
effectively only forbids `null` as `<datatype'>` because every other
`<datatype'>` such that `null <: <datatype'>` would also satisfy
`<datatype> <: <datatype'>` and thus violate the other
pre-condition. Since the corresponding coercion rule coerces to `null` in
this case anyway, we should enable `null` as `<datatype'>` in this rule
(as otherwise a separate rule to cover this case is needed with the same
coercion rule). Hence, the updated rule should read
```

```

## Subtyping fix 4

Analogously to the existing rule
```

----------------------

null <: opt <datatype>
```
we need to add a new rule
```

--------------------------

reserved <: opt <datatype>
```
otherwise the type `reserved` is not a subtype of an option (contrary to
the claim that every type is a subtype of an option).

## Subtyping fix 5

The specification claims that the rules allow "any type to be regarded
as a subtype of an option". Hence, we need to add the rule
```
not (null <: <datatype>)

-----------------------------

<datatype> <: opt <datatype'>
```
otherwise the type `nat` is not a subtype of `opt nat64`.

## Fixed subtyping rules

The final set of subtyping rules for optional types is
```
----------------------

null <: opt <datatype>
```

```

--------------------------

reserved <: opt <datatype>
```

```
<datatype> <: <datatype'>

```

```

```

```
not (null <: <datatype>)

<datatype> <: <datatype'>

-----------------------------

<datatype> <: opt <datatype'>
```

```
not (null <: <datatype>)

-----------------------------

<datatype> <: opt <datatype'>
```
and it is easy to see that these rules are non-overlapping and
correspond to coercion rules.

# Coercion rules

## Coercion fix 1

The rule
```

<v> : reserved ~> null : opt <t'>
```
is overlapping with
```
null <: <t>

-----------------------------

null : <t> ~> null : opt <t'>
```
and thus the former should be fixed to
```
not (<v> = null)

-----------------------------

<v> : reserved ~> null : opt <t'>
```

## Coercion fix 2

The coercion rule
```
not (null <: <t>)

not (null <: <t'>)

<v> : <t> ~> <v'> : <t'>

--------------------------------

<v> : <t> ~> opt <v'> : opt <t'>
```
has a pre-condition `not (null <: <t'>)` that does not correspond to
any pre-condition in the corresponding subtyping rule (the fixed rule in
the section "Subtyping fix 2") and prevents coercion of `nat` into `opt opt nat`
(for instance). Hence, we drop the pre-condition and the updated
rule is
```
not (null <: <t>)

<v> : <t> ~> <v'> : <t'>

--------------------------------

<v> : <t> ~> opt <v'> : opt <t'>
```

## Coercion fix 3

The rule
```
not (null <: <t'>)

not (exists <w>. <v> : <t> ~> <w> : <t'>)

--------------------------------

<v> : <t> ~> null : opt <t'>
```
allows `opt 5 : opt nat` to coerce to `null : opt nat` which does not make
sense as it loses data (`opt 5`) that could be preserved. This rule should
be extended with an additional pre-condition `not (null <: <t>)` since
there are separate coercion rules for such cases.

## Coercion fix 4

The rule
```
null <: <t'>

not (null <: <t>)

----------------------------

<v> : <t> ~> null : opt <t'>
```
allows `5 : nat` to coerce to `null : opt opt nat` which does not make
sense. This rule should be extended with an additional pre-condition
```
not (exists <w>. <v> : <t> ~> <w> : <t'>)
```
to only coerce to `null : opt <t'>` if coercion to `<t'>` failed.

## Coercion fix 5

But now we easily see that the two rules
```
not (null <: <t>)

not (null <: <t'>)

not (exists <w>. <v> : <t> ~> <w> : <t'>)

--------------------------------

<v> : <t> ~> null : opt <t'>
```
and
```
not (null <: <t>)

null <: <t'>

not (exists <w>. <v> : <t> ~> <w> : <t'>)

----------------------------

<v> : <t> ~> null : opt <t'>
```
can be merged into a single rule
```
not (null <: <t>)

not (exists <w>. <v> : <t> ~> <w> : <t'>)

--------------------------------

<v> : <t> ~> null : opt <t'>
```

## Fixed coercion rules

The final set of coercion rules is
```
not (<v> = null)

<v> : reserved ~> null : opt <t'>
```

```
null <: <t>

-----------------------------

null : <t> ~> null : opt <t'>
```

```
<v> : <t> ~> <v'> : <t'>

----------------------------------------

opt <v> : opt <t> ~> opt <v'> : opt <t'>
```

```
not (exists <w>. <v> : <t> ~> <w> : <t'>)

----------------------------------------

opt <v> : opt <t> ~> null : opt <t'>
```

```
not (null <: <t>)

<v> : <t> ~> <v'> : <t'>

--------------------------------

<v> : <t> ~> opt <v'> : opt <t'>
```

```
not (null <: <t>)

not (exists <w>. <v> : <t> ~> <w> : <t'>)

--------------------------------

<v> : <t> ~> null : opt <t'>
```
and it is easy to see that these rules are non-overlapping and
correspond to subtyping rules.

### DIFF
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -953,12 +953,12 @@ An optional value coerces at an option type, if the constituent value has a suit
 ----------------------------------------
 opt <v> : opt <t> ~> opt <v'> : opt <t'>
 
-not (<v> : <t> ~> _ : <t'>)
+not (exists <w>. <v> : <t> ~> <w> : <t'>)
 ----------------------------------------
 opt <v> : opt <t> ~> null : opt <t'>
 ```
 
-Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it can be decoded successfully:
+Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it can be decoded successfully at the content type:
 ```
 not (null <: <t>)
 not (<v'> = null)
@@ -979,7 +979,7 @@ not (<v> = null)
 <v> : reserved ~> null : opt <t'>
 
 not (null <: <t>)
-not (<v> : <t> ~> _ : <t'>)
+not (exists <w>. <v> : <t> ~> <w> : <t'>)
 ----------------------------
 <v> : <t> ~> null : opt <t'>
 ```
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -961,15 +961,9 @@ opt <v> : opt <t> ~> null : opt <t'>
 Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it can be decoded successfully at the content type:
 ```
 not (null <: <t>)
-not (<v'> = null)
 <v> : <t> ~> <v'> : <t'>
 --------------------------------
 <v> : <t> ~> opt <v'> : opt <t'>
-
-not (null <: <t>)
-<v> : <t> ~> null : <t'>
-----------------------------
-<v> : <t> ~> null : opt <t'>
 ```
 
 Any other value goes to `null`:
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -812,7 +812,7 @@ opt <datatype> <: opt <datatype'>
 
 not (null <: <datatype>)
 not (<datatype> <: <datatype'>)
----------------------------------
+-------------------------------
 <datatype> <: opt <datatype'>
 ```
 *Note:* These rules are necessary in the presence of the unusual record and variant rules shown below. Without them, certain upgrades may generally be valid one step at a time, but not taken together, which could cause problems for clients catching up with multiple upgrades.
--- a/spec/Candid.md
+++ b/spec/Candid.md
@@ -785,37 +785,40 @@ An option type can be specialised via its constituent type.
 opt <datatype> <: opt <datatype'>
 ```
 
-Furthermore, an option type can be specialised to `null`:
+Furthermore, an option type can be specialised to `null` and `reserved`:
 ```
 ------------------------
 null <: opt <datatype>
+
+--------------------------
+reserved <: opt <datatype>
 ```
 
 It can also be specialised to its constituent type, unless that type is itself optional:
 
 ```
-not (null <: <datatype'>)
+not (null <: <datatype>)
 <datatype> <: <datatype'>
 -----------------------------
 <datatype> <: opt <datatype'>
 ```
-The premise means that the rule does not apply when the constituent type is itself `null`, an option or `reserved`. That restriction is necessary so that there is no ambiguity. For example, otherwise there would be two ways to interpret `null` when going from `opt nat` to `opt opt nat`, either as `null` or as `?null`.
+The premise means that the rule does not apply when the constituent type is itself `null`, an option or `reserved`. That restriction is necessary so that there is no ambiguity. For example, otherwise there would be two ways to interpret `null` when going from `opt nat` to `opt opt nat`, either as `null` or as `opt null`.
 
 Finally, in order to maintain *transitivity* of subtyping, two unusual rules allow, in fact, *any* type to be regarded as a subtype of an option.
 ```
-not (<datatype> <: opt <datatype'>)
----------------------------------
-opt <datatype> <: opt <datatype'>
-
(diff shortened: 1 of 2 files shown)


spec/Candid.md
```diff
@@ -958,24 +961,25 @@ opt <v> : opt <t> ~> null : opt <t'>
...
 <v> : reserved ~> null : opt <t'>
 
-not (null <: <t'>)
-not (<v> : <t> ~> _ : <t'>)
---------------------------------
-<v> : <t> ~> null : opt <t'>
-
-null <: <t'>
 not (null <: <t>)
+not (<v> : <t> ~> _ : <t'>)
```
crusso:
```suggestion
not (exists w, <v> : <t> ~> w : <t'>)
```
or

```
all w. not (<v> : <t> ~> w : <t'>)
```

spec/Candid.md
```diff
@@ -785,37 +785,40 @@ An option type can be specialised via its constituent type.
 opt <datatype> <: opt <datatype'>
 ```
 
-Furthermore, an option type can be specialised to `null`:
+Furthermore, an option type can be specialised to `null` and `reserved`:
 ```
 ------------------------
 null <: opt <datatype>
+
```
crusso, line 792:
````suggestion
```

```
````
(just to separate the rules)

File: spec/Candid.md [hunk 3]
```diff
@@ -785,37 +785,40 @@ An option type can be specialised via its constituent type.
...
 ```
-The premise means that the rule does not apply when the constituent type is itself `null`, an option or `reserved`. That restriction is necessary so that there is no ambiguity. For example, otherwise there would be two ways to interpret `null` when going from `opt nat` to `opt opt nat`, either as `null` or as `?null`.
+The premise means that the rule does not apply when the constituent type is itself `null`, an option or `reserved`. That restriction is necessary so that there is no ambiguity. For example, otherwise there would be two ways to interpret `null` when going from `opt nat` to `opt opt nat`, either as `null` or as `opt null`.
 
 Finally, in order to maintain *transitivity* of subtyping, two unusual rules allow, in fact, *any* type to be regarded as a subtype of an option.
 ```
-not (<datatype> <: opt <datatype'>)
```
crusso, line 812:
````suggestion
```

```
````
(just to separate the rules)

File: spec/Candid.md (hunk 1 above)
Comment (crusso) [reply spec/Candid.md]: (there's another ambiguous underscore above but I can't suggest a correction).

File: spec/Candid.md [hunk 4]
```diff
@@ -958,24 +961,25 @@ opt <v> : opt <t> ~> null : opt <t'>
 Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it can be decoded successfully:
```
crusso:
```suggestion
Coercing a non-null, non-optional and non-reserved type at an option type treats it as an optional value, if it can be decoded successfully at the content type:
```

crusso (approved):
Looks good to me and I think the rules all match up with your final fixes in the doc.
To be honest, it might be good to include the doc in the PR description, in case anyone want to understand the PR later.

PR #679, opened 2025-09-22
fix: coercion of values into nested optional types

mraszyk:
This PR fixes coercion of values into nested optional types.

Before the fix from this PR, `5 : nat64` would coerce to `opt 5 : opt na64` (as expected) and `null : opt opt nat64` (unexpected). This PR fixes the unexpected result to be `opt opt 5 : opt opt nat64` (as expected).

Before the fix from this PR, `5 : int` would coerce to `null : opt opt nat`. This PR fixes the result to be `opt null : opt opt nat`. The motivation for this change is to enable using optional types as follows:
- outer option characterizes if a value (of any type) is provided;
- inner option expresses to optimistically deserialize the actual value into the given type (without failing to deserialize and without losing the information if a value of any type is provided).

This PR also adds multiple test cases covering the change to coercion (including coercion to nested optional types over `null` and `reserved`).

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -839,16 +839,8 @@ impl<'de> de::Deserializer<'de> for &mut Deserializer<'de> {
             }
             (_, TypeInner::Opt(t2)) => {
                 self.expect_type = self.table.trace_type(t2)?;
-                if !matches!(
-                    self.expect_type.as_ref(),
-                    TypeInner::Null | TypeInner::Reserved | TypeInner::Opt(_)
-                ) {
-                    check_recursion! {
-                        self.recoverable_visit_some(visitor)
-                    }
-                } else {
-                    self.deserialize_ignored_any(serde::de::IgnoredAny)?;
-                    visitor.visit_none()
+                check_recursion! {
+                    self.recoverable_visit_some(visitor)
                 }
             }
             (_, _) => check!(false),
--- a/rust/candid/tests/serde.rs
+++ b/rust/candid/tests/serde.rs
@@ -1,6 +1,6 @@
 use candid::{
     decode_one_with_config, encode_one, CandidType, Decode, DecoderConfig, Deserialize, Encode,
-    Int, Nat,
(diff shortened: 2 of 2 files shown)


ilbertt (approved):

test/construct.test.did
```diff
@@ -119,8 +119,8 @@ assert blob "DIDL\00\01\7e\00" == "(opt false)"   : (opt bool) "opt: parsing fal
 assert blob "DIDL\00\01\7e\01" == "(opt true)"    : (opt bool) "opt: parsing true : opt bool";
 assert blob "DIDL\00\01\7e\02"                   !: (opt bool) "opt: parsing invalid bool at opt bool";
 assert blob "DIDL\01\6e\7f\01\00\00" == "(null)"  : (opt opt null) "opt: parsing (null : opt null) at opt opt null gives null, not opt null";
-assert blob "DIDL\00\01\7e\01" == "(null)"        : (opt opt bool) "opt: parsing true : opt opt bool gives null";
-assert blob "DIDL\00\01\7e\01" == "(null)"        : (Opt) "opt: parsing (true : bool) at \"fix opt\" gives null";
+assert blob "DIDL\00\01\7e\01" == "(opt opt true)" : (opt opt bool) "opt: parsing true : opt opt bool gives null";
```
mraszyk, line 123: That's the change by this PR: `true : bool` coerces to `opt opt true : opt opt bool`.

test/construct.test.did
```diff
@@ -119,8 +119,8 @@ assert blob "DIDL\00\01\7e\00" == "(opt false)"   : (opt bool) "opt: parsing fal
...
+assert blob "DIDL\00\01\7e\01"                   !: (Opt) "opt: parsing (true : bool) at \"fix opt\" fails";
```
mraszyk, line 124: `true : bool` cannot coerce into an _infinitely_ nested optional type

crusso (approved):
Looks fine to me (also ported the changes to Motoko with little problem)

PR #680, opened 2025-10-01
fix: fail decoding of reserved into null

mraszyk:
This PR makes decoding of a value of type `reserved` into type `null` always fail, consistently with the subtyping relation (`reserved` is not a subtype of `null`).

So far,
```
$ didc encode '(variant { foo = null })' -t '(variant { foo : reserved })' | didc decode -t '(variant { foo : null })'
Error: input: 4449444c016b01868eb70270010000_
table: type table0 = variant { 5_097_222 : reserved }
wire_type: reserved, expect_type: null

Caused by:
    0: Fail to decode argument 0 from table0 to variant { foo }
    1: Subtyping error: unit_variant
```
fails, but
```
$ didc encode '(null)' -t '(reserved)' | didc decode -t '(null)'
(null : null)
$ didc encode '(record { foo = null })' -t '(record { foo : reserved })' | didc decode -t '(record { foo : null })'
(record { foo = null : null })
```
both succeed. This behavior is inconsistent and this PR makes all such cases fail.

Additionally, the wire type of missing data is now set to `null` (instead of `reserved`) so that it can be successfully decoded to `null` as expected type.

### DIFF
--- a/rust/candid/src/de.rs
+++ b/rust/candid/src/de.rs
@@ -74,7 +74,7 @@ impl<'de> IDLDeserialize<'de> {
                 TypeInner::Opt(_) | TypeInner::Reserved | TypeInner::Null
             ) {
                 self.de.expect_type = expected_type;
-                self.de.wire_type = TypeInner::Reserved.into();
+                self.de.wire_type = TypeInner::Null.into();
                 return T::deserialize(&mut self.de);
             } else if self.de.config.full_error_message
                 || text_size(&expected_type, MAX_TYPE_LEN).is_ok()
@@ -754,8 +754,7 @@ impl<'de> de::Deserializer<'de> for &mut Deserializer<'de> {
     {
         self.unroll_type()?;
         check!(
-            *self.expect_type == TypeInner::Null
-                && matches!(*self.wire_type, TypeInner::Null | TypeInner::Reserved),
+            *self.expect_type == TypeInner::Null && matches!(*self.wire_type, TypeInner::Null),
             "unit"
         );
         self.add_cost(1)?;
@@ -1125,7 +1124,7 @@ impl<'de> de::SeqAccess<'de> for Compound<'_, 'de> {
                 self.de.wire_type = wire
                     .pop_front()
                     .map(|f| f.ty)
-                    .unwrap_or_else(|| TypeInner::Reserved.into());
+                    .unwrap_or_else(|| TypeInner::Null.into());
                 seed.deserialize(&mut *self.de).map(Some)
             }
             _ => Err(Error::subtype("expect vector or tuple")),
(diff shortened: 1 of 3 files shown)


rust/candid/src/types/value.rs
```diff
@@ -248,6 +248,7 @@ impl IDLValue {
                         .get(id.as_ref())
                         .cloned()
                         .or_else(|| match env.trace_type(ty).unwrap().as_ref() {
+                            TypeInner::Null => Some(&IDLValue::Null),
                             TypeInner::Opt(_) => Some(&IDLValue::None),
                             TypeInner::Reserved => Some(&IDLValue::Reserved),
```
crusso, line 253: Odd that these are morally all the value null but represent by different IDLValues, Null None and Reserved. But I don't know enought about the implementation....

rust/candid/src/de.rs
```diff
@@ -74,7 +74,7 @@ impl<'de> IDLDeserialize<'de> {
                 TypeInner::Opt(_) | TypeInner::Reserved | TypeInner::Null
             ) {
                 self.de.expect_type = expected_type;
-                self.de.wire_type = TypeInner::Reserved.into();
+                self.de.wire_type = TypeInner::Null.into();
```
crusso, line 77: What does this actually do?

test/construct.test.did
```diff
@@ -205,6 +205,23 @@ assert blob "DIDL\07\6c\07\c3\e3\aa\02\01\d3\e3\aa\02\7e\d5\e3\aa\02\02\db\e3\aa
 
 assert blob "DIDL\07\6c\07\c3\e3\aa\02\01\d3\e3\aa\02\7e\d5\e3\aa\02\02\db\e3\aa\02\01\a2\e5\aa\02\04\bb\f1\aa\02\06\86\8e\b7\02\75\6c\02\d3\e3\aa\02\7e\86\8e\b7\02\75\6b\02\d1\a7\cf\02\7f\f1\f3\92\8e\04\03\6c\02\a0\d2\ac\a8\04\7c\90\ed\da\e7\04\02\6e\05\6c\02\a0\d2\ac\a8\04\7c\90\ed\da\e7\04\04\6b\02\d3\e3\aa\02\7f\86\8e\b7\02\7f\01\00\01\0b\00\00\00\01\00\00\0a\00\00\00\01\14\00\00\2a\00\00\00" !: (record { foo: int; new_field: bool }) "new record field";
 
+// parsing reserved as null
```
crusso, line 209: The tests look good to me.

rust/candid/src/de.rs
mraszyk: This sets a "virtual" wire type for values that are actually missing on the wire: in this case, we're decoding an element in a top-level tuple that does not correspond to a value on the wire, e.g., decoding the last element in a tuple of type `(nat, nat, null)` with the tuple `(5:nat, 6:nat)` actually on the wire. In this case, the type of such an extra element in the tuple must be `null`, `opt`, or `reserved`, but only `null` is a subtype of all these types and thus the only legal choice in general (`reserved` used to work before because it did coerce into `null`, too, which this PR is fixing).

rust/candid/src/types/value.rs
mraszyk: This is to allow decoding record fields of type `null` missing on the wire, e.g., in
```
$ didc encode '(record {})' -t '(record { 0 : null })'
```
where the field with tag `0` is only in the expected type, but missing on the wire (`record {}`).

PR #690, opened 2025-11-10
Add span-aware AST and lossy parser for language server

wiyota:
**Overview**

Introduce a span-aware AST that the parser now builds first so experimental [`candid-language-server`](https://github.com/perforate-org/candid-language-server) can attach precise locations to every declaration, field, and service item.

The new `syntax::spanned` module defines the source-of-truth structures and parsing helpers (`rust/candid_parser/src/syntax/spanned.rs:13`), `grammar.lalrpop` drives lalrpop to emit those structures (`rust/candid_parser/src/grammar.lalrpop:4`), and the existing public AST in `syntax` converts from the spanned version to keep the current API stable (`rust/candid_parser/src/syntax/mod.rs:49`).

Parser entry points and new regression tests (`rust/candid_parser/tests/parse_prog.rs:16`) exercise the updated pipeline so doc comments and services survive the round-trip.

A lossy parser variant was added so the language server can recover as much structure as possible while still collecting tokens.

**Requirements**

- Provide source spans for every AST node so the language server can power go-to-definition, hover, and diagnostics without reparsing.  
- Preserve the existing span-less `candid_parser::syntax` types for downstream crates while offering a lossless path to spanned data.  
- Ensure token-recording and lossy parsing entry points keep working with the new AST (covered by the new parser tests in `rust/candid_parser/tests/parse_prog.rs:16`).

**Considered Solutions**

- Extending the existing AST structs with optional spans would have been a breaking change for every consumer.  
- Maintaining a separate parser just for the language server risked grammar drift and duplicated maintenance.  
- The spanned-AST-plus-conversion approach keeps a single grammar and minimizes API churn.

**Recommended Solution**

Adopt the spanned AST internally and convert to the existing structs for callers that rely on them. Bindings/generators were updated where they destructure AST nodes, the lossy parser entry point (`parse_prog_lossy`) now rides on the spanned types, and new tests cover both exact and lossy parsing scenarios.

This keeps the public surface stable while supplying the language server with the spans it needs.

**Considerations**

- This work primarily benefits experimental `candid-language-server`; other consumers interact with the familiar API but pay a small overhead because span-less structs are now derived from the spanned ones on every parse.  
- When consumers need the old AST, we convert out of the spanned version, so there is a slight allocation/copy cost that we’ll monitor.  
- We now have two AST definitions (`syntax/mod.rs` vs. `syntax/spanned.rs`), so future changes must modify both in lockstep to prevent drift.


rust/candid_parser/src/syntax/mod.rs
```diff
@@ -24,6 +28,8 @@ pub enum IDLType {
     PrincipalT,
 }
 
+pub type IDLTypeKind = IDLType;
```
ilbertt: Why do we need this alias?

rust/candid_parser/src/syntax/spanned.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+impl<T> Spanned<T> {
+    pub fn new(value: T, span: Span) -> Self {
+        Spanned { value, span }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IDLType {
```
ilbertt: I would call this struct `IDLTypeWithSpan` or so, so that it doesn't get confused with the existing `syntax::IDLType`

rust/candid_parser/src/typing.rs
```diff
@@ -136,12 +136,9 @@ fn check_meths(env: &Env, ms: &[Binding]) -> Result<Vec<(String, Type)>> {
...
-        match dec {
-            Dec::TypD(Binding { id, typ, docs: _ }) => {
-                let t = check_type(env, typ)?;
-                env.te.0.insert(id.to_string(), t);
-            }
-            Dec::ImportType(_) | Dec::ImportServ(_) => (),
+        if let Dec::TypD(Binding { id, typ, .. }) = dec {
+            let t = check_type(env, typ)?;
+            env.te.0.insert(id.to_string(), t);
         }
```
ilbertt: Even though this makes the code a bit cleaner, I would do it in a separate PR to keep the scope of this PR as close as possible to its goal. Feel free to open another PR!

rust/candid_parser/src/typing.rs
```diff
@@ -225,8 +222,11 @@ fn load_imports(
     list: &mut Vec<(PathBuf, String, IDLProg)>,
 ) -> Result<()> {
     for dec in prog.decs.iter() {
-        let include_serv = matches!(dec, Dec::ImportServ(_));
-        if let Dec::ImportType(file) | Dec::ImportServ(file) = dec {
+        if let Some((file, include_serv)) = match dec {
+            Dec::ImportType(file) => Some((file, false)),
+            Dec::ImportServ(file) => Some((file, true)),
+            _ => None,
+        } {
```
ilbertt: Same as above, I would avoid refactoring the code if it's not in the scope of the current PR

rust/candid_parser/src/syntax/spanned.rs
ilbertt, line 1: We have the [next](https://github.com/dfinity/candid/tree/next) branch in which we are collecting all the breaking changes. There we can change the `syntax::IDLType` and all the related structs to have the new `span` field. What do you think? I feel like repeating the code of the `IDLType` here is not a good approach in terms of maintainability

rust/candid_parser/src/bindings/typescript.rs
```diff
@@ -29,23 +29,38 @@ fn pp_ty_rich<'a>(
...
-        (TypeInner::Vec(ref t), Some(IDLType::VecT(syntax_inner))) => {
-            pp_vec(env, t, Some(syntax_inner), is_ref)
+        TypeInner::Vec(ref t) => {
+            if let Some(IDLTypeKind::VecT(syntax_inner)) = syntax {
+                return pp_vec(env, t, Some(syntax_inner), is_ref);
+            }
+            pp_vec(env, t, None, is_ref)
         }
-        (_, _) => pp_ty(env, ty, is_ref),
+        _ => pp_ty(env, ty, is_ref),
```
ilbertt: Similar to the `typing.rs` file, I would avoid refactoring code that is not in scope with the PR's goal

rust/candid_parser/src/syntax/mod.rs
wiyota: I added the alias to mirror IDLTypeKind in the spanned module, but looking again, it doesn’t actually add any value...

rust/candid_parser/src/syntax/spanned.rs
wiyota: I placed the traditional module alongside the new one due to a lack of understanding of the destructive change's impact. However, as you suggested, unifying it into a structure that includes span is indeed a better approach. I will redesign it accordingly.

wiyota:
I've made changes to keep it as close to the `master` branch as possible. How does it look?

PR #692, opened 2025-12-10
ci: release candid_ui

lwshang:
GItHub releases will include prebuilt `candid_ui.wasm` so that tools like `icp-cli` can integrate.

A test release can be found [here](https://github.com/dfinity/candid/releases/tag/test_release_candid_ui_2).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,7 +125,7 @@ jobs:
           wasm-opt --strip-debug -Oz target/wasm32-unknown-unknown/canister/didjs.wasm -o target/wasm32-unknown-unknown/canister/didjs_opt.wasm
       - name: Prepare asset
         run: |
-          cp tools/ui/target/wasm32-unknown-unknown/canister/didjs.wasm candid_ui.wasm
+          cp tools/ui/target/wasm32-unknown-unknown/canister/didjs_opt.wasm candid_ui.wasm
       - name: Upload candid_ui to release
         uses: svenstaro/upload-release-action@v2
         with:


.github/workflows/release.yml
```diff
@@ -16,7 +16,7 @@ jobs:
             name: linux64
             artifact_name: target/x86_64-unknown-linux-musl/release/didc
             asset_name: didc-linux64
-          - os: macos-13-large
+          - os: macos-14-large
```
adamspofford: We can get rid of the 'large' runners these days with `macos-15-intel`.

.github/workflows/release.yml
```diff
@@ -114,3 +104,36 @@ jobs:
...
+        run: |
+          sudo apt-get update -yy
+          sudo apt-get install -yy binaryen
+      - name: Build candid_ui
+        working-directory: tools/ui
+        run: |
+          npm install
+          npm run build
+          cargo build --target wasm32-unknown-unknown --release --package didjs
+          wasm-opt --strip-debug -Oz target/wasm32-unknown-unknown/release/didjs.wasm -o target/wasm32-unknown-unknown/release/didjs_opt.wasm
```
adamspofford: wasm-opt isn't nearly as good at this after the fact as rustc is (1.6m vs 1.4m) - it'd be better to introduce a canister profile in Cargo.toml:
```toml
[profile.canister]
inherits = "release"
opt-level = "z"
strip = "debuginfo"
```

.github/workflows/release.yml
lwshang: IIRC, the idea was that we build on the relatively "old" OS and then verify that the binaries work on both the "old" and latest OS.
For example, build on `macos-14-large` and verify against both `macos-14-large` and `macos-15-large`.

lwshang: And IIUC, `macos-15-large` and `macos-15-intel` are alternative labels of the same runner image.
If there were `macos-14-intel`, I would love to change all to `*-intel`.
Having `macos-14-large` and `macos-15-intel` together will be more confusing.

.github/workflows/release.yml
lwshang: release profile: 2.0 M -> wasm-opt 1.6 M
canister profile: 1.9 M -> wasm-opt 1.2 M

Perhaps we should use both the canister profile and wasm-opt.

.github/workflows/release.yml
adamspofford: The images are identical but the runner is beefier and costs more. Building on an older OS version is mostly a concern for Linux where the glibc version is baked into the binary and so a 24.04-built executable won't run on 22.04; macOS doesn't have this issue. But if you think it is better to keep checking on 14 then you can ignore this one.

lwshang: I switched build to `macos-15-intel` and drop the test for macos.

PR #698, opened 2026-01-30
chore: enforce limits in binary parsing

venkkatesh-sekar:


rust/candid/src/binary_parser.rs
```diff
@@ -22,16 +24,22 @@ fn read_sleb<R: Read + Seek>(reader: &mut R, ro: &ReadOptions, (): ()) -> BinRes
 
 #[derive(BinRead, Debug)]
 #[br(magic = b"DIDL")]
+#[br(import(max_type_len: Option<usize>))]
 pub struct Header {
+    #[br(args(max_type_len))]
     table: Table,
     #[br(parse_with = read_leb)]
+    #[br(assert(len <= max_type_len.unwrap_or(MAX_TYPE_TABLE_LEN as usize) as u64, "args length exceeded"))]
```
lwshang:
```suggestion
```
The `len` here represents the number of args (values), not types. Using `MAX_TYPE_TABLE_LEN` to limit it seems inconsistent with its purpose. Additionally, the existing `decoding_quota` and `skipping_quota` mechanisms already provide protection against payloads with an excessive number of values.

PR #702, opened 2026-02-13
feat: Add BoundedVec 

michael-weigelt:
This makes the `BoundedVec` type, thus far used in the [management canister types](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/types/management_canister_types/src/bounded_vec.rs?L14), available for public use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Candid
+
+* Non-breaking changes:
+  + Add `BoundedVec` type to `candid::types::bounded_vec` for bounding a vector by number of elements, total data size, and per-element data size during deserialization
+
 ## 2026-02-10
 
 ### Candid 0.10.22


rust/candid/src/types/bounded_vec.rs
```diff
@@ -0,0 +1,457 @@
...
+
+use crate::{types::TypeInner, CandidType};
+
+/// Indicates that `BoundedVec<...>` template parameter (eg. length, total data size, etc) is unbounded.
+pub const UNBOUNDED: usize = usize::MAX;
+
+/// Struct for bounding vector by different parameters:
+/// - number of elements
+/// - total data size in bytes
+/// - single element data size in bytes
```
dsarlis: Perhaps a set of doc tests would be useful in showcasing how the type can be used and expected results.

michael-weigelt: I added a minimal example.

PR #705, opened 2026-02-26
feat: Implement DataSize for Principal

dsarlis:
This PR implements `DataSize` for the `Principal` type so that it can be used as a parameter in `BoundedVec` in cases where interface implementors want to bound a vector of principals they require as input.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+## 2026-02-27
+
+### Candid 0.10.24
+
+* Non-breaking changes:
+  + Implement `DataSize` for `Principal`, enabling `Principal` as an element type in `BoundedVec`
+
 ## 2026-02-20
 
 ### Candid 0.10.23
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -209,13 +209,13 @@ dependencies = [
 
 [[package]]
 name = "candid"
-version = "0.10.23"
+version = "0.10.24"
 dependencies = [
  "anyhow",
  "bincode",
  "binread",
  "byteorder",
- "candid_derive 0.10.23",
+ "candid_derive 0.10.24",
  "candid_parser 0.3.0",
  "hex",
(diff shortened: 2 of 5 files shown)


rust/candid/src/types/bounded_vec.rs
```diff
@@ -367,8 +367,7 @@ mod tests {
 }
 
 mod data_size {
-    // We implement a standalone trait `DataSize` here and use it
-    // instead of `CountBytes`.
```
dsarlis, line 371: This comment does not make sense I think as it's a leftover from when the code was moved from the ic monorepo.

rust/candid/src/types/bounded_vec.rs
```diff
@@ -416,6 +415,12 @@ mod data_size {
         }
     }
 
+    impl DataSize for Principal {
+        fn data_size(&self) -> usize {
+            self.size()
+        }
+    }
```
dsarlis: Perhaps a better alternative would be to have this implementation in the `ic_principal` crate -- this way we don't have to introduce any function on the `Principal` type (assuming we do not want to expose internal details of its structure). However, this would create a cyclic dependency as `candid` already depends on `ic_principal`.

A way around would be to have `DataSize` moved somewhere else but I'm also not sure if that's better as it's closely related to `BoundedVec` (and in any case it would be a bigger change). If there's a strong preference over any of the aforementioned, or another option I have not considered, let me know.

lwshang: Simplified: used `self.as_slice().len()` directly in `DataSize` instead of adding a new `size()` method. `Principal` already exposes `as_slice()`, so no new API needed.

lwshang (approved):

rust/candid/src/types/bounded_vec.rs
dsarlis: I suppose I was trying to be more accurate as the `len` field itself takes some space too. But it's also constant overhead 1, so it does not really matter.

PR #728, opened 2026-05-06
fix(candid_parser): emit `Float32` for Candid `float32` in Motoko binding

Kamirus:
## Summary

- Replace `panic!("float32 not supported in Motoko")` with `str("Float32")` in the Motoko binding generator.
- Add a `float.did` test fixture covering `float32`/`float64` round-trip methods.

## Why

Motoko added `Float32` in version 1.4.0 ([dfinity/motoko#5906](https://github.com/dfinity/motoko/pull/5906)), updating both the import and export IDL mappings. The spec (`design/IDL-Motoko.md`) is being updated in a companion PR. This brings `didc bind --target mo` in line with what `moc` already does when importing `.did` files directly.

Concretely, any `.did` file using `float32` (e.g. the IC management canister interface) would previously cause `didc bind --target mo` to panic at runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### candid_parser
+
+* Bug fixes:
+  + Motoko binding: emit `Float32` for Candid `float32` instead of panicking. `float32` support was added to Motoko in version 1.4.0.
+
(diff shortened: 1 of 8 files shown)


ggreif (approved):
Perfect!

PR #758, opened 2026-08-11
fix(candid): bound length-prefixed allocations in header parser + release candid 0.10.35

lwshang:
Patch release. Bounds allocation for length-prefixed fields in the type-table header parser. See CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Candid 0.10.35
 
 * Bug fixes:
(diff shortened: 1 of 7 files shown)


marc0olo (dismissed):
Review content removed. Follow-up offline.

marc0olo (approved):
Re-reviewed at f28574a. LGTM.

Checked the parser change against the binrw 0.15.2 internals it relies on, and confirmed the other length-prefixed read sites in `binary_parser.rs` behave correctly and need no equivalent change. The chunk size lines up with what `Take::read_to_end` reserves internally, so buffer growth stays linear. Loop termination and the short-read condition are correct.

Release hygiene is consistent: `candid` and `candid_derive` both to 0.10.35, the `=0.10.35` pin matches, both `Cargo.lock`s updated, and `candid_parser`'s caret dependency correctly needs no bump. `cargo test -p candid` is green locally, and the new tests fail without the parser change.

The follow-up commit is CHANGELOG-only with no code delta, and the reworded compatibility note reads accurately now.

Non-blocking, pre-existing, worth a separate issue: the remaining `#[br(count = len)]` sites narrow a `u64` wire length to `usize`. This PR keeps `u64` end-to-end in the two fields it touches; the rest could be tidied up later.

PR #769, opened 2026-09-22
chore: Release candid 0.10.36

lwshang:
Patch release of `candid` / `candid_derive`, 0.10.35 → 0.10.36. No source changes — it dates the two decoder fixes that are already on `master`.

## Summary
- Bump `candid` and `candid_derive` 0.10.35 → 0.10.36 (and the `=` pin between them), refresh `Cargo.lock`, `rust/bench/Cargo.lock` and `tools/ui/Cargo.lock`
- Move the `Unreleased` CHANGELOG entries into a dated `2026-09-22` / `Candid 0.10.36` section, covering:
  + Scoping a map's decode fast paths to their own half of an entry, so a key and a value each decode under their own declared type (#768)
  + Sharing the decoder's undecoded argument queue behind a reference count, so skipping present optional arguments is no longer superlinear in the argument count (#767)

`tools/ui` patches `candid` to the workspace path, so its lockfile pins the path crate's version and has to move with the bump. `candid-ui-release.yml` builds `didjs` there with `--locked` and triggers on the date tag this release creates, so a stale lock there would only fail at tag time, after merge — not in PR checks.

## Test plan
- [x] `cargo check -p candid -p candid_derive -p candid_parser`
- [x] `cargo test -p candid` — all suites pass
- [x] `cd tools/ui && cargo build --target wasm32-unknown-unknown --profile canister --package didjs --locked` — the release workflow's exact command, succeeds
- [ ] Reviewer to confirm release scope and changelog date
- [ ] Publish to crates.io after merge

## Follow-up (not in this PR)
`rust/candid/fuzz/Cargo.lock` is stale at candid 0.10.34, already on `master`. Nothing builds it with `--locked`, so it is harmless, but all four lockfiles should move together as a release-checklist step.


marc0olo (approved):
Lockfile fix confirmed. `--locked` now passes in all three workspaces, and the
verbatim didjs build command from `candid-ui-release.yml` builds clean — so the
date-tag workflow will go through.

Release itself checks out: no source changes, no `0.10.35` left anywhere, and the
changelog covers exactly #767 and #768, both on master. `cargo test -p candid
--features all` green.

Non-blocking: `rust/candid/fuzz/Cargo.lock` is still at 0.10.34, stale since
before this release. Nothing builds it with `--locked`, but worth a checklist
line so all four lockfiles move together next time.

LGTM.

